PR #746, opened 2026-07-07
OPT-1053 Preserve active waveform cache successors

PORTALSURFER:
## Scope

- Preserve the latest waveform-cache store job submitted for a cache path that already has an active write.
- Split queue diagnostics/outcomes so queued replacement and active-write deferral are distinguishable.
- Add focused regression coverage for stale active writes followed by retained fresh successors.

## Definition of Done

- Fresh cache writes submitted while an older write for the same cache path is active are retained as bounded latest successors.
- A stale active write can abort without losing the newest retained replacement.
- Duplicate queued writes still coalesce to the latest queued job.
- Queue drain accounting includes active successors.

## Validation commands

- `cargo test -p wavecrate store_queue_behavior -- --test-threads=1`
- `cargo test -p wavecrate waveform_cache -- --test-threads=1`
- `bash scripts/ci.sh smoke`

## Known limitations

None for this change. Note: `bash scripts/agent.sh request` reports a pre-existing native-app boundary guardrail violation in `src/native_app/audio/playback/progress.rs`; the requested smoke lane passed.

## Review status

Status: waiting for user review. User review is required before merge.

### DIFF
--- a/src/native_app/waveform/audio_file/waveform_cache/store_queue.rs
+++ b/src/native_app/waveform/audio_file/waveform_cache/store_queue.rs
@@ -1,5 +1,5 @@
 use std::{
-    collections::{HashSet, VecDeque},
+    collections::{HashMap, HashSet, VecDeque},
     path::{Path, PathBuf},
     sync::{Arc, Condvar, LazyLock, Mutex},
     thread,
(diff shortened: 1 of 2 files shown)


PORTALSURFER:
@codex review

PR #780, opened 2026-07-11
OPT-1102 Own ANN loader backing with loaded indexes

PORTALSURFER:
## Scope

- Replace the permanent `Box::leak(HnswIo)` disk-load path with explicit loaded-index ownership.
- Keep each loader at a stable boxed address for any mmap-backed point borrows.
- Encapsulate the HNSW lifetime extension inside a non-extractable owner that drops the graph before its loader.
- Preserve the existing built-index path, search behavior, mutations, and disk format.
- Add repeated load/drop and eight-way racing cache-population backing-lifetime tests.

## Definition of Done

- Every successful disk load owns exactly one loader backing allocation.
- Dropping a loaded state releases that backing.
- Racing loads release all losing states and retain at most the selected winner.
- ANN search, update, container, and legacy migration behavior remains unchanged.

## Validation commands

- `cargo check -p wavecrate-analysis --all-targets`
- `cargo test -p wavecrate-analysis loader_backing -j1`
- `cargo test -p wavecrate-analysis ann_index -j1`
- `cargo test -p wavecrate-analysis -j1`

## Known limitations

The HNSW dependency exposes a lifetime tied to `HnswIo` for optional mmap data but no owning loaded type. This PR contains one documented unsafe lifetime extension inside `LoadedAnnHnsw`; its stable boxed loader and explicit graph-before-loader drop order are the enforced safety invariant.

Closes OPT-1102

### DIFF
--- a/crates/wavecrate-analysis/src/analysis/ann_index/build.rs
+++ b/crates/wavecrate-analysis/src/analysis/ann_index/build.rs
@@ -1,12 +1,12 @@
 use super::container;
 use super::state::{
-    AnnIndexMetaRow, AnnIndexParams, AnnIndexState, build_id_lookup, default_params,
+    AnnHnsw, AnnIndexMetaRow, AnnIndexParams, AnnIndexState, LoadedAnnHnsw, build_id_lookup,
+    default_params,
 };
 use super::storage::{
     default_index_path, hnsw_dump_paths, legacy_id_map_path_for, load_legacy_id_map, read_meta,
 };
 use crate::analysis::decode_f32_le_blob;
-use hnsw_rs::hnswio::HnswIo;
 use hnsw_rs::prelude::*;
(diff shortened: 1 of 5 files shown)


PORTALSURFER:
Addressed the HNSW lifetime finding in 3637ea227. The loaded owner no longer returns the transmuted `Hnsw<'static, ...>` reference at all—not even through a private common accessor. `AnnHnsw` now dispatches only scoped `insert`, `search`, `file_dump`, and `get_nb_point` operations, and `LoadedAnnHnsw` implements those operations without exposing graph or point-indexation access.

Validation:
- `cargo test -p wavecrate-analysis ann_index -- --quiet` (17 passed)
- `cargo test -p wavecrate-analysis -- --quiet` (88 passed)
- `git diff --check`

@codex review

PR #783, opened 2026-07-11
OPT-1099 Wavecrate scan: release DB writer during discovery

PORTALSURFER:
## Scope
- discover file facts and compute required hashes before opening any source-database write transaction
- retain at most 64 prepared files, then revalidate and apply them in a bounded transaction
- hash immediately before apply and bracket hashing with filesystem-fact checks
- refresh current database rows inside the writer transaction before rename or missing-row metadata moves
- preload rename candidates and filesystem-existence checks before acquiring the writer
- skip unchanged file transactions and bound missing-row reconciliation to 64 items
- preserve quick, hard, targeted, rename, progress, and cancellation behavior

## Definition of Done
- traversal and hashing hold no SQLite writer transaction
- metadata writes can complete while discovery is active and survive rename/missing reconciliation
- stale staged file facts or hashes cannot overwrite the source database
- each persistence transaction and prepared-memory batch has a bounded item count
- cancellation never reports a partially committed scan as canceled

## Validation commands
- `cargo test -p wavecrate-scan --lib` (40 passed)
- `cargo clippy -p wavecrate-scan --all-targets --no-deps -- -D warnings`
- `git diff --check`

## Known limitations
- The broader dependency-inclusive clippy invocation is currently blocked on `main` by the pre-existing `CopyJournalEntry::new_copy` argument-count finding addressed independently in PR #781. The Wavecrate Scan all-target lane passes with `--no-deps`.

### DIFF
--- a/crates/wavecrate-library/src/sample_sources/db/write.rs
+++ b/crates/wavecrate-library/src/sample_sources/db/write.rs
@@ -5,6 +5,7 @@ mod database;
 mod event;
 mod mutation;
 mod paths;
+mod scan_queries;
 mod transaction;
 mod upsert;
 
--- a/crates/wavecrate-library/src/sample_sources/db/write/scan_queries.rs
+++ b/crates/wavecrate-library/src/sample_sources/db/write/scan_queries.rs
@@ -1,0 +1,63 @@
+use std::path::PathBuf;
+
+use rusqlite::params;
+
+use super::super::util::{map_sql_error, parse_relative_path_from_db};
+use super::super::{SourceDbError, SourceWriteBatch};
+
+impl SourceWriteBatch<'_> {
+    /// Fetch current hash-matching paths while holding this writer transaction.
+    pub fn list_paths_with_content_hash(
+        &self,
+        content_hash: &str,
+    ) -> Result<Vec<PathBuf>, SourceDbError> {
+        let mut statement = self
+            .tx
+            .prepare("SELECT path FROM wav_files WHERE content_hash = ?1 ORDER BY path ASC")
+            .map_err(map_sql_error)?;
+        let paths = statement
+            .query_map(params![content_hash], |row| row.get::<_, String>(0))
+            .map_err(map_sql_error)?
+            .collect::<Result<Vec<_>, _>>()
+            .map_err(map_sql_error)?;
+        Ok(parse_paths(paths, "hash"))
+    }
+
+    /// Fetch current file-fact-matching paths while holding this writer transaction.
+    pub fn list_paths_with_file_facts(
+        &self,
+        file_size: u64,
+        modified_ns: i64,
+    ) -> Result<Vec<PathBuf>, SourceDbError> {
+        let mut statement = self
+            .tx
+            .prepare(
+                "SELECT path
+                 FROM wav_files
+                 WHERE file_size = ?1 AND modified_ns = ?2
+                 ORDER BY path ASC",
+            )
+            .map_err(map_sql_error)?;
+        let paths = statement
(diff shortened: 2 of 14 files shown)


PORTALSURFER:
@codex review

PORTALSURFER:
@codex review

PORTALSURFER:
@codex review

PORTALSURFER:
Addressed both latest live-row race findings in bc6f21bdc.

- A prepared no-op now re-reads the current DB row before dropping it from the scan snapshot. If a concurrent targeted writer removed, marked missing, or changed that row, the full scan rebases its snapshot and prepares the file for repair.
- Missing-row staging now verifies the path is still absent while holding the bounded writer batch, so a concurrently restored live path is not moved back into pending-renames or removed.
- Added regressions for concurrent no-op removal and concurrent live-row restoration.

Validation:
- `cargo test -p wavecrate-scan scan_refreshes_noop_row_after_concurrent_removal -- --nocapture`
- `cargo test -p wavecrate-scan concurrently -- --nocapture`
- `cargo test -p wavecrate-scan -- --quiet` (43 passed)
- `git diff --check`

@codex review

PORTALSURFER:
Addressed the latest three review findings in `e0f0e92aa`.

- Missing reconciliation now treats only supported regular audio files as restored; directories and skipped symlinks no longer keep stale rows alive.
- No-op validation now compares the current row's content-hash state with the scan snapshot and rebases when a concurrent writer clears it.
- Apply preparation computes a usable hash for every hash-required repair before the writer batch, preventing a concurrent hash clear from reaching the former `required_prepared_hash` panic.
- Added focused regressions for directory replacement, concurrent no-op hash clearing, and missing-row repair during a concurrent hash clear.

Validation:
- `cargo test -p wavecrate-scan missing_stage_prunes_path_replaced_by_directory -- --quiet`
- `cargo test -p wavecrate-scan scan_rebases_noop_when_concurrent_writer_clears_hash -- --quiet`
- `cargo test -p wavecrate-scan missing_repair_survives_concurrent_hash_clear -- --quiet`
- `cargo test -p wavecrate-scan -- --quiet` (46 passed)
- `git diff --check`

@codex review

PORTALSURFER:
Addressed the latest review findings in `55ea36f05`.

- missing reconciliation now recognizes the same scannable-file contract as the walker, so tracked rows below dot-prefixed directories are pruned even when a regular `.wav` still exists there;
- files replaced by directories or skipped symlinks after discovery remain in the stale set for missing reconciliation instead of being dropped on the apply-skip path;
- rename candidate uniqueness is now queried from the current writer transaction by hash or file facts, so a concurrent second missing row makes the candidate ambiguous before metadata can transfer;
- removed the now-redundant pre-writer rename-candidate caches and added focused hidden-path, unsupported-replacement, and fresh-uniqueness regressions.

Validation:
- `cargo test -p wavecrate-scan basics -- --quiet` (21 passed)
- `cargo test -p wavecrate-scan unique_missing_path -- --quiet`
- `cargo test -p wavecrate-scan -- --quiet` (48 passed)
- `git diff --check`

@codex review

PORTALSURFER:
Addressed both follow-up findings in b20f20a9a:

- hidden-directory checks now inspect parent components only, so supported dot-prefixed WAV basenames remain scannable while files below hidden directories remain excluded
- writer-locked hash/facts candidate queries now apply the same supported-audio and AppleDouble filter as read-side lookups

Added regression coverage for both cases.

Validation:
- `cargo test -p wavecrate-scan -- --quiet` (49 passed)
- `cargo test -p wavecrate-library scan_candidate_queries_ignore_unsupported_and_apple_double_rows -- --quiet`

@codex review

PR #784, opened 2026-07-11
OPT-1098 Reson: bound real-time recording capture transport

PORTALSURFER:
## Scope
- replace the unbounded WAV-writer channel with a two-second preallocated SPSC sample ring
- replace callback-side monitor locking/cloning with a bounded half-second SPSC ring and bounded control queue
- drop whole newest callback blocks on overflow so channel/frame alignment remains intact
- reuse preallocated format-conversion scratch storage in the CPAL callback
- expose atomic writer/monitor overrun, disconnect, control-drop, and worker-failure health snapshots
- make writer and monitor workers drop-safe across stream-build and stream-start failures
- share monitor liveness so `InputMonitor::stop()` prevents later recorder appends
- surface completed-session transport loss through Wavecrate telemetry and warning status

## Definition of Done
- slow writer and monitor consumers cannot grow memory without bound
- capture submission does not block, lock, clone sample blocks, or allocate per normal callback
- lost samples and worker/control failures are explicitly observable
- queued samples drain before normal WAV finalization
- startup failure and public monitor-stop paths cannot leak workers or resume monitoring

## Validation commands
- `cargo test -p reson --lib` (147 passed)
- `cargo test -p reson --lib recording` (10 passed)
- `cargo clippy -p reson --all-targets -- -D warnings`
- `git diff --check`

## Known limitations
- Real audio-device recording smoke testing was not available in the locked desktop session; deterministic transport, drain, lifecycle, health, and full Reson library coverage pass.

### DIFF
--- a/crates/reson/src/input/stream.rs
+++ b/crates/reson/src/input/stream.rs
@@ -9,92 +9,125 @@ pub(crate) fn build_input_stream(
     config: &cpal::StreamConfig,
     sample_format: cpal::SampleFormat,
     selection: StreamChannelSelection,
-    mut on_samples: impl FnMut(Vec<f32>) + Send + 'static,
+    mut on_samples: impl FnMut(&[f32]) + Send + 'static,
 ) -> Result<cpal::Stream, AudioInputError> {
     let err_fn = move |err| {
         warn!("Audio input stream error: {err}");
     };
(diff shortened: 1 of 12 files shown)


PORTALSURFER:
@codex review

PR #786, opened 2026-07-11
OPT-1096 Wavecrate scan: verify identity before rename transfer

PORTALSURFER:
## Scope
- remove immediate rename and pending-rename identity transfer based only on size/mtime
- add unmatched large files neutrally and retain removed metadata as pending
- let deep hashing transfer metadata and analysis identity only after an exact content-hash match
- remove the obsolete facts-candidate cache and no-hash rename mutation path

## Definition of Done
- unrelated same-size/same-mtime files never inherit identity, ratings, tags, or analysis artifacts
- genuine large renames reconcile after deep hashing
- deferred rename state survives database restart
- small/already-hashable rename handling remains immediate and hash-authoritative

## Validation commands
- `cargo test -p wavecrate-scan --lib rename_reconciliation` (6 passed)
- `cargo test -p wavecrate-scan --lib` (33 passed)
- `cargo clippy -p wavecrate-scan --all-targets --no-deps -- -D warnings`
- `git diff --check`

## Known limitations
- Dependency-inclusive clippy remains independently blocked on `main` by the `CopyJournalEntry::new_copy` argument-count finding addressed in PR #781; the Wavecrate Scan all-target lane passes with `--no-deps`.

### DIFF
--- a/crates/wavecrate-library/src/sample_sources/db/mod.rs
+++ b/crates/wavecrate-library/src/sample_sources/db/mod.rs
@@ -11,6 +11,7 @@ mod open_profiles;
 mod pending_renames;
 /// Read-only database queries for sample sources.
 pub mod read;
+mod rename_destinations;
 /// SQLite schema management for sample source databases.
 pub mod schema;
 /// Durable per-source tag catalog and sample assignment helpers.
--- a/crates/wavecrate-library/src/sample_sources/db/pending_renames.rs
+++ b/crates/wavecrate-library/src/sample_sources/db/pending_renames.rs
@@ -22,6 +22,8 @@ pub struct PendingRenameEntry {
     pub modified_ns: i64,
     /// Last known content hash, if one was computed before pruning.
     pub content_hash: Option<String>,
+    /// Stable filesystem-object identity captured before pruning, when supported.
+    pub file_identity: Option<String>,
     /// Current rating/tag for the file.
     pub tag: Rating,
     /// Whether the sample was marked looped.
@@ -96,6 +98,7 @@ impl SourceDatabase {
                     file_size,
                     modified_ns: row.get(2)?,
                     content_hash: row.get(3)?,
+                    file_identity: row.get(14)?,
                     tag: Rating::from_i64(row.get::<_, i64>(4)?),
                     looped: row.get::<_, i64>(5)? != 0,
                     sound_type: row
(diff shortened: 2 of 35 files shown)


PORTALSURFER:
@codex review

PORTALSURFER:
@codex review

PORTALSURFER:
@codex review

PORTALSURFER:
@codex review

PORTALSURFER:
Addressed all four findings from the latest review in 972ddf5f2.

- Targeted scans now persist newly discovered rename destinations transactionally with a scan generation, so a destination survives one split watcher batch but expires after an unrelated batch.
- Quick scans persist their candidate snapshot so the exported detached deep-hash helper can still reconcile proven renames without an in-memory `ScanStats` candidate set.
- Full folder discovery is responsive again: it publishes after the quick scan and schedules detached deep hashing.
- Deferred hash/list failures (including cancellation) now propagate from synchronous completion instead of being logged as success.
- Unique candidate hash matches remain valid even when mtime changes, while plain deletes and expired duplicate candidates do not transfer identity.

Validation:
- `cargo test -p wavecrate-scan rename_reconciliation -- --quiet` (14 passed)
- `cargo test -p wavecrate-scan -- --quiet` (41 passed)
- `cargo test -p wavecrate-library sample_sources::db -- --quiet` (107 passed)
- `git diff --check`

@codex review

PORTALSURFER:
Addressed the latest review findings in `2f703cc46`:

- full quick scans now preserve rename destinations carried between split watcher batches;
- deferred hashing is gated on actual rename work or explicitly pending hashes, so a plain delete cannot trigger an unrelated whole-source backfill;
- the controller preserves and publishes already-committed quick-scan deltas if deferred hashing fails.

Validation:
- `cargo test -p wavecrate-scan rename_reconciliation -- --quiet`
- `cargo test -p wavecrate deferred_failure_preserves_committed_quick_scan_stats -- --quiet`
- `cargo test -p wavecrate-scan -- --quiet`

@codex review

PORTALSURFER:
Addressed the latest review findings in `e7ce35e73`.

- rename destinations now age across both targeted and full quick scans: they survive one intervening full scan between watcher halves, but expire on a second full scan instead of persisting indefinitely;
- candidate-only deferred reconciliation hashes only actual rename destination paths, leaving unrelated large unhashed rows for the normal backfill lane;
- empty-source maintenance now publishes the already-committed scan refresh when deferred hashing fails, while logging the hash failure with source context;
- added focused expiry, candidate-scope, and committed-refresh regressions.

Validation:
- `cargo test -p wavecrate-scan rename_reconciliation -- --quiet` (16 passed)
- `cargo test -p wavecrate deferred_hash_failure_preserves_committed_empty_source_refresh -- --quiet`
- `cargo test -p wavecrate-scan -- --quiet` (43 passed)
- `git diff --check`

@codex review

PORTALSURFER:
Addressed all three follow-up findings in 686a77da3:

- retained rename destinations are now re-read and re-hashed before they can receive identity
- duplicate live paths remain ambiguous; facts-based disambiguation now considers every live path, not only retained candidates
- filesystem refresh reports success once `sync_paths` commits, while logging a deferred-hash failure instead of misreporting the committed sync as failed

Added regressions for replacement-before-delete and a later copy alongside an existing duplicate.

Validation:
- `cargo test -p wavecrate-scan -- --quiet` (45 passed)

@codex review

PR #794, opened 2026-07-11
OPT-1082 Make waveform detail follow zoom level

PORTALSURFER:
## Scope

- Add viewport-aware detail refinement for coarse file-backed WAV overviews.
- Decode only the current visible range on a background business worker while retaining the overview.
- Cap each request at 8,388,608 visible frames and 65,536 summary buckets.
- Coalesce rapid zoom/pan changes to one in-flight request and request the newest viewport after stale completion.
- Gate publication by path, source content revision, and exact frame viewport.
- Render refined summaries in local viewport coordinates so overlays remain frame-aligned.
- Add diagnostics and duration-independent detail, transient, single-flight, and stale-result regressions.

## Definition of Done

- Long file-backed WAVs progressively reveal finer detail when zoomed instead of magnifying overview buckets.
- Equal local content at equal samples-per-pixel produces equal peak buckets regardless of total file duration.
- Overview rendering remains immediately available during refinement.
- Work, memory, and published summaries are explicitly bounded.
- Stale source/viewport work is rejected and cannot create retry loops.

## Validation commands

- `cargo test -p wavecrate waveform_detail`
- `cargo test -p wavecrate zoomed_long_wav_refines_visible_range_independently_of_total_duration`
- scoped formatting and `git diff --check`

## Known limitations

- Progressive range refinement currently targets seekable WAV sources, which are the long-file file-backed path; other decoded formats retain their existing full decoded summaries.
- Manual DAW comparison and live latency/memory capture were unavailable while the user was away; automated equivalence, transient, bound, coalescing, and stale-identity coverage is included.

Linear: OPT-1082

### DIFF
--- a/src/native_app/app/message.rs
+++ b/src/native_app/app/message.rs
@@ -359,6 +359,7 @@ pub(in crate::native_app) enum GuiMessage {
     CancelBrowserDragOnSampleList,
     DropWaveformSelectionOnSampleList,
     Waveform(WaveformInteraction),
+    WaveformDetailRefined(crate::native_app::waveform::WaveformDetailResult),
     WaveformFileDrop(NativeFileDrop),
     Frame,
 }
--- a/src/native_app/shell/message_dispatch.rs
+++ b/src/native_app/shell/message_dispatch.rs
@@ -250,6 +250,9 @@ impl NativeAppState {
                 self.apply_navigation_dispatch(message, context);
             }
             GuiMessage::Waveform(message) => self.apply_waveform_message(message, context),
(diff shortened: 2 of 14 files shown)


PORTALSURFER:
Addressed all three review findings in 90651319a: gain envelopes are remapped into detail-local coordinates, virtual-silence and sample-slide views retain the overview coordinate domain, and path remaps recompute file-backed revisions while clearing stale detail requests/results. Added regressions for local gain mapping, virtual margins, and moved WAV refinement. `cargo test -p wavecrate signal_widget -- --nocapture` (16 passed) and all-target check pass. @codex review

PR #795, opened 2026-07-11
OPT-1079 Reconcile selected sources with disk on open

PORTALSURFER:
## Scope

- reconcile a source against disk every time it is selected, including already loaded sources
- keep the cached source tree visible while the background scan runs
- preserve valid folder selection and expansion state when the refreshed tree is published
- prevent duplicate reconciliation scans while a source scan is already active
- retain the current source tree when adding and selecting another source so it remains available as an immediate cache

## Definition of Done

- newly added and removed files appear after reopening a source without restarting Wavecrate
- source selection remains responsive by showing cached contents immediately
- repeated selection cannot queue duplicate work
- stale-result and existing scan ownership gates remain unchanged
- regression coverage exercises cached source switching and same-source disk reconciliation

## Validation commands

- `cargo test -p wavecrate reselecting_loaded_source_reconciles_disk_without_clearing_cached_tree -- --nocapture`
- `cargo test -p wavecrate selecting_loaded_cached_source_keeps_tree_visible_and_reconciles_in_background -- --nocapture`
- `cargo test -p wavecrate source_scan -- --nocapture` (17 passed; one pre-existing isolated failure in `scan_worker_flushes_discoveries_in_batches`, which observes file count 67 rather than its expected 66)

## Known limitations

- Manual GUI verification was not available in this environment.
- The unrelated source-scan worker count assertion remains failing and is not changed by this PR.


PORTALSURFER:
Addressed the latest review findings in c9629fd22: pending source placeholders are no longer parked as loaded cache trees, and cached source selection during another active scan now queues a deferred reconcile through the existing pending-refresh owner. Added focused regression tests; all-target check passes. The broader `source_scan` filter remains 19/20 with the pre-existing worker fixture count mismatch (67 vs 66). @codex review

PORTALSURFER:
Addressed all four latest selection/scan findings in 055d48f1f.

- Reselecting the source that already owns the active scan now preserves its visible tree and selection.
- Selecting a missing source during another scan marks it missing without adding a permanent pending-refresh loop.
- Uncached selections now return an explicit `Deferred` outcome, so the UI reports `scan_already_running` instead of `source_not_found`.
- Add Source now respects the workflow’s single active-scan owner; duplicate configured paths cannot launch a second worker or overwrite progress, and available existing sources can be deferred for reconcile.

Added workflow regressions for all four cases.

Validation:
- `cargo test -p wavecrate source_scan_workflow -- --quiet` (8 passed)
- cached-source background reconcile regression
- `git diff --check`

@codex review

PORTALSURFER:
Addressed the latest deferred-selection review findings in `386803d9f`.

- pending refreshes are now ordered, de-duplicated, and popped latest-first so the most recent user selection wins over older deferred clicks;
- selecting the active scan owner switches the visible browser back to that source without launching a duplicate scan;
- reselecting the already-visible loaded source preserves its live tree rather than replacing it with a placeholder;
- a deferred source that disappears before execution returns `IgnoredMissing` and is dropped instead of being requeued permanently;
- added focused regressions for switch-back, live-tree preservation, latest-first ordering, and missing-before-execution cleanup.

Validation:
- `cargo test -p wavecrate --bin wavecrate source_scan_workflow::tests -- --quiet` (10 passed)
- `git diff --check`

@codex review

PORTALSURFER:
Addressed the four latest review findings in 27415c54b:

- brand-new add-source choices are registered immediately and their scan is deferred while another scan owns the worker
- absent/removed queued sources are dropped instead of permanently requeueing
- partially discovered trees are parked and restored without treating a bare placeholder as cached content
- pending work now tracks selection and required-scan reasons separately, so clicking the active source cannot cancel an overflow/filesystem reconciliation

Added focused regression tests for all four cases.

Validation:
- `cargo test -p wavecrate --bin wavecrate source_scan_workflow::tests -- --quiet` (14 passed)
- `cargo test -p wavecrate --bin wavecrate source_scanning -- --quiet` (10 passed)

@codex review

PR #800, opened 2026-07-11
OPT-602: unify source DB mutations as typed write commands

PORTALSURFER:
## Scope

- introduce a public `SourceWriteCommand` model covering scan upserts, sample metadata, collection membership, path removal/remap, analysis identity remap, and metadata keys
- allow the same command to execute as a one-off `SourceDatabase` write or inside a caller-owned `SourceWriteBatch`
- route the existing direct convenience APIs through the typed command path for compatibility
- preserve the existing immediate transaction, revision bump, wav-path revision, WAL checkpoint, tag, and collection semantics
- split collection persistence, path mutations, and transaction commit ownership out of the oversized batch implementation
- add focused parity and revision-policy tests for direct and batched command execution

## Definition of Done

- [x] one-off and grouped writes share one typed mutation model
- [x] caller-owned transaction boundaries remain explicit through `SourceWriteBatch`
- [x] file-path-changing commands dirty the wav-path revision while metadata-only commands do not
- [x] WAL checkpoint and revision behavior is preserved
- [x] `batch.rs` is reduced from 419 lines to 287 lines and each extracted owner stays below the cleanup target
- [x] focused and package-wide source DB tests pass

## Validation commands

- `cargo test -p wavecrate-library typed_commands -- --quiet`
- `cargo test -p wavecrate-library metadata_commands_do_not_dirty_the_wav_path_revision -- --quiet`
- `cargo test -p wavecrate-library -- --quiet` (144 passed)
- `cargo test -p wavecrate --lib metadata_tag -- --nocapture` (filter currently matches 0 tests)
- `cargo check -p wavecrate --no-default-features`
- `git diff --check`

## Known limitations

- Existing named convenience methods remain as compatibility entrypoints, but their one-off implementations now delegate through `SourceWriteCommand`; callers can migrate incrementally without behavior churn.
- Raw SQLite connection migration remains separately scoped to OPT-603.

### DIFF
--- a/crates/wavecrate-library/src/sample_sources/db/mod.rs
+++ b/crates/wavecrate-library/src/sample_sources/db/mod.rs
@@ -36,6 +36,10 @@ pub use open_profiles::SourceDatabaseConnectionRole;
 pub use pending_renames::PendingRenameEntry;
 pub use types::{Rating, SampleCollection, SampleSoundType, SourceTag, SourceTagUsage, WavEntry};
 pub use util::normalize_relative_path;
+pub use write::{
+    SourceCollectionWrite, SourceContentHashWrite, SourceFileWrite, SourceTagWrite,
+    SourceWriteCommand,
+};
 
 /// Hidden filename used for per-source databases.
 pub const DB_FILE_NAME: &str = ".wavecrate.db";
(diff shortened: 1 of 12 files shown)


PORTALSURFER:
@codex review

PR #802, opened 2026-07-11
OPT-604: Make source DB hot reads schema-static

PORTALSURFER:
## Scope
- replace current-schema metadata collection and curation projections with static SQL
- retain targeted legacy-schema guards for read-only `last_curated_at` and collection lookups
- read and stage current-schema collection ownership through the canonical membership table
- use explicit current-schema SQL for writable pending-rename claims while projecting safe defaults for legacy read-only pending-rename lists
- keep legacy compatibility at writable migration/current-stamp repair boundaries and supported read-only metadata fallbacks

## Definition of Done
- current-schema collection reads treat membership rows as authoritative and ignore stale legacy-column values
- read-only legacy databases without `wav_files.last_curated_at` return `None` instead of a SQLite error
- read-only legacy databases without `wav_file_collections` project `wav_files.collection`, or an empty result when neither shape exists
- read-only legacy pending-rename rows hydrate missing optional metadata as `NULL`/`0` defaults
- writable pending-rename claim queries remain static and migration-backed
- old collection and pending-rename layouts are repaired by existing migrations
- sample metadata behavior remains covered by focused and full source-library tests

## Validation commands
- `cargo test -p wavecrate-library -- --quiet` (159 passed)
- `cargo test -p wavecrate-library legacy_read_only_pending_renames_project_optional_defaults -- --nocapture` (1 passed)
- `cargo test -p wavecrate-library collection_ -- --nocapture` (7 passed)
- `cargo test -p wavecrate-library legacy_read_only_ -- --nocapture` (4 passed)

## Known limitations
- None


PORTALSURFER:
@codex review

PORTALSURFER:
@codex review

PR #820, opened 2026-07-15
OPT-418: Adopt revision-backed projection and layout invalidation

PORTALSURFER:
## Scope
- Pin Radiant to merged PR #1422 commit `72103b62602f95fe71b6f3ce1fc58a0262ea6799`.
- Replace Wavecrate frame-clock before/after repaint snapshots with stable `SurfaceRevisions`.
- Classify structural, layout, projection, and paint-only frame changes at the application boundary.
- Move stable declarative animation paths such as source-cache progress and copy flash from full-surface refreshes to projection refreshes.

## Definition of Done
- [x] Radiant projection/layout revision stages land first.
- [x] Wavecrate pins the canonical merged Radiant commit.
- [x] The production `FrameRepaintScopeSnapshot` contract is removed.
- [x] High-frequency frame paths use monotonic revision keys and preserve correctness-first full refreshes for structural transitions.
- [x] Focused tests cover `Surface`, `Layout`, `Projection`, and `PaintOnly` routing.

## Validation
- `cargo test -p wavecrate --bin wavecrate native_app::tests::toolbar_playback::frame_overlay -q` (33 passed)
- `cargo test -p wavecrate --bin wavecrate native_app::tests::sample_browser -q` (86 passed)
- `cargo test -p wavecrate --bin wavecrate native_app::tests::waveform_playback::active_folder_cache -q` (28 passed)
- `cargo test -p wavecrate --bin wavecrate native_app::shell::message_dispatch::tests -q` (4 passed)
- `cargo test -p wavecrate --bin wavecrate native_app::tests::toolbar_playback -q` (62 passed)
- `cargo test --no-run -q`
- `bash scripts/ci.sh smoke`
- `git diff --check`

## Known limitations
None.

Radiant dependency: PORTALSURFER/radiant#1422
Linear: https://linear.app/boostnlvp/issue/OPT-418

### DIFF
--- a/src/native_app/audio/playback/progress/telemetry.rs
+++ b/src/native_app/audio/playback/progress/telemetry.rs
@@ -1,5 +1,7 @@
 use std::time::Duration;
 
+use radiant::runtime::{RepaintScope, SurfaceRevisions};
+
 use super::super::diagnostics::PlayheadFrameMessageDiagnostics;
 use crate::native_app::{
     app::{NativeAppState, SamplePlaybackSession},
@@ -8,13 +10,23 @@ use crate::native_app::{
     },
 };
 
-#[derive(Clone, Copy, Debug, PartialEq)]
-pub(in crate::native_app) struct FrameRepaintScopeSnapshot {
+#[derive(Default)]
+/// Converts frame-relevant application state into monotonic runtime revision keys.
+pub(in crate::native_app) struct FrameSurfaceRevisionTracker {
+    last: Option<FrameSurfaceRevisionInputs>,
+    revisions: SurfaceRevisions,
+}
+
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+struct FrameSurfaceRevisionInputs {
+    structure: FrameStructureState,
+    layout: FrameLayoutState,
+    projection: FrameProjectionState,
+}
+
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+struct FrameStructureState {
     playing: bool,
-    playback_visual_generation: u64,
-    play_selection_flash_active: bool,
-    copy_flash_frames: u8,
-    protected_source_error_flash_frames: u8,
     drag_hover_auto_expand_pending: bool,
     folder_progress_active: bool,
     normalization_progress_active: bool,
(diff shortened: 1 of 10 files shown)


src/native_app/audio/playback/progress/telemetry.rs
```diff
@@ -24,38 +36,64 @@ pub(in crate::native_app) struct FrameRepaintScopeSnapshot {
...
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+struct FrameSurfaceRevisionSample {
+    revisions: SurfaceRevisions,
+    scope: RepaintScope,
+    playback_involved: bool,
+}
+
+impl NativeAppState {
+    pub(in crate::native_app) fn frame_surface_revisions(&mut self) -> SurfaceRevisions {
+        let inputs = FrameSurfaceRevisionInputs::from_state(self);
```
PORTALSURFER: [PR-001][P2] This still materializes `FrameSurfaceRevisionInputs` on both revision-policy calls (before and after every frame), and that input contains essentially the same 20 fields as the removed `FrameRepaintScopeSnapshot`. The runtime boundary is typed, but Wavecrate still pays for and maintains the broad snapshot, so the DoD requiring at least one high-frequency path to no longer need it is not met. Please make a high-frequency path such as steady playback use directly maintained revision keys (with a conservative structural fallback on mode transitions), and add a regression proving repeated frames read only the stable revisions rather than rebuilding the broad input set.

PORTALSURFER (commented):
Review pass 1 found one actionable contract gap. The Wavecrate frame path exposes revisions to Radiant, but it still reconstructs the broad before/after state snapshot that OPT-418 requires at least one high-frequency path to eliminate.

src/native_app/audio/playback/progress/telemetry.rs
PORTALSURFER: Fixed in `6c73fd104f3aeb79714c6a9f254f1a3ecc99cbd6`.

Steady playback now stays on a revision fast path: the Radiant before/after callbacks read only the maintained `SurfaceRevisions`. Frame dispatch advances those keys from a narrower playback-local guard, while playback entry/exit and unknown transitions retain the conservative `Surface` fallback. The regression runs two queued playback frames and proves the broad observer count does not change.

Exact-head validation: 62 toolbar/playback tests, 28 active-folder-cache tests, 86 sample-browser tests, 4 message-dispatch tests, `cargo test --no-run -q`, and `bash scripts/ci.sh smoke` all pass. Real sandbox smoke also covered playback, row selection, context-menu open/dismiss, waveform drag selection, and resize.

PORTALSURFER (commented):
Clean re-review of the complete `origin/main...6c73fd104` diff. PR-001 is fixed and its thread is resolved. The steady-playback policy now reads stable revisions without rebuilding the broad state aggregate; entry, exit, structural, layout, projection, and paint-only behavior remain correctness-first and are covered by exact-head tests. No unresolved P0-P3 findings remain. Residual limitation: the final sandbox source-switch/scroll probe entered a long source-load state, so scroll evidence relies on the focused sample-browser suite; playback, selection, menus, waveform drag, and resize were exercised in the real app.

PR #825, opened 2026-07-16
fix: deduplicate agent preflight during merge closeout

PORTALSURFER:
## Goal
Prevent merge closeout from launching duplicate Cargo-backed agent preflight work.

## Scope and non-goals
- Hooks run bounded repository-state checks only.
- Explicit agent request/preflight retains ownership of the full preflight and coalesces concurrent callers.
- Adds Bash regression coverage and mirrors single-flight behavior in PowerShell.
- Does not change pre-commit, pre-push, CI, or GitHub merge strategy.

## Definition of Done
- Merge plus automatic/explicit checkouts cannot start duplicate full preflight runs.
- Concurrent full-preflight callers coalesce, propagate the owner result, and recover stale locks.
- Hook installation and final hook behavior are covered.

## Validation
- PASS: `bash scripts/internal/agent/test_agent_preflight_coordination.sh`
- PASS: `bash scripts/internal/check/check_script_guardrails.sh`
- PASS: `git diff --check`
- BLOCKED (infrastructure): `bash scripts/ci.sh agent` stalled twice in existing Radiant no-default compilation; retry used `CARGO_BUILD_JOBS=1 CARGO_INCREMENTAL=0` and reproduced a sleeping `rustc` with no linker progress. The process was stopped rather than leaving a stale validation run.

## Known limitations
The full agent lane needs a clean rerun on a host where the existing Cargo/Rustc stall does not occur. `pwsh` is unavailable in this environment; PowerShell syntax is covered by the repository script guardrail.


scripts/internal/agent/run_agent_preflight.sh
```diff
@@ -36,6 +37,79 @@ while (( $# > 0 )); do
...
+  local status=$?
+  release_lock "$status"
+  exit "$status"
+}
+
+trap on_exit EXIT
+trap 'exit 130' INT
+trap 'exit 143' TERM
+
+while ! mkdir "$LOCK_DIR" 2>/dev/null; do
```
PORTALSURFER: Required fix: `mkdir "$LOCK_DIR"` makes this process the owner before it publishes `$LOCK_DIR/owner` (written later at line 105). A concurrent caller that reaches this loop in that window finds no readable owner PID and takes the stale-lock path, removes the live lock, and becomes a second owner; the first process then continues and both run the full check. Publish owner metadata atomically with lock acquisition, or make callers wait/retry when the lock has no valid owner record rather than removing it. Apply the same race-free protocol to the mirrored PowerShell runner, and add a deterministic regression that exercises the acquisition/publication window.

PORTALSURFER (commented):
Required fix: the single-flight owner publication has a race that violates the PR Definition of Done.

scripts/internal/agent/run_agent_preflight.sh
PORTALSURFER: Fixed in b9555828: Bash now atomically publishes populated lock metadata via hard-link acquisition; PowerShell uses a named mutex; and the regression deterministically contends immediately after lock publication. Focused regression and script guardrails pass.

PORTALSURFER:
Fresh stateless review at b9555828aedfa5014848163645594b82d9e8b643: clean. The current diff meets the stated goal and Definition of Done: hooks are bounded to repository-state checks, explicit full preflights coalesce and return the owner status, and the Bash regression covers hook ownership plus stale-lock recovery. Verified `bash scripts/internal/agent/test_agent_preflight_coordination.sh`, `bash scripts/internal/check/check_script_guardrails.sh`, and `git diff --check` locally.

PR #826, opened 2026-07-16
OPT-1179: Make source discovery revisioned and self-healing

PORTALSURFER:
## Goal

Make source discovery authoritative, ordered, and self-healing so downstream readiness never starts from pre-commit watcher hints and missed filesystem events eventually converge.

## Scope and non-goals

- Publish a monotonic committed source revision with structured created, changed, moved, and deleted identity deltas from full scans, targeted sync, deferred hashing, and audits.
- Refresh browser projection and wake readiness only after the authoritative source transaction commits.
- Exactly hash existing targeted files, including same-size/mtime-preserving edits, while keeping new large-file hashing owned by the source-processing supervisor.
- Use one canonical creation-fenced filesystem identity format. A one-time migration removes obsolete identity shapes, converts the experimental v2 prefix, and rebuilds derived readiness state from the canonical manifest.
- Track revisioned readiness work through restart, refresh similarity results as embeddings commit, and route discovery plus processing feedback through the shared bottom-right job indicator.
- Bound watcher ingress and per-source path churn; restart failed watches with capped backoff and escalate overflow or uncertainty to authoritative scans.
- Ignore source-database and ANN temporary-file watcher noise.
- Retry targeted sync transient failures, ignore stale completions for removed sources, and avoid recreating missing source roots.
- Periodically reconcile active manifests and a rotating bounded content batch.
- Reserve shared processing capacity for admitted foreground folder scans by cancelling and requeueing incompatible lower-priority supervisor work.
- Return and publish the exact latest committed checkpoint if a later scan phase fails or is cancelled, then keep incomplete work due for retry.
- Batch harvest persistence by owning source and transaction.
- Provide a branch-specific, signed macOS development app bundle so the exact local build can be targeted independently by UI automation.
- No changes to embedding models, similarity scoring policy, playback behavior, or classification policy.

## Definition of Done

- Every discovery commit returns its exact committed revision and deterministic identity delta, including intermediate checkpoints before later cancellation.
- Targeted watcher hints cannot mutate the browser or schedule prep before commit.
- Same-size edits and identity replacements invalidate the correct generations.
- Deferred hash completion uses a platform change marker as a stable-read fence, publishes a later committed revision, and is re-reconciled by the supervisor.
- Missing identities self-heal; obsolete identity encodings are removed by the one-time canonical migration and current filesystem facts are republished.
- An analysis job that discovers stale source content triggers exact-path source reconciliation, then retries against the committed content generation.
- Watcher overflow, restart, missing/reappearing roots, and transient sync errors converge without UI-thread filesystem work.
- Successful initial watch installation does not enqueue a full scan of every source.
- External metadata remains writable for readiness availability updates while the audio root is missing.
- Periodic audits repair changes missed while Wavecrate is closed or watcher delivery is lost.
- A foreground folder scan cannot remain indefinitely at zero behind incompatible background processing.
- Background discovery is immediately visible as indeterminate activity. Source processing uses a stable one-source-at-a-time queue: each source retains source-scoped stage, file, and completed/total feedback until its runnable work drains, then the queue advances. Idle state removes the indicator entirely.
- Processing text no longer displaces the normal footer status.
- A local macOS build produces a uniquely identified app that Computer Use can address without launching the installed Wavecrate bundle.

## Validation

Current head `ba6522e12`:

- `cargo test -p wavecrate-library --lib -- --test-threads=1` — 210 passed.
- `cargo test -p wavecrate-scan --lib -- --test-threads=1` — 91 passed.
- `cargo test -p wavecrate-library commit_manifest_changes -- --test-threads=1` — 2 passed.
- `cargo test -p wavecrate-scan interleaved_manifest_writer_forces_exact_committed_resnapshot -- --test-threads=1` — passed; a later writer in the former post-commit readback window cannot contaminate the published checkpoint.
- `cargo test --bin wavecrate scheduled_manifest_audit_does_not_recreate_source_removed_after_discovery -- --test-threads=1` — passed; a missing root is parked and is not immediately requeued.
- `cargo test --bin wavecrate native_app::source_processing::scheduler::tests -- --test-threads=1` — 5 passed at the preceding exact source-processing head.
- `cargo test --bin wavecrate manifest_audit -- --test-threads=1` — 4 passed at the preceding exact source-processing head.
- `cargo test --bin wavecrate source_processing_ -- --test-threads=1` — 10 passed at the preceding exact source-processing head.
- Focused zero-byte terminal-classification and legacy unsupported-failure recovery regressions passed.
- `cargo test --bin wavecrate stale_analysis_hash_triggers_targeted_reconciliation_and_converges -- --nocapture` — passed.
- `bash scripts/build.sh --release --variant OPT-1179` — passed at `ba6522e12` and re-signed `Wavecrate OPT-1179.app` with bundle ID `org.portalsurfer.wavecrate.dev.opt1179`.
- The broader serial supervisor suite advanced through 31 tests before the existing `real_hash_execution_waits_for_shared_scan_database_budget` panic-cleanup hang; sampling showed the test process waiting in `wait_for_external_scans`, outside the changed commit/publication and parked-root paths.
- Windows cross-target checking was attempted, but this macOS host lacks `ml64.exe` and the Windows C runtime required by native dependencies; no Rust diagnostic was reached.

Live exact-bundle validation:

- The previously colliding source database migrated from experimental schema 7 to schema 8 without retry or uniqueness errors. Derived readiness rows were republished from the canonical manifest while durable analysis caches remained intact.
- The two non-empty `gwge3` WAVs reconciled to their actual BLAKE3 content hashes, and each now has a 732-byte embedding, a 1544-byte similarity aspect descriptor, and four durable completed readiness stages.
- In the active similarity view those two rows left the N/A tail. The remaining five N/A rows are genuine 0-byte WAV files; the footer reported 7,226 resolved similar samples.
- The source queue drained `samples`, advanced to `inbox`, returned only when a new `samples` prerequisite became runnable, and then reached true idle with the progress surface removed; source names no longer flickered between individual work items.
- A prior active-anchor profile reduced the bad-state roughly 96% CPU loop to roughly 13-16% idle and 6-13% during continuing multi-source work.
- Durable audit restart validation interrupted `samples` after a committed checkpoint and relaunched at 6,885/7,232 instead of zero. Completed checkpoints are cleared atomically with the audit timestamp, and the normal full-audit cadence is now 24 hours.
- Legacy AppleDouble `._*.wav` rows no longer make identity repair permanently due; the live source stopped immediately re-enqueueing full audits.
- A live 0-byte WAV that previously held similarity progress at 28,918/28,929 is now published as terminally unsupported for playback, analysis, and embedding. The queue retired it, advanced to the next source, and reached idle.
- At `ba6522e12`, the signed b6630 app remained idle with no progress indicator through a complete 30-second safety sweep. Only one OPT-1179 process was running, at about 8.1% CPU.

## Known limitations

A pre-migration file renamed while Wavecrate was closed cannot retain move classification if its database only contained an obsolete two-field identity. It is reconciled safely as delete/create; all current and future identities use the canonical creation-fenced format.

### DIFF
--- a/crates/wavecrate-library/src/sample_sources/db/write/transaction.rs
+++ b/crates/wavecrate-library/src/sample_sources/db/write/transaction.rs
@@ -37,32 +37,48 @@ impl SourceWriteBatch<'_> {
         Ok(snapshot)
     }
 
-    /// Commit the batch and return its exact revision plus only manifest paths touched by it.
+    /// Commit the batch and return its exact revision plus manifest state owned by that revision.
     ///
-    /// This keeps chunked scans linear while allowing callers to maintain the authoritative
-    /// manifest incrementally. Each optional row is read from the committing transaction after
-    /// the revision bump; `None` means the touched path is absent from the live manifest.
+    /// When the caller's cached revision is current, the second tuple element contains only
+    /// touched paths and the optional full snapshot is `None`, keeping chunked scans linear. When
+    /// another writer has advanced the manifest, the touched-path list is empty and the optional
+    /// full snapshot is captured inside this committing transaction before the write lock is
+    /// released.
     pub fn commit_with_manifest_changes(
         self,
-    ) -> Result<(u64, Vec<(PathBuf, Option<SourceManifestEntry>)>), SourceDbError> {
+        expected_previous_revision: u64,
+    ) -> Result<
+        (
+            u64,
+            Vec<(PathBuf, Option<SourceManifestEntry>)>,
+            Option<Vec<SourceManifestEntry>>,
+        ),
+        SourceDbError,
+    > {
         self.prepare_commit()?;
         let revision = manifest_revision(&self.tx)?;
-        let changes = self
-            .manifest_touched_paths
-            .iter()
-            .map(|path| {
-                let normalized = PathBuf::from(normalize_relative_path(path)?);
-                let entry = manifest_entry_for_path(&self.tx, &normalized)?;
-                Ok((normalized, entry))
-            })
-            .collect::<Result<Vec<_>, SourceDbError>>()?;
+        let (changes, snapshot) = if revision == expected_previous_revision.saturating_add(1) {
+            let changes = self
+                .manifest_touched_paths
+                .iter()
+                .map(|path| {
+                    let normalized = PathBuf::from(normalize_relative_path(path)?);
+                    let entry = manifest_entry_for_path(&self.tx, &normalized)?;
+                    Ok((normalized, entry))
+                })
+                .collect::<Result<Vec<_>, SourceDbError>>()?;
+            (changes, None)
+        } else {
+            let (_, snapshot) = manifest_snapshot(&self.tx)?;
+            (Vec::new(), Some(snapshot))
+        };
         self.tx.commit().map_err(map_sql_error)?;
         crate::sqlite_wal::maybe_checkpoint_database_file(
             &self.db_path,
             "source_db",
             self.telemetry_label,
         );
-        Ok((revision, changes))
+        Ok((revision, changes, snapshot))
     }
 
     fn prepare_commit(&self) -> Result<(), SourceDbError> {
@@ -219,6 +235,7 @@ mod tests {
             .upsert_file(Path::new("untouched.wav"), 6, 20)
             .expect("insert untouched file");
 
+        let expected_previous_revision = database.get_revision().expect("previous revision");
         let mut batch = database.write_batch().expect("manifest batch");
         batch
             .set_missing(Path::new("removed.wav"), true)
@@ -226,11 +243,12 @@ mod tests {
         batch
             .upsert_file_with_hash(Path::new("created.wav"), 7, 30, "created-hash")
             .expect("insert created file");
-        let (revision, changes) = batch
-            .commit_with_manifest_changes()
+        let (revision, changes, snapshot) = batch
+            .commit_with_manifest_changes(expected_previous_revision)
             .expect("commit manifest changes");
 
         assert_eq!(revision, database.get_revision().expect("current revision"));
+        assert!(snapshot.is_none());
         assert_eq!(changes.len(), 2);
         assert_eq!(
             changes[0],
@@ -252,15 +270,17 @@ mod tests {
     fn commit_manifest_changes_normalizes_windows_separator_paths() {
         let directory = tempfile::tempdir().expect("source root");
         let database = SourceDatabase::open(directory.path()).expect("source database");
+        let expected_previous_revision = database.get_revision().expect("previous revision");
         let mut batch = database.write_batch().expect("manifest batch");
         batch
             .upsert_file_with_hash(Path::new(r"nested\kick.wav"), 7, 30, "kick-hash")
             .expect("insert nested file");
 
-        let (_revision, changes) = batch
-            .commit_with_manifest_changes()
+        let (_revision, changes, snapshot) = batch
+            .commit_with_manifest_changes(expected_previous_revision)
             .expect("commit manifest changes");
 
+        assert!(snapshot.is_none());
         assert_eq!(changes.len(), 1);
         assert_eq!(changes[0].0, Path::new("nested/kick.wav"));
         assert_eq!(
--- a/crates/wavecrate-library/src/sample_sources/db/write/transaction.rs
+++ b/crates/wavecrate-library/src/sample_sources/db/write/transaction.rs
@@ -37,6 +37,30 @@ impl SourceWriteBatch<'_> {
         Ok(snapshot)
     }
 
+    /// Commit the batch and return its exact revision plus only manifest paths touched by it.
+    ///
+    /// This keeps chunked scans linear while allowing callers to maintain the authoritative
+    /// manifest incrementally. Each optional row is read from the committing transaction after
+    /// the revision bump; `None` means the touched path is absent from the live manifest.
+    pub fn commit_with_manifest_changes(
+        self,
+    ) -> Result<(u64, Vec<(PathBuf, Option<SourceManifestEntry>)>), SourceDbError> {
+        self.prepare_commit()?;
+        let revision = manifest_revision(&self.tx)?;
+        let changes = self
+            .manifest_touched_paths
+            .iter()
+            .map(|path| Ok((path.clone(), manifest_entry_for_path(&self.tx, path)?)))
+            .collect::<Result<Vec<_>, SourceDbError>>()?;
+        self.tx.commit().map_err(map_sql_error)?;
+        crate::sqlite_wal::maybe_checkpoint_database_file(
+            &self.db_path,
+            "source_db",
+            self.telemetry_label,
+        );
+        Ok((revision, changes))
+    }
+
     fn prepare_commit(&self) -> Result<(), SourceDbError> {
         SourceDatabase::bump_revision(&self.tx)?;
         if self.paths_revision_dirty {
@@ -49,17 +73,7 @@ impl SourceWriteBatch<'_> {
 fn manifest_snapshot(
     connection: &rusqlite::Connection,
 ) -> Result<(u64, Vec<SourceManifestEntry>), SourceDbError> {
-    let revision = connection
-        .query_row(
-            "SELECT value FROM metadata WHERE key = 'revision'",
-            [],
-            |row| row.get::<_, String>(0),
-        )
-        .optional()
-        .map_err(map_sql_error)?
-        .map(|raw| raw.parse::<u64>().map_err(|_| SourceDbError::Unexpected))
-        .transpose()?
-        .unwrap_or_default();
+    let revision = manifest_revision(connection)?;
     let filter = crate::sample_sources::supported_audio_where_clause();
     let sql = format!(
         "SELECT path, file_identity, content_hash, file_size, modified_ns
@@ -111,6 +125,56 @@ fn manifest_snapshot(
     Ok((revision, entries))
 }
 
+fn manifest_revision(connection: &rusqlite::Connection) -> Result<u64, SourceDbError> {
+    connection
+        .query_row(
+            "SELECT value FROM metadata WHERE key = 'revision'",
+            [],
+            |row| row.get::<_, String>(0),
+        )
+        .optional()
+        .map_err(map_sql_error)?
+        .map(|raw| raw.parse::<u64>().map_err(|_| SourceDbError::Unexpected))
+        .transpose()
+        .map(|revision| revision.unwrap_or_default())
+}
+
+fn manifest_entry_for_path(
+    connection: &rusqlite::Connection,
+    relative_path: &std::path::Path,
+) -> Result<Option<SourceManifestEntry>, SourceDbError> {
+    let raw_path = relative_path.to_string_lossy();
+    let row = connection
+        .query_row(
+            "SELECT path, file_identity, content_hash, file_size, modified_ns
+             FROM wav_files
+             WHERE path = ?1 AND missing = 0",
+            [raw_path.as_ref()],
+            |row| {
+                Ok((
+                    row.get::<_, String>(0)?,
+                    row.get(1)?,
+                    row.get(2)?,
+                    row.get::<_, i64>(3)?,
+                    row.get(4)?,
+                ))
+            },
+        )
+        .optional()
+        .map_err(map_sql_error)?;
+    let Some((raw_path, file_identity, content_hash, file_size, modified_ns)) = row else {
+        return Ok(None);
+    };
+    let normalized = normalize_relative_path(std::path::Path::new(&raw_path))?;
+    Ok(Some(SourceManifestEntry {
+        relative_path: PathBuf::from(normalized),
+        file_identity,
+        content_hash,
+        file_size: file_size.max(0) as u64,
+        modified_ns,
+    }))
+}
+
 #[cfg(test)]
 mod tests {
     use std::path::Path;
@@ -138,5 +202,45 @@ mod tests {
         assert_eq!(committed_manifest.len(), 1);
         assert_eq!(committed_manifest[0].relative_path, Path::new("first.wav"));
         assert!(database.get_revision().expect("current revision") > committed_revision);
+    }
+
+    #[test]
+    fn commit_manifest_changes_reports_only_touched_live_rows() {
+        let directory = tempfile::tempdir().expect("source root");
+        let database = SourceDatabase::open(directory.path()).expect("source database");
+        database
+            .upsert_file(Path::new("removed.wav"), 5, 10)
+            .expect("insert removed file");
+        database
+            .upsert_file(Path::new("untouched.wav"), 6, 20)
+            .expect("insert untouched file");
+
+        let mut batch = database.write_batch().expect("manifest batch");
+        batch
+            .set_missing(Path::new("removed.wav"), true)
+            .expect("mark file missing");
+        batch
+            .upsert_file_with_hash(Path::new("created.wav"), 7, 30, "created-hash")
+            .expect("insert created file");
+        let (revision, changes) = batch
+            .commit_with_manifest_changes()
+            .expect("commit manifest changes");
+
+        assert_eq!(revision, database.get_revision().expect("current revision"));
+        assert_eq!(changes.len(), 2);
+        assert_eq!(
+            changes[0],
+            (
+                PathBuf::from("created.wav"),
+                Some(SourceManifestEntry {
+                    relative_path: PathBuf::from("created.wav"),
+                    file_identity: None,
+                    content_hash: Some(String::from("created-hash")),
+                    file_size: 7,
+                    modified_ns: 30,
+                })
+            )
+        );
+        assert_eq!(changes[1], (PathBuf::from("removed.wav"), None));
     }
 }
--- a/crates/wavecrate-library/src/sample_sources/db/write/transaction.rs
+++ b/crates/wavecrate-library/src/sample_sources/db/write/transaction.rs
@@ -50,7 +50,11 @@ impl SourceWriteBatch<'_> {
         let changes = self
             .manifest_touched_paths
             .iter()
-            .map(|path| Ok((path.clone(), manifest_entry_for_path(&self.tx, path)?)))
+            .map(|path| {
+                let normalized = PathBuf::from(normalize_relative_path(path)?);
+                let entry = manifest_entry_for_path(&self.tx, &normalized)?;
+                Ok((normalized, entry))
+            })
             .collect::<Result<Vec<_>, SourceDbError>>()?;
         self.tx.commit().map_err(map_sql_error)?;
         crate::sqlite_wal::maybe_checkpoint_database_file(
@@ -242,5 +246,29 @@ mod tests {
             )
         );
         assert_eq!(changes[1], (PathBuf::from("removed.wav"), None));
+    }
+
+    #[test]
+    fn commit_manifest_changes_normalizes_windows_separator_paths() {
+        let directory = tempfile::tempdir().expect("source root");
+        let database = SourceDatabase::open(directory.path()).expect("source database");
+        let mut batch = database.write_batch().expect("manifest batch");
+        batch
+            .upsert_file_with_hash(Path::new(r"nested\kick.wav"), 7, 30, "kick-hash")
+            .expect("insert nested file");
+
+        let (_revision, changes) = batch
+            .commit_with_manifest_changes()
+            .expect("commit manifest changes");
+
+        assert_eq!(changes.len(), 1);
+        assert_eq!(changes[0].0, Path::new("nested/kick.wav"));
+        assert_eq!(
+            changes[0]
+                .1
+                .as_ref()
+                .map(|entry| entry.relative_path.as_path()),
+            Some(Path::new("nested/kick.wav"))
+        );
     }
 }
--- a/crates/wavecrate-library/src/sample_sources/db/write/transaction.rs
+++ b/crates/wavecrate-library/src/sample_sources/db/write/transaction.rs
@@ -1,13 +1,14 @@
-use super::super::util::map_sql_error;
-use super::super::{SourceDatabase, SourceDbError, SourceWriteBatch};
+use std::path::PathBuf;
+
+use rusqlite::OptionalExtension;
+
+use super::super::util::{map_sql_error, normalize_relative_path};
+use super::super::{SourceDatabase, SourceDbError, SourceManifestEntry, SourceWriteBatch};
 
 impl SourceWriteBatch<'_> {
     /// Commit all batched operations atomically.
     pub fn commit(self) -> Result<(), SourceDbError> {
-        SourceDatabase::bump_revision(&self.tx)?;
-        if self.paths_revision_dirty {
-            SourceDatabase::bump_wav_paths_revision(&self.tx)?;
-        }
+        self.prepare_commit()?;
         self.tx.commit().map_err(map_sql_error)?;
         crate::sqlite_wal::maybe_checkpoint_database_file(
             &self.db_path,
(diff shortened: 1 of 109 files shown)


src/native_app/source_processing/supervisor.rs
```diff
@@ -1831,6 +1869,40 @@ fn execute_candidate(
...
+                    target: "wavecrate::source_processing",
+                    source_id = candidate.source.id.as_str(),
+                    revision = outcome.committed_delta.revision,
+                    created = outcome.committed_delta.created.len(),
+                    changed = outcome.committed_delta.changed.len(),
+                    moved = outcome.committed_delta.moved.len(),
+                    deleted = outcome.committed_delta.deleted.len(),
+                    "Periodic source manifest audit committed"
+                );
+                Ok(ExecutionOutcome::Completed)
```
PORTALSURFER, line 2466: Required fix: Periodic audit deltas never reach browser projection. `audit_source_and_record` can commit created, changed, moved, or deleted rows after a watcher miss, but this branch only logs the counts and returns `Completed`; the coordinator refresh that follows only rediscovers source-processing/readiness work. Unlike the targeted-sync path, no UI message or authoritative projection refresh is emitted, so the source DB/readiness can converge while the folder tree and sample list stay stale indefinitely. Route a non-empty audit delta to a post-commit UI projection update (or schedule an authoritative background source refresh), and cover a missed watcher event becoming visible.

PORTALSURFER (commented):
Independent review of the current PR description and complete merge-base-to-head diff found three required fixes. GitHub does not permit requesting changes on a PR authored by the current identity, so this review is submitted as COMMENT; the inline Required fix threads are blocking.

src/native_app/sample_library/folder_scan_actions/filesystem_refresh.rs
```diff
@@ -93,19 +94,38 @@ impl NativeAppState {
...
+                self.ui.status.sample = format!("Synced {changed_count} filesystem change(s)");
+                self.queue_source_prep(
+                    source_id.clone(),
+                    SourcePrepTrigger::FilesystemChanged,
+                    context,
+                );
+                if self
+                    .library
+                    .folder_browser
+                    .refresh_filesystem_paths(&source_id, &projection_paths)
```
PORTALSURFER: Required fix: The newly added post-commit projection call still performs filesystem work on the UI update path. `refresh_filesystem_paths` synchronously checks each path with `is_dir`/`is_file` and can load folder/file state, while this handler runs after `SourceFilesystemSyncFinished` on the UI thread. That fails the DoD that overflow/restart/missing-root convergence occur without UI-thread filesystem work and can stall the UI on slow or unavailable source volumes. Have the background worker return projection-ready entries/snapshots, or perform the filesystem projection work in a background task and apply only owned data here.

crates/wavecrate-scan/src/sample_sources/scanner/manifest.rs
```diff
@@ -0,0 +1,265 @@
...
+    database.list_manifest_entries().map_err(ScanError::from)
+}
+
+pub(super) fn publish_committed_delta(
+    database: &SourceDatabase,
+    stats: &mut ScanStats,
+    before: Vec<SourceManifestEntry>,
+) -> Result<(), ScanError> {
+    let after = capture_manifest(database)?;
+    let revision = database.get_revision()?;
```
PORTALSURFER: Required fix: The committed manifest snapshot and its revision are read in two separate autocommit queries. A concurrent source writer can commit after `capture_manifest` and before `get_revision`, producing an `after` snapshot for revision N paired with revision N+1; the published delta then claims a revision whose manifest changes it does not contain. This violates the exact committed revision/ordered delta contract and can let consumers advance past an omitted change. Read the manifest and revision from one SQLite read snapshot, or retain/return the write revision while fencing later writers through delta capture.

src/native_app/source_processing/supervisor.rs
PORTALSURFER: Fixed in 0e22ade5a. Non-empty committed audit deltas now publish `SourceManifestAuditCommitted` through the supervisor-owned GUI channel, and the UI schedules the existing deduplicated background source refresh. Added `periodic_manifest_audit_wakes_browser_projection_after_committed_repair`; it passes.

src/native_app/sample_library/folder_scan_actions/filesystem_refresh.rs
PORTALSURFER: Fixed in 0e22ade5a. `SourceFilesystemSyncFinished` no longer calls `refresh_filesystem_paths` on the UI thread; every successful sync schedules the existing background source refresh and applies its owned result later. The updated removal regression proves the UI retains its owned tree until background completion, and the non-blocking architecture guard passes.

crates/wavecrate-scan/src/sample_sources/scanner/manifest.rs
PORTALSURFER: Fixed in 0e22ade5a. `SourceDatabase::manifest_snapshot_with_revision` reads revision and ordered manifest entries inside one SQLite read transaction, and delta publication consumes that coherent pair. `wavecrate-library` (202 tests) and `wavecrate-scan` (81 tests) pass.

crates/wavecrate-scan/src/sample_sources/scanner/manifest.rs
```diff
@@ -0,0 +1,264 @@
...
+) -> Result<Vec<SourceManifestEntry>, ScanError> {
+    database.list_manifest_entries().map_err(ScanError::from)
+}
+
+pub(super) fn publish_committed_delta(
+    database: &SourceDatabase,
+    stats: &mut ScanStats,
+    before: Vec<SourceManifestEntry>,
+) -> Result<(), ScanError> {
+    let (revision, after) = database.manifest_snapshot_with_revision()?;
```
PORTALSURFER: Required fix: Tie the published revision to the discovery commit. Callers invoke this only after their write batch has committed, then this opens a new read snapshot. A concurrent source-database write in that gap can advance `metadata.revision` (or the manifest), so `committed_delta.revision` is not necessarily the exact revision produced by this discovery operation, contrary to the Definition of Done. Capture/return the revision and final manifest from the authoritative write transaction, or retain an equivalent writer fence through publication, instead of rereading current state post-commit.

src/native_app/source_processing/supervisor.rs
```diff
@@ -1829,8 +1892,51 @@ fn readiness_target_fingerprint_with_cancel(
...
     cancel: &AtomicBool,
+    worker_sender: Option<&Sender<GuiMessage>>,
 ) -> Result<ExecutionOutcome, String> {
     let result = match &candidate.task {
+        RuntimeTask::ManifestAudit => {
+            let database_root = candidate
+                .source
+                .database_root()
+                .map_err(|error| error.to_string())?;
+            let database = SourceDatabase::open_for_background_job_with_database_root(
```
PORTALSURFER, line 2372: Required fix: Recheck source-root availability immediately before opening the writable audit database. The discovery-time `is_dir` guard has a TOCTOU gap; for normal embedded metadata `database_root` is the source root, and the writable open creates a missing database root. If the root disappears after candidate discovery but before execution, this recreates it, violating the explicit missing-root safety requirement. Treat a non-directory root as unavailable at execution time before this open, and cover the scheduled-then-removed race.

crates/wavecrate-scan/src/sample_sources/scanner/scan/runner.rs
```diff
@@ -37,6 +37,32 @@ pub fn hard_rescan(db: &SourceDatabase) -> Result<ScanStats, ScanError> {
...
+/// Reconcile a source audit and durably record its completion in the same final revision.
+pub fn audit_source_and_record(
+    db: &SourceDatabase,
+    cancel: Option<&AtomicBool>,
+    max_hashes: usize,
+    completed_at: i64,
+) -> Result<ScanStats, ScanError> {
+    let mut stats = scan(db, ScanMode::Quick, cancel, None)?;
+    let verified =
+        super::super::scan_hash::verify_content_batch(db, cancel, max_hashes, Some(completed_at))?;
```
PORTALSURFER: Required fix: Preserve and publish the scan delta when the later content-audit phase cancels or fails. `scan(...)` has already committed and returned its non-empty delta, but any error here makes the whole audit return `Err`; the supervisor sends no browser wake, and the retry starts from the already-updated manifest so its delta is empty. A missed watcher change can therefore commit without ever refreshing the browser, violating eventual self-healing and the requirement that every discovery commit return its delta. Make the phases atomic/recoverable, or return/publish the committed scan outcome even when the follow-up verification does not complete.

PORTALSURFER (commented):
Fresh stateless review of the current PR description against the complete merge-base-to-head diff. Three commit/publication boundary issues prevent the stated exact-revision and self-healing guarantees. GitHub does not permit this account to request changes on its own PR, so the blocking findings are submitted as review comments.

crates/wavecrate-scan/src/sample_sources/scanner/manifest.rs
PORTALSURFER: Fixed in 6bce83908. SourceWriteBatch now captures the revision and ordered live manifest inside the same active write transaction, after the revision bump and before COMMIT. Full scan, targeted sync, bounded content audit, and deferred hash commits build their published delta from that exact commit-owned snapshot. Added commit_snapshot_stays_bound_to_its_own_revision; wavecrate-library (203 tests) and wavecrate-scan (82 tests) pass.

src/native_app/source_processing/supervisor.rs
PORTALSURFER: Fixed in 6bce83908. Manifest-audit execution rechecks source availability immediately before writable open, and embedded writable DB path preparation no longer creates the source root or its parent after that check. Added scheduled_manifest_audit_does_not_recreate_source_removed_after_discovery; the focused app test and smoke gate pass.

crates/wavecrate-scan/src/sample_sources/scanner/scan/runner.rs
PORTALSURFER: Fixed in 6bce83908. If the manifest phase committed a non-empty identity repair and later content verification fails or is cancelled, the scan outcome is retained for publication while the audit-completion timestamp remains unchanged for retry. Supervisor publication now happens before the cancellation return. Added manifest_audit_publishes_scan_repair_when_content_verification_is_cancelled; all 82 scan tests pass.

src/native_app/source_processing/supervisor.rs
```diff
@@ -1309,6 +1390,23 @@ fn discover_source_candidates(
         return Ok(Cancellable::Cancelled);
     }
     let database_root = source.database_root().map_err(|error| error.to_string())?;
+    if !source.root.is_dir() {
+        if database_root != source.root && database_root.is_dir() {
+            let connection = SourceDatabase::open_connection_with_role_and_database_root(
```
PORTALSURFER: Required fix: handle missing audio roots with external metadata storage without reopening through a root-validating SourceDatabase API. open_connection_with_role_and_database_root rejects any root for which root.is_dir is false, so this question-mark exits with InvalidRoot and mark_readiness_temporarily_unavailable is unreachable. Such sources repeatedly fail discovery and retain active readiness state instead of converging to unavailable. Add a safe metadata-only external-DB open path that cannot recreate the audio root, and cover the external-storage missing-root case.

src/native_app/source_processing/supervisor.rs
```diff
@@ -1247,6 +1314,20 @@ fn run_coordinator(shared: Arc<Shared>) {
...
 
+fn scheduler_candidate_indices(
+    candidates: &[RuntimeCandidate],
+    external_scan_admitted: bool,
+) -> Vec<usize> {
+    candidates
+        .iter()
+        .enumerate()
+        .filter_map(|(index, candidate)| {
+            (!(external_scan_admitted && candidate.schedule.lane == ProcessingLane::Scan))
```
PORTALSURFER: Required fix: reserve the actual constrained resources needed by the foreground scan. This filter excludes only Scan candidates, but ProcessingLane::demand assigns database=1 to every lane and the global database budget is also 1. A running or repeatedly scheduled Hashing, Cleanup, DecodeSummary, or other candidate can therefore keep acquire_scan failing after admission, leaving the foreground UI queued at 0 indefinitely. Prevent or preempt all lower-priority supervisor work that competes for the foreground permit, or add a priority waiter that receives the next compatible release, and cover a non-Scan holder in the regression.

crates/wavecrate-scan/src/sample_sources/scanner/scan_hash.rs
```diff
@@ -202,7 +314,8 @@ pub(super) fn deep_hash_scan(
     stats.renames_reconciled = renamed_samples.len();
     stats.renamed_samples = renamed_samples;
 
-    batch.commit()?;
+    let committed_snapshot = batch.commit_with_manifest_snapshot()?;
```
PORTALSURFER, line 344: Required fix: generation-fence deferred hashing against mutation during the read before committing this snapshot. The loop records facts before compute_content_hash and never performs a stable-read check afterward, so an in-place writer can change the file while it is being hashed and a mixed or otherwise non-authoritative hash is then published as the committed content generation. Revalidate the file after hashing and retry or defer when the read was not stable, matching the guarded targeted-sync and audit paths, with a mutation-during-hash regression.

src/native_app/app/state/source_scan_workflow.rs
```diff
@@ -221,11 +220,9 @@ impl SourceScanWorkflow {
             return SourceFilesystemChangePlan::IgnoredSourceMissing { source_id };
         }
         if !overflowed && !paths.is_empty() {
```
PORTALSURFER, line 235: Required fix: move the availability probe immediately above this branch off the UI update path. plan_filesystem_change is called directly from GUI message dispatch, and refresh_source_availability_from_disk reaches Path::is_dir. A disconnected network or removable source can block that metadata call, contrary to the Definition of Done requiring watcher/root convergence without UI-thread filesystem work. Carry availability from watcher/background reconciliation or schedule an asynchronous root verification before mutating UI state.

crates/wavecrate-scan/src/sample_sources/scanner/scan_db_sync.rs
```diff
@@ -55,8 +56,9 @@ pub(super) fn db_sync_phase(
     if cancel_requested(cancel) {
         return Err(ScanError::Canceled);
     }
-    batch.commit()?;
-    Ok(())
+    batch
+        .commit_with_manifest_snapshot()
```
PORTALSURFER: Required fix: preserve every already-committed scan checkpoint when a later phase cancels or fails. Missing-row chunks commit earlier in this function, and walk_phase also commits 64-file batches, but this is the only manifest snapshot returned. If cancellation or an error occurs after one of those commits and before this final commit, scan returns Err; the retry captures its before manifest after the checkpoint, so those created, changed, or deleted identities can never appear in the published delta. Either make the operation atomic or retain and return a committed outcome that merges every checkpoint even when later work is incomplete.

PORTALSURFER (commented):
Changes requested at head 22dfeed163a1357d12582e5eb638fe53ffc9aea3. Five Required fix threads remain: foreground admission can still starve behind non-Scan work sharing the sole database permit; intermediate scan checkpoints can be lost from published deltas; watcher root probing still runs on the UI thread; external-metadata sources cannot mark readiness unavailable when the audio root is missing; and deferred hashes can publish an unstable content generation when a file changes during hashing. GitHub does not permit this account to formally request changes on its own pull request.

crates/wavecrate-scan/src/sample_sources/scanner/scan_hash.rs
PORTALSURFER: Fixed in a0e26ce3e. Deferred hashing now reads facts before and after the hash, includes the Unix metadata change marker in the stable-read fence, and leaves an unstable file pending. A hook-driven regression mutates same-size bytes between hash and revalidation and verifies no hash commits.

src/native_app/source_processing/supervisor.rs
PORTALSURFER: Fixed in a0e26ce3e. Missing protected-source roots now open their existing external metadata database through a dedicated unavailable-root API, update readiness to offline without recreating the audio root, and have a regression test covering the full transition.

src/native_app/source_processing/supervisor.rs
PORTALSURFER: Fixed in a0e26ce3e. Foreground admission now preempts every active source consuming incompatible shared capacity and blocks all supervisor candidates while the reservation is queued. The regression holds the database permit with Hashing, proving this is no longer limited to the Scan lane.

crates/wavecrate-scan/src/sample_sources/scanner/scan_db_sync.rs
PORTALSURFER: Fixed in a0e26ce3e. Every walk and missing-row chunk commit now captures its transaction-owned revision and manifest snapshot. If later work fails or is cancelled, scan and targeted sync return that exact checkpoint delta with an incomplete marker; consumers publish it before waking or queueing retry. Full and targeted cancellation regressions verify 64 committed creates are returned.

src/native_app/app/state/source_scan_workflow.rs
PORTALSURFER: Fixed in a0e26ce3e. The watcher thread now supplies the observed root availability with each debounced event; the GUI workflow only applies that observation. Tests cover root disappearance in watcher state and prove the GUI follows a supplied unavailable observation even while the directory still exists.

crates/wavecrate-scan/src/sample_sources/scanner/scan/runner.rs
```diff
@@ -54,11 +131,47 @@ fn scan(
...
+        Err(error) => {
+            let Some(committed_snapshot) = context.last_committed_snapshot.take() else {
+                return Err(error);
+            };
+            super::super::manifest::publish_committed_delta(
+                &mut context.stats,
+                manifest_before,
+                committed_snapshot,
+            );
+            context.stats.incomplete_error = Some(error.to_string());
```
PORTALSURFER: Required fix [P1]: Do not encode an incomplete scan as an ordinary success without migrating every caller. Once one batch commits, any later error or cancellation is converted here to Ok(ScanStats). Existing callers including the native folder-index sync, similarity-prep scan bootstrap, and controller scan completion treat every Ok as complete and do not inspect incomplete_error. A post-checkpoint failure can therefore be reported as a completed index and downstream prep can start from only the partial manifest instead of keeping the work due. Return a distinct completion type that callers must handle, or update every caller to publish the checkpoint and schedule retry before starting downstream work.

crates/wavecrate-scan/src/sample_sources/scanner/scan/context.rs
```diff
@@ -46,6 +49,15 @@ impl ScanContext {
         self.rename_candidate_generation = Some(generation);
         Ok(())
     }
+
+    pub(in crate::sample_sources::scanner) fn commit_batch(
+        &mut self,
+        batch: SourceWriteBatch<'_>,
+    ) -> Result<(u64, Vec<SourceManifestEntry>), ScanError> {
+        let snapshot = batch.commit_with_manifest_snapshot()?;
```
PORTALSURFER: Required fix [P1]: Avoid taking a full manifest snapshot on every chunk commit. commit_batch is called for each 64-file apply or missing-row batch, while commit_with_manifest_snapshot selects and sorts every live wav_files row and this line then clones that snapshot. An initial scan of N files therefore reads roughly N squared / 64 manifest rows, while holding the write transaction, and has unbounded cancellation latency for large libraries. Preserve the exact revision by accumulating the manifest or delta in memory, or return revision plus changed rows per commit and take at most one full snapshot.

crates/wavecrate-scan/src/sample_sources/scanner/manifest.rs
```diff
@@ -0,0 +1,263 @@
...
+        |entry| normalized(entry.file_identity.as_deref()),
+        &mut matched_before,
+        &mut matched_after,
+        &mut matches,
+    );
+    match_unique(
+        before,
+        after,
+        |entry| {
+            normalized(entry.file_identity.as_deref())
```
PORTALSURFER, line 69: Required fix [P2]: Preserve unchanged hard-linked paths. The file-identity pass only matches keys that are unique; if two manifest rows are hard links to the same inode, unique_indexes discards that identity. Both later fallbacks explicitly exclude entries that have a file identity, so even identical before and after manifests classify every hard link as both created and deleted. That violates the exact-delta contract and wakes projection/readiness on no-op commits. Add an unmatched fallback for the same path plus the same identity (while retaining create/delete for a true same-path identity replacement) and cover duplicate identities.

crates/wavecrate-scan/src/sample_sources/scanner/scan_walk.rs
```diff
@@ -296,9 +296,7 @@ fn prepare_for_apply(
 }
 
 fn facts_match(prepared: &PreparedFile, current: &super::scan_fs::FileFacts) -> bool {
-    current.size == prepared.facts.size
-        && current.modified_ns == prepared.facts.modified_ns
-        && current.file_identity == prepared.facts.file_identity
+    current.same_file_facts(&prepared.facts)
```
PORTALSURFER, line 363: Required fix [P2]: Apply the new change marker to the targeted hashing fence. read_facts records ctime and same_content_snapshot compares it, but this helper only checks size, mtime, and file identity. If another application rewrites the same inode with the same length and restores mtime during compute_content_hash, both the before/after checks pass and the targeted transaction can publish a torn or stale hash. Compare the content snapshot marker around hashing and again before apply, with a mutation-during-targeted-hash regression.

PORTALSURFER (commented):
Changes requested: four required fixes block the PR contract at this head. The authenticated identity is the PR author, so GitHub does not permit a formal REQUEST_CHANGES state; the inline comments are the blocking review record.

crates/wavecrate-library/src/sample_sources/db/write/transaction.rs
```diff
@@ -16,4 +17,230 @@ impl SourceWriteBatch<'_> {
...
+        .map(|raw| raw.parse::<u64>().map_err(|_| SourceDbError::Unexpected))
+        .transpose()
+        .map(|revision| revision.unwrap_or_default())
+}
+
+fn manifest_entry_for_path(
+    connection: &rusqlite::Connection,
+    relative_path: &std::path::Path,
+) -> Result<Option<SourceManifestEntry>, SourceDbError> {
+    let raw_path = relative_path.to_string_lossy();
```
PORTALSURFER, line 166: Required fix [P1]: Normalize touched paths before reading them back. All wav-file mutations store paths through normalize_relative_path, which converts Windows separators to '/', but this lookup binds relative_path.to_string_lossy() directly. On Windows a normal nested path such as folder\kick.wav is stored as folder/kick.wav, so the query returns None; commit_with_manifest_changes then removes or omits the live row from the in-memory committed manifest. Initial nested adds can publish no created delta and nested edits can be reported as deletes. Bind normalize_relative_path(relative_path)? (and use the normalized entry path as the map key), then add a Windows-separator regression.

crates/wavecrate-scan/src/sample_sources/scanner/scan/context.rs
```diff
@@ -46,6 +59,32 @@ impl ScanContext {
         self.rename_candidate_generation = Some(generation);
         Ok(())
     }
+
+    pub(in crate::sample_sources::scanner) fn commit_batch(
+        &mut self,
+        batch: SourceWriteBatch<'_>,
+    ) -> Result<u64, ScanError> {
+        let (revision, changes) = batch.commit_with_manifest_changes()?;
```
PORTALSURFER: Required fix [P1]: This checkpoint only applies paths touched by the scan's own batch to a manifest captured before scanning. Native rename, drag-drop, and metadata writers can commit wav_files changes without this scan lane, so another writer can commit unrelated path X between capture and this batch; SQLite serializes the writes, but the returned snapshot at the later revision still omits X. That breaks the exact revision/delta contract. Serialize every manifest writer with the scan, or detect interleaved manifest revisions and resnapshot/use a change log before publishing while retaining the linear uncontended path.

src/app/controller/library/scans.rs
```diff
@@ -11,6 +11,19 @@ mod worker;
...
+        let Some(source) = self
+            .library
+            .sources
+            .iter()
+            .find(|source| &source.id == source_id)
+            .cloned()
+        else {
+            return;
+        };
+        self.request_scan_for_source(&source, mode, ScanKind::Auto);
```
PORTALSURFER: Required fix [P2]: Preserve targeted paths when retrying an incomplete targeted scan. This calls request_scan_for_source with only the mode; that path passes None to the worker, and Targeted unwraps it to an empty path list. After a checkpointed multi-chunk watcher sync fails or is canceled, the advertised retry therefore scans none of the remaining paths and the consumed watcher hint is lost until a later broad audit. Carry the original target set in retry state/result, or promote incomplete Targeted work to an authoritative Quick scan.

crates/wavecrate-scan/src/sample_sources/scanner/scan/runner.rs
```diff
@@ -37,6 +37,81 @@ pub fn hard_rescan(db: &SourceDatabase) -> Result<ScanStats, ScanError> {
...
+    audit_source_and_record_after_scan(db, cancel, max_hashes, completed_at, after_scan)
+}
+
+fn merge_audit_verification(
+    stats: &mut ScanStats,
+    verification: Result<ScanStats, ScanError>,
+) -> Result<ScanStats, ScanError> {
+    match verification {
+        Ok(verified) => stats.merge_deferred_hashes(verified),
+        Err(error) if !stats.committed_delta.is_empty() => {
```
PORTALSURFER: Required fix [P2]: Treat an empty manifest delta with a nonzero revision as a committed checkpoint. An unchanged scan still commits completion metadata and advances revision, but is_empty() ignores revision. If subsequent audit hashing fails or is canceled, this falls through to raw Err and discards the exact committed revision instead of returning Incomplete, contrary to the checkpoint contract. Gate on committed revision or checkpoint presence rather than created, changed, moved, and deleted counts.

PORTALSURFER (commented):
Changes requested: four required fixes block the PR's exact revision/delta and retry guarantees. The authenticated identity owns this PR, so GitHub does not allow a formal REQUEST_CHANGES event; this COMMENT review is blocking.

crates/wavecrate-library/src/sample_sources/db/write/transaction.rs
PORTALSURFER: Fixed in aa8d019f5: touched paths are normalized before transactional readback and used as normalized map keys; the Windows-separator regression is in transaction.rs.

crates/wavecrate-scan/src/sample_sources/scanner/scan/context.rs
PORTALSURFER: Fixed in aa8d019f5: ScanContext tracks the captured manifest revision, detects any interleaved revision before applying a linear delta, and resnapshots the authoritative manifest/revision. The interleaved writer regression covers the external path.

src/app/controller/library/scans.rs
PORTALSURFER: Fixed in aa8d019f5: incomplete targeted scans are promoted to an authoritative Quick retry, so consumed watcher targets cannot collapse into an empty targeted retry.

crates/wavecrate-scan/src/sample_sources/scanner/scan/runner.rs
PORTALSURFER: Fixed in aa8d019f5: incomplete audit publication gates on committed_delta.revision > 0, including unchanged checkpoints with an empty identity delta; the cancellation regression covers this case.

PORTALSURFER (commented):
Changes requested. GitHub does not permit this account to request changes on its own PR; the two inline comments below are blocking Required fixes.

crates/wavecrate-scan/src/sample_sources/scanner/scan_fs.rs
```diff
@@ -120,9 +133,22 @@ pub(super) fn read_facts(root: &Path, path: &Path) -> Result<FileFacts, ScanErro
...
 
+#[cfg(unix)]
+fn metadata_change_marker(metadata: &fs::Metadata) -> Option<(i64, i64)> {
+    use std::os::unix::fs::MetadataExt;
+
+    Some((metadata.ctime(), metadata.ctime_nsec()))
+}
+
+#[cfg(not(unix))]
+fn metadata_change_marker(_metadata: &fs::Metadata) -> Option<(i64, i64)> {
```
PORTALSURFER: Required fix: `metadata_change_marker` returns `None` on Windows, so `same_content_snapshot` can accept a same-size mutation whose mtime is restored while hashing. That breaks the stable-read fence and same-size-edit guarantee on a supported platform. Add a Windows change marker that changes on content mutation (for example file change time obtained from Windows file information), fail safely when no marker is available, and cover mutation-during-hash with preserved size/mtime.

crates/wavecrate-scan/src/sample_sources/scanner/manifest.rs
```diff
@@ -0,0 +1,402 @@
...
+pub(super) fn build_committed_delta(
+    before: &[SourceManifestEntry],
+    after: &[SourceManifestEntry],
+    revision: u64,
+) -> CommittedSourceDelta {
+    let mut matched_before = BTreeSet::new();
+    let mut matched_after = BTreeSet::new();
+    let mut matches = Vec::new();
+
+    match_unique(
```
PORTALSURFER, line 43: Required fix: exact identity matching happens before the only legacy-v2 compatible comparison, and that compatible comparison is restricted to the same path. A legacy `unix:<dev>:<ino>` row renamed while being upgraded to v2 therefore becomes created+deleted rather than moved. Add a deterministic uniqueness-checked compatibility pass across paths using `same_filesystem_object_identity`, preserve ambiguity safety for hard links, and add a renamed legacy-row regression.

crates/wavecrate-scan/src/sample_sources/scanner/scan_fs.rs
PORTALSURFER: Fixed in 167a80853. Stable reads now use a platform change marker: Unix ctime and Windows FILE_BASIC_INFO.ChangeTime. Missing markers fail closed, and the mutation regression preserves size and restores mtime.

crates/wavecrate-scan/src/sample_sources/scanner/manifest.rs
PORTALSURFER: Superseded by the amended PR contract and the requested removal of runtime legacy compatibility. 167a80853 uses one canonical creation-fenced identity only; schema v8 converts the experimental v2 prefix, clears obsolete two-field identities, and rebuilds derived readiness state. The documented limitation is that a rename made while closed before migration reconciles safely as delete/create.

PORTALSURFER (commented):
Fresh stateless review of the amended PR description against the complete merge-base-to-head diff at 167a80853d71a0b091b9d176025569ccfe62a1c4: clean. The current diff achieves the Goal and Definition of Done. Required fixes: none. Non-blocking observations: none. GitHub does not permit this account to approve its own pull request.

PORTALSURFER (commented):
Fresh stateless review of the current PR description against the complete merge-base-to-head diff at ba6522e1201d27e535cf084c54e2b631fcba20dd: clean. The diff achieves the Goal and Definition of Done. Required fixes: none. Non-blocking observations: none. GitHub does not permit this account to approve its own pull request.

PR #828, opened 2026-07-17
OPT-1181: Make similarity artifacts converge automatically and exactly

PORTALSURFER:
## Goal

Make native similarity processing converge automatically and exactly through one supported current pipeline, with no legacy similarity builders, queues, artifact formats, fallback loaders, or manual drain path.

Linear: OPT-1181

## Scope and non-goals

- Make the readiness-managed source-processing supervisor the sole owner of feature, embedding, aspect, descriptor, ANN, layout, and cluster convergence.
- Base readiness and publication on committed content identities, the exact eligible source manifest, and explicit algorithm/model/schema generations.
- Treat deferred hashes as invalidations of fallback-derived work and schedule exact recomputation.
- Publish ANN/layout/cluster artifacts atomically from the current manifest; replace changed vectors and prune deleted or stale identities.
- Remove the legacy similarity-prep controller, status flow, manual enqueue/drain paths, incremental/read-time ANN mutation, split-file ANN migration/fallbacks, administrative builders, benchmark, and workspace tool.
- Prune retired similarity jobs, progress snapshots, ANN files, and dirty/timestamp metadata during current publication.
- Keep Process Source as a priority wake for automatic reconciliation, not a separate processing implementation.
- Non-goals: change the current similarity algorithms, query controls, or general source-database compatibility outside the similarity subsystem.

## Definition of Done

- Every active source automatically reconciles toward an exact, versioned similarity-ready state without manual seeding or draining.
- Only readiness-managed jobs are recovered as current work; retired similarity jobs and progress state are removed.
- Missing or stale current ANN generations fail closed instead of loading unnamed, legacy, or database-derived fallback state.
- ANN publication replaces changed vectors, removes stale identities, and removes retired persisted formats and metadata.
- Layout and clusters are computed only from the same exact eligible manifest as the published ANN generation.
- Deferred identity resolution and committed-generation fences prevent false-ready or stale publication.
- The application, documentation, and tests expose only the automatic current-system behavior.

## Validation

- `CARGO_TARGET_DIR=/tmp/wavecrate-opt1181-target bash scripts/ci.sh agent` — passed (`[ci_agent] OK`; 1,855 application library tests passed, 20 ignored, 2 filtered out; Radiant and runtime suites also passed)
- `cargo test -p wavecrate-analysis similarity_artifacts -- --nocapture` — passed (6 tests)
- `cargo test -p wavecrate-analysis` — passed (81 tests)
- `cargo test -p wavecrate --bin wavecrate retired_jobs_are_pruned_and_only_readiness_jobs_are_recovered` — passed
- `cargo test -p wavecrate --bin wavecrate native_app::tests::sample_browser::similarity` — passed (11 tests)
- `cargo test -p wavecrate --bin wavecrate production_supervisor_publishes_claims_and_completes_readiness_without_manual_seed` — passed
- `cargo test -p wavecrate --test source_finalizer_process` — passed
- `cargo check -p wavecrate-analysis --all-targets` — passed
- `cargo check -p wavecrate --all-targets` — passed
- `git diff --check` — passed

An additional workspace-wide all-target diagnostic reaches existing unrelated compile warnings-as-errors in Radiant popup-window examples; the supported Wavecrate agent gate and scoped all-target checks pass.

## Known limitations

None.


src/native_app/sample_library/similarity_prep/worker.rs
```diff
@@ -618,6 +936,49 @@ pub(in crate::native_app) fn resolve_similarity_prep_status(
...
+    }
+    let mut failed_count = 0usize;
+    let mut unsupported_count = 0usize;
+    for entry in &snapshot.entries {
+        match entry.classification {
+            ReadinessClassification::PermanentFailure { .. } => {
+                failed_count = failed_count.saturating_add(1);
+            }
+            ReadinessClassification::Unsupported => {
+                unsupported_count = unsupported_count.saturating_add(1);
```
PORTALSURFER: **Required fix** — A source that has only terminal-unsupported inputs cannot report that state. `snapshot.is_fully_ready()` returns `UpToDate` before classifications are counted; otherwise this branch returns `Blocked` only when `failed_count > 0`, so an unsupported-only snapshot falls through to `Outdated`. The PR Definition of Done explicitly requires status to report terminal unsupported content correctly. Classify unsupported entries before choosing the status and return a status/UI payload that preserves the unsupported count even when there are no permanent failures; add a regression for an unsupported-only source.

PORTALSURFER (commented):
Required fix identified: the current diff does not meet the Definition of Done for reporting terminal unsupported content. (GitHub does not permit this identity to request changes on its own pull request.)

src/native_app/sample_library/similarity_prep/worker.rs
PORTALSURFER: Fixed in 0caf6a429. Readiness classifications are now counted before choosing the status. Exact-ready sources with terminal unsupported entries report `ReadyWithUnsupported { unsupported_count }`; non-ready sources preserve the count via `OutdatedWithUnsupported`; both remain distinct from permanent-failure `Blocked`. The ready helpers also keep excluded unsupported inputs from re-enqueueing/finalizing. Added an unsupported-only source regression. Validation: focused worker suite 19 passed / 1 ignored, and exact-head `bash scripts/ci.sh agent` passed with 1,919 Wavecrate tests.

crates/wavecrate-analysis/src/analysis/ann_index/mod.rs
```diff
@@ -352,6 +363,155 @@ pub fn rebuild_index_with_publication_fence(
...
+    if let Some(existing) = registry.get(&key).cloned() {
+        drop(registry);
+        drop(placeholder_state);
+        return publish_exact_into_existing_state(
+            conn,
+            &mut state,
+            existing,
+            publication_fence,
+            &mut publish_sql_artifacts,
+        );
```
PORTALSURFER, line 242: **Required fix** — This registers the DB-derived placeholder before the publication transaction has passed its fence. If `publish_exact_transaction` returns `false` (or errors), the function returns without removing/replacing that entry, so a process with no prior cached ANN index will subsequently serve the placeholder—built from the pre-publication `embeddings` table, including identities the rejected exact generation was meant to replace/prune. That exposes an ANN state that did not pass the committed manifest/membership fence, violating the stale-publication requirement.

Keep a placeholder private until commit succeeds, or remove this exact registry entry on every rejected/error path before returning; only retain/install the state after a successful commit. Add a regression that evicts the cache, rejects the fence, and verifies no placeholder ANN state is available/exposed afterward.

PORTALSURFER (commented):
## Required fixes

The current diff does not yet meet the Definition of Done’s stale-fence guarantee: a rejected exact publication can expose an uncommitted ANN state. Please address the inline finding and add a regression covering the no-cached-index / rejected-fence path.

`REQUEST_CHANGES` is unavailable because this GitHub identity authored the PR.

crates/wavecrate-analysis/src/analysis/ann_index/mod.rs
PORTALSURFER: Fixed in 6ac573a04. On a first-load exact publication, the ANN registry now remains write-locked and empty until the fenced SQL/ANN transaction commits; rejected and errored paths unwind without installing any state, while concurrent cache misses cannot publish database-derived state during the transaction. Added the requested evicted-cache + rejected-fence assertion. Validation: all 100 `wavecrate-analysis` tests passed and exact-head `bash scripts/ci.sh agent` passed with 1,919 Wavecrate tests.

crates/wavecrate-analysis/src/analysis/ann_index/build.rs
```diff
@@ -26,6 +27,7 @@ pub(crate) fn load_or_build_index(conn: &Connection) -> Result<AnnIndexState, St
         let mut state = outcome.state;
         if outcome.needs_migration {
             super::update::flush_index(conn, &mut state)?;
+            remove_superseded_generation(&meta_row.index_path, &state.index_path);
```
PORTALSURFER: Required fix: Exact publication writes `ann_index_meta.index_path` to a generation-specific `similarity_hnsw.<generation>.ann`, but `load_index_from_disk` still opens `default_index_path()` (`similarity_hnsw.ann`) rather than the metadata path. After an upgrade/restart, any existing default container can therefore be loaded instead of the just-published generation, reintroducing stale vectors or identities and defeating the stated exact/current restart guarantee. Load the container from `meta.index_path` (with an explicit legacy fallback/migration only when appropriate), and add a restart regression seeded with a stale default container plus a newer generation-specific published container.

PORTALSURFER (commented):
## Review summary

The current diff has one blocking exact-generation restart defect.

crates/wavecrate-analysis/src/analysis/ann_index/build.rs
PORTALSURFER: Fixed in fa66c6cc3. The loader already tries `ann_index_meta.index_path` first; this change makes an exact generation path strictly authoritative by preventing both default-container and legacy fallback if that generation is unavailable. The caller then rebuilds from the committed embeddings table instead of serving stale vectors. Added a regression with a stale default container plus a newer generation-specific container, proving the metadata generation wins and a missing generation never falls back stale. Validation: all 101 `wavecrate-analysis` tests passed and exact-head `bash scripts/ci.sh agent` passed with 1,919 Wavecrate tests.

PORTALSURFER (commented):
Clean review — reviewed head `fa66c6cc31a558f7ae4b5af6f8f1d2028f5f6ec9` against the complete merge-base diff from `2cedebd04629d31ceaee4b88367b766ef1d37879`. The diff achieves the stated Goal and Definition of Done: it fences exact current readiness, invalidates deferred identity work, atomically publishes the exact ANN/layout/cluster bundle while pruning stale rows, reports terminal unsupported state, and includes targeted regression coverage. No required fixes found.

crates/wavecrate-analysis/src/analysis/ann_index/build.rs
```diff
@@ -1,52 +1,46 @@
...
+    let meta = read_meta(conn, &params.model_id)?
+        .ok_or_else(|| "Current ANN generation has not been published".to_string())?;
+    if meta.params != params {
+        return Err("Current ANN generation uses a different contract version".to_string());
     }
-    let index_path = default_index_path(conn)?;
-    let mut state = build_index_from_db(conn, params, index_path)?;
-    super::update::flush_index(conn, &mut state)?;
-    Ok(state)
+    load_container_index(&meta.index_path, &meta.params)?
```
PORTALSURFER: Required fix — This still accepts the retired unnamed ANN container as a current generation. `load_current_index` trusts any `ann_index_meta.index_path` whose parameters match; the diff retains the prior unnamed path as `similarity_hnsw.ann` (storage.rs:6, 70-78). Therefore a pre-existing database whose metadata points at that valid container will load it successfully after upgrade, before a readiness-managed publication replaces it. That contradicts the Goal/DoD requirement to remove legacy fallback state and fail closed rather than load unnamed or legacy generations. Require an explicitly current generation pointer (tied to the current similarity artifact contract/state and generation-named container), or invalidate old metadata so the supervisor recomputes it; add a regression that seeds legacy metadata plus `similarity_hnsw.ann` and verifies lookup fails until current publication.

PORTALSURFER (commented):
Required fix: the current-head diff does not yet meet the current-system-only ANN fail-closed Definition of Done.

crates/wavecrate-analysis/src/analysis/ann_index/build.rs
PORTALSURFER: Fixed in `14c7b3817`. Restart loading now requires the current `similarity_artifact_state_v1` record, matching current model, and the exact generation-derived container path; legacy `ann_index_meta` pointing at `similarity_hnsw.ann` fails closed. Exact current publication also removes that retired unnamed container. Added a regression that seeds valid legacy metadata/container state, proves lookup fails, publishes the current generation, proves the old container is removed, and verifies restart succeeds. Validation: all 80 `wavecrate-analysis` tests passed and exact-head `bash scripts/ci.sh agent` passed with 1,855 application library tests.

PORTALSURFER (commented):
Required fix identified; unable to request changes because this identity owns the pull request.

crates/wavecrate-analysis/src/analysis/ann_index/mod.rs
```diff
@@ -46,10 +54,10 @@ fn get_index_entry(conn: &Connection) -> Result<Arc<RwLock<state::AnnIndexState>
         }
     }
 
-    // Slow path: load or build index without holding global lock
+    // Slow path: load the published current index without holding global lock
     // Note: This might do redundant work if multiple threads race here,
     // but better than blocking the world.
-    let loaded_state = build::load_or_build_index(conn)?;
+    let loaded_state = build::load_current_index(conn)?;
```
PORTALSURFER, line 60: Required fix\n\nThe fast cache path above this load returns the previous `AnnIndexState` without checking the current artifact metadata. In the same diff, source-target publication writes `similarity_artifact_state_v1` as `state: "dirty"` whenever the manifest changes (src/native_app/source_processing/supervisor.rs:2249-2262), but a query in that same process continues to use the already-cached old ANN until finalization happens to replace it. That serves a stale generation instead of failing closed, violating the stated stale-generation and false-ready guarantees. Associate the cache entry with the published artifact generation (or validate/evict it before the fast-path return) and add a regression test that queries once, marks the generation dirty, then verifies the next query is rejected until current publication succeeds.

PORTALSURFER: Fixed in `c63f7991a`. Dirty-state publication now takes the ANN registry and per-source state write locks, persists the dirty bundle state, and evicts the cached generation before releasing them. Searches validate the persisted current generation while holding the matching state read lock, so searches already in flight finish before dirty commits and every later query fails closed until exact publication replaces the generation. Added the requested warm-cache → dirty → rejection → current-publication regression. Validation: all 81 `wavecrate-analysis` tests passed and exact-head `bash scripts/ci.sh agent` passed with 1,855 application library tests.

PORTALSURFER (commented):
Reviewed head `c63f7991a73465f15eda441250ee287d72a5d6bf` against the current PR description and complete merge-base-to-head diff.

Clean: the diff implements the stated current-system-only automatic similarity convergence contract and Definition of Done. No required fixes found.

PR #830, opened 2026-07-18
Wavecrate: cancel source work and clean derived state without losing fast re-add

PORTALSURFER:
## Goal

Make source removal an immediate, generation-fenced lifecycle transition that abandons old work safely, releases volatile/path-owned state, and preserves authoritative source data for fast re-add.

## Scope and non-goals

- Add exact source runtime lifecycle generations across supervisor admissions, external scans, in-flight work, and UI progress projection.
- Retire removed epochs asynchronously after exact work drains, with retry/backoff and startup recovery from retained inactive descriptors.
- Preserve source databases, manifests, reusable analysis/similarity state, and content-addressed caches; remove readiness-managed jobs, playback-summary ownership, unshared managed cache payloads, and source-owned in-memory waveform/preview queues.
- Restore complete retained source descriptors, including protected/AppData policy, and fail closed on corrupt present values.
- Release the global scan workflow and source-owned native tasks on removal without waiting for cleanup.
- Keep individual-file deletion on its existing exact manifest/similarity cleanup path; no source folder, audio, source DB, synchronous vacuum, or unrelated cleanup changes.

## Definition of Done

- Removal returns while old scan/publication work is still held, cancels old admission, and prevents late epoch publication.
- Full-scan removal cannot leave the global scan lane active.
- Same-database remove/re-add waits only at new processing admission until the old epoch drains, suppresses destructive retirement, and preserves active readiness.
- True retirement deletes only volatile/readiness playback ownership and unshared managed cache refs while retaining reusable source artifacts.
- Missing/read-only storage retries asynchronously; interrupted cleanup is rediscovered at startup.
- Native cache/task/progress state is purged by source ownership and late UI progress from an older re-added epoch is rejected.
- Protected/AppData descriptors survive removal/re-add; unknown persisted enum values reject the add instead of defaulting.
- Targeted sync admits the exact configured descriptor and role/storage changes advance the source generation before later work can publish.

## Validation

- `cargo check --all-targets` — passed
- `cargo test -p wavecrate-library sample_sources::library::tests` — 19 passed
- `cargo test --bin wavecrate source_processing::supervisor::tests:: -- --test-threads=1` — 53 passed, 1 profile test ignored
- `cargo test --bin wavecrate native_app::shell::message_dispatch::tests:: -- --test-threads=1` — 9 passed
- `cargo test --bin wavecrate native_app::tests::config_sources::scan_handoff:: -- --test-threads=1` — 4 passed
- protected extraction target-source role transition regression — passed
- focused source-scan, waveform runtime eviction, and stale UI epoch tests — passed
- `./scripts/check.sh non-blocking-architecture` — passed
- `git diff --check` — passed

## Known limitations

None.

Linear: OPT-1183


src/native_app/source_processing/supervisor.rs
```diff
@@ -1516,6 +1743,11 @@ fn publish_source_processing_progress(
     let _ = worker_sender.send(GuiMessage::SourceProcessingProgress(
         SourceProcessingProgress {
             source_id: candidate.source.id.as_str().to_string(),
+            lifecycle_generation: control
```
PORTALSURFER: Required fix: carry the lifecycle generation captured by `begin_in_flight_work` through every coordinator progress publication instead of reading `control.source_lifecycle_generations` at send time. The candidate may finish, be removed, and be re-added before the post-execution calls at lines 1460/1489 run; this lookup then stamps that old candidate's progress with the new epoch, which `SourceProcessingProgress` accepts. That violates the stated DoD that late UI progress from an older re-added epoch is rejected. Thread `in_flight_work.lifecycle_generation` into `publish_source_processing_progress` (and use it at this field), then add a race-focused regression test for remove/re-add between candidate execution and the coordinator's completion progress publish.

PORTALSURFER (commented):
Reviewed head `73112fdf17a7731c089919ddf32cc4a949320cc9`: **changes requested**. GitHub does not permit this identity to submit a formal request-changes review on its own PR; the required fix is inline.

src/native_app/source_processing/supervisor.rs
PORTALSURFER: Fixed in 575d87c7c. Coordinator progress publication now carries the lifecycle generation captured by begin_in_flight_work through the pre-execution and both post-execution paths, so a removed/re-added source cannot relabel old candidate progress with the new epoch. Added late_progress_preserves_executing_generation_across_remove_and_readd to cover that exact race. Validation on the pushed head: focused regression passed; supervisor suite passed (51 passed, 1 ignored profile test); cargo check --all-targets passed; git diff --check passed.

PORTALSURFER (commented):
Clean review of `575d87c7ca0188cf6b0c8261e54f82d11690f382`: the current merge-base diff meets the PR Goal, Scope, and Definition of Done. It generation-fences admissions, in-flight work, and progress; releases the scan lane and source-owned state immediately; retires derived playback/readiness ownership asynchronously with recovery; preserves retained descriptors and reusable artifacts for fast re-add; and covers stale re-added-epoch progress. No required fixes or non-blocking observations.

src/native_app/source_processing/supervisor.rs
```diff
@@ -890,14 +1003,27 @@ struct ControlState {
...
     fn source_is_active(&self, source_id: &str) -> bool {
-        self.sources.contains_key(source_id) && !self.quarantined_sources.contains(source_id)
+        let Some(source) = self.sources.get(source_id) else {
+            return false;
+        };
+        !self.quarantined_sources.contains(source_id)
+            && !self
+                .pending_retirements
+                .values()
+                .any(|retirement| source_storage_identity_matches(source, &retirement.source))
```
PORTALSURFER, line 1047: Required fix (High): Preserve the re-add request at this admission fence. With this change, a configured source sharing storage with a pending retirement is reported as inactive. `acquire_scan` treats that state as terminal and returns `None`; the added `fast_readd_waits_for_old_epoch_then_preserves_shared_source_storage` test explicitly asserts that result. The production folder-scan worker converts `None` into a cancelled scan, so a same-database fast re-add loses its initial scan instead of waiting at new processing admission until the old epoch drains, contrary to the Definition of Done. Distinguish configured-but-blocked-by-an-exact-retirement from removed/quarantined, wait on the lifecycle wake until that retirement drains (or the generation/configuration changes or shutdown begins), then admit the current generation. The regression should hold an old guard, start the new admission on another thread, prove it remains pending, release the old guard, and prove the waiter receives the new-generation permit.

src/native_app/source_processing/supervisor.rs
```diff
@@ -1536,6 +1781,11 @@ fn publish_source_processing_discovery(shared: &Shared, source: &SampleSource) {
     let _ = worker_sender.send(GuiMessage::SourceProcessingProgress(
         SourceProcessingProgress {
             source_id: source.id.as_str().to_string(),
+            lifecycle_generation: control
```
PORTALSURFER: Required fix (Medium): Carry the captured epoch into discovery publication instead of reading the current generation here. `sources_to_discover` is built from a coordinator snapshot after releasing `ControlState`; removal plus same-ID re-add can replace that source generation before this function reacquires the lock. This lookup then labels the old snapshot with the new generation, and the new UI fence accepts the late old-epoch progress—the exact remove/re-add case the Definition of Done says to reject. Snapshot `(source, lifecycle_generation, cancel identity)` together, validate that exact epoch before publishing, and add a race regression that replaces the source between snapshot and discovery publication.

src/native_app/source_processing/supervisor.rs
```diff
@@ -73,23 +81,39 @@ pub(in crate::native_app) struct SourceProcessingBudgetPermit {
     shared: Arc<Shared>,
     permit: Option<super::scheduler::BudgetPermit>,
     registration_id: u64,
+    lifecycle_generation: u64,
```
PORTALSURFER, line 84: Required fix (Medium): Propagate this epoch through source-scoped completion delivery, not only permit execution. The external worker drops the permit before `SourceFilesystemSyncFinished` is dispatched, but that result contains only `source_id`; its handler accepts any re-added source with that ID and, for an old cancelled result, calls `wake_source`, cancelling the new epoch. `execute_candidate` likewise receives a generation but emits `SourceManifestAuditCommitted` without it, and that handler also checks only source existence. Thus epoch 1 can drain, unblock epoch 2, and then mutate/cancel epoch 2 through a late completion. Add lifecycle generation to these completion payloads (and compare it with `source_lifecycle_generations` before side effects), with remove/re-add tests that deliver an old completion after epoch 2 is current.

PORTALSURFER (commented):
Unsafe/concurrency specialist review of the PR description against the complete merge-base diff. Changes are required for same-storage re-add admission and end-to-end epoch publication fencing. The core exact-generation drain/finalization checks otherwise use a coherent lock order. GitHub cannot record REQUEST_CHANGES because the authenticated reviewer is also the PR author, so the blocking findings are submitted as Required fix inline comments.

src/native_app/source_processing/supervisor.rs
PORTALSURFER: Fixed in 311ab0b4c. acquire_scan now distinguishes a configured source blocked by same-storage retirement from a removed/quarantined source: the new request waits on the lifecycle condition, and retirement handoff wakes it after the exact old epoch drains. The updated fast_readd_waits_for_old_epoch_then_preserves_shared_source_storage regression holds old work, proves the new admission remains pending, drains retirement, and verifies the returned permit carries the current generation. Supervisor suite: 52 passed, 1 ignored.

src/native_app/source_processing/supervisor.rs
PORTALSURFER: Fixed in 311ab0b4c. Coordinator discovery now snapshots source descriptor, lifecycle generation, and cancel-token identity together, validates that exact snapshot before publication, and publishes the captured generation. Added discovery_snapshot_from_previous_readded_epoch_is_not_published for remove/re-add between snapshot and publication. Supervisor suite: 52 passed, 1 ignored.

src/native_app/source_processing/supervisor.rs
PORTALSURFER: Fixed in 311ab0b4c. SourceFilesystemSyncFinished and SourceManifestAuditCommitted now carry the executing lifecycle generation end to end; handlers reject any completion whose generation is not current before status, queue, or wake/cancel side effects. Added source_completions_from_previous_readded_epoch_are_ignored to prove old completions cannot cancel the re-added generation. Focused test, cargo check --all-targets, and scripts/ci.sh smoke passed.

src/native_app/source_processing/supervisor.rs
```diff
@@ -162,6 +195,7 @@ impl SourceProcessingBudgetHandle {
                 );
             }
             let admission_cancel = Arc::clone(&control.source_work_cancels[source_id]);
+            let lifecycle_generation = control.source_lifecycle_generations[source_id];
```
PORTALSURFER: Required fix (High): Bind admission to the generation/descriptor captured with the external request. `acquire_scan` accepts only `source_id` and reads whatever lifecycle generation is current here, after its initial wait. Both folder scans and filesystem syncs have already captured `root`/`database_root` before the background closure calls this method. If that closure waits for pause, retirement, or scheduling and `replace_sources` installs a same-ID descriptor with a new generation, the stale request is admitted under the replacement generation; `SourceFilesystemSyncFinished` is then stamped with that new generation and passes the new completion fence even though it operated on the old storage. Capture the expected generation (or an exact descriptor identity) when the request is registered/queued, require it in `acquire_scan`, wait only while that same generation remains configured, and reject rather than adopting a replacement generation. Add a regression where an old request is held, the same ID is replaced with different storage, and the old request never receives or publishes as the new-generation permit.

src/native_app/source_processing/supervisor.rs
```diff
@@ -1468,6 +1631,102 @@ fn run_coordinator(shared: Arc<Shared>) {
...
 }
 
+fn process_ready_source_retirements(shared: &Shared) {
+    let now = now_epoch_seconds();
+    let candidates = {
+        let control = shared.control();
+        control
+            .pending_retirements
+            .iter()
+            .filter(|(_, retirement)| retirement.retry_at <= now)
```
PORTALSURFER: Required fix (Medium): A same-storage re-add must supersede retirement immediately even when cleanup is in backoff. Candidate selection currently excludes every retirement whose `retry_at` is in the future. After one cleanup failure sets the five-second retry deadline, re-adding that source wakes the coordinator but this filter still skips it; `source_is_active` remains false and the new `acquire_scan` waiter blocks despite there being no old external or in-flight work. That contradicts the DoD that re-add waits only until the old epoch drains. Let an active matching storage identity bypass the cleanup retry deadline (while retaining the exact activity checks), or reset the matching retirement deadline on re-add, and add a regression with a future retry deadline proving the handoff and admission resume immediately.

PORTALSURFER (commented):
Fresh stateless unsafe/concurrency specialist review of the current PR description against the complete merge-base diff. The exact drain accounting, current-generation discovery projection, and source-scoped filesystem/audit completion checks are coherent, but two remaining generation-handoff mismatches require fixes. GitHub cannot record REQUEST_CHANGES because the authenticated reviewer is also the PR author, so the blocking findings are submitted as Required fix inline comments.

src/native_app/source_processing/supervisor.rs
PORTALSURFER: Fixed in 7ac54707b. External folder scans and targeted filesystem syncs now capture lifecycle generation when their source descriptor/root request is queued and call acquire_scan_for_generation with that expected epoch. Admission waits only while that exact generation remains configured and rejects descriptor replacement instead of adopting the new generation. Added external_admission_rejects_generation_captured_before_descriptor_replacement. Supervisor suite: 53 passed, 1 ignored.

src/native_app/source_processing/supervisor.rs
PORTALSURFER: Fixed in 7ac54707b. Retirement candidate selection now lets an actively re-added matching storage identity bypass cleanup retry_at, while exact external/in-flight drain checks still gate handoff. The fast re-add regression now assigns a future retry deadline, proves admission remains pending only while old work is held, and proves it resumes immediately after the old epoch drains. cargo check --all-targets and scripts/ci.sh smoke passed.

src/native_app/sample_library/folder_scan_actions/worker.rs
```diff
@@ -67,11 +67,14 @@ impl NativeAppState {
         );
         let budget = self.background.source_processing.budget_handle();
         let source_id = request.source_id.clone();
+        let expected_lifecycle_generation = budget.lifecycle_generation(&source_id);
```
PORTALSURFER: Required fix (High): Do not recover an existing generation after descriptor registration failed. `register_source_for_scan` rejects a same-ID mismatched descriptor, but `launch_folder_scan` only logs that error and then unconditionally reads the supervisor generation here. The worker can therefore acquire the authoritative source generation while executing the rejected request’s captured `root`/`database_root`, defeating the exact request-generation binding and allowing old or different storage to run under the current epoch. Make successful descriptor registration atomically return the generation to use (or otherwise couple exact descriptor validation with generation capture), keep the expected generation absent on any registration error, and cancel the queued worker without touching storage. Add a production-path regression with a same-ID mismatched request proving registration failure cannot obtain a permit.

src/native_app/shell/message_dispatch/browser.rs
```diff
@@ -45,6 +45,26 @@ impl NativeAppState {
...
+                            .is_empty()
+                } else {
+                    self.library
+                        .folder_browser
+                        .source_exists(&progress.source_id)
+                        && self
+                            .background
+                            .source_lifecycle_generations
+                            .get(&progress.source_id)
+                            .is_none_or(|generation| *generation == progress.lifecycle_generation)
```
PORTALSURFER: Required fix (Medium): Missing generation state must not mean “current.” Removal synchronizes `source_lifecycle_generations` to remove the source, but initial/re-add `launch_folder_scan` only registers the source inside the supervisor and does not update this UI map until the scan-finish path later calls `sync_source_watcher`. During that interval an old epoch can still be draining and publish progress; because this check uses `is_none_or`, the re-added browser entry accepts that old generation. Current manifest-audit completions can conversely be rejected by their stricter `Some` comparison in the same gap. Synchronize the returned registration generation into the UI owner before spawning the scan and require an exact `Some(current_generation)` match here. Add a fast remove/re-add regression where old progress arrives after registration but before scan completion and is ignored while current-generation delivery is accepted.

PORTALSURFER (commented):
Fresh stateless unsafe/concurrency specialist review of the current PR description against the complete merge-base diff. Exact-generation waiting, drain/finalization accounting, retirement reactivation over retry backoff, discovery fencing, and filesystem/audit completion checks are coherent, but two production handoff gaps still require fixes. GitHub cannot record REQUEST_CHANGES because the authenticated reviewer is also the PR author, so the blocking findings are submitted as Required fix inline comments.

src/native_app/sample_library/folder_scan_actions/worker.rs
PORTALSURFER: Fixed in 19ca91fec. register_source_for_scan now atomically validates the exact descriptor and returns its lifecycle generation. launch_folder_scan uses only that Result: registration failure leaves the worker expected generation absent, so the cancelled path cannot acquire a permit or touch the rejected storage. Added mismatched_folder_scan_registration_cannot_adopt_existing_source_generation. The registration-path test group passed 3/3.

src/native_app/shell/message_dispatch/browser.rs
PORTALSURFER: Fixed in 19ca91fec. Successful scan registration writes the returned supervisor generation into the UI-owned source_lifecycle_generations map before spawning background work, and SourceProcessingProgress now requires an exact Some(generation) match. The first-scan handoff regression asserts UI and supervisor generations are synchronized before admission; stale/current epoch routing tests and the 9-test message-dispatch group pass.

src/native_app/sample_library/folder_scan_actions/filesystem_refresh.rs
```diff
@@ -236,25 +251,31 @@ impl NativeAppState {
             return;
         };
         let budget = self.background.source_processing.budget_handle();
+        let expected_lifecycle_generation = budget.lifecycle_generation(&source_id);
```
PORTALSURFER: **Required fix (High): bind filesystem-sync admission to the exact source descriptor.** `root` and `database_root` above come from the folder-browser descriptor, while this lookup obtains a generation independently by source ID. Those states can diverge today: changing a source role updates `metadata_storage` and `database_root`, but the role mutation path only persists settings and does not call `sync_source_watcher`, so the supervisor still owns the prior descriptor and generation. A subsequent watcher event then acquires that authoritative generation here and writes through the new database roots; its completion is also accepted because the UI generation map still contains that same generation. This means a request whose descriptor does not match the supervisor can still adopt the authoritative epoch, defeating the PR’s external-scan fencing and risking writes/publication against the wrong source database. Admit the filesystem sync with the complete `SampleSource` descriptor atomically (or otherwise verify the captured roots/storage against the supervisor descriptor before returning the generation), synchronize role/storage replacements into the supervisor, and add a regression test showing a mismatched descriptor cannot acquire or publish under the existing generation.

PORTALSURFER (commented):
Unsafe/concurrency specialist review: one required fix remains in external admission binding. The folder-scan path now rejects a mismatched descriptor, but targeted filesystem sync can still pair roots from one descriptor with the authoritative generation of another.

src/native_app/sample_library/folder_scan_actions/filesystem_refresh.rs
PORTALSURFER: Fixed in 45a9b0e8a. Targeted filesystem sync now reconstructs the complete configured SampleSource descriptor, derives both storage roots from that descriptor, and admits it atomically through register_source_for_scan before capturing the exact lifecycle generation. Source role/storage mutations now replace the supervisor and watcher descriptors, advancing the generation and cancelling older work. Added targeted_sync_cannot_pair_updated_storage_with_previous_source_generation to prove a browser/supervisor descriptor mismatch is rejected without changing either published generation. Exact-head validation: cargo check --all-targets; supervisor tests 53 passed, 1 ignored; message dispatch tests 9 passed; scan handoff tests 4 passed; protected extraction target flow passed; scripts/ci.sh smoke passed; scripts/check.sh non-blocking-architecture passed; git diff --check passed.

PORTALSURFER (commented):
Fresh stateless exact-head review completed at 45a9b0e8ac1c2ecf70dbada2cec600001c87b2d8 against merge base 88088143a6b45c4e379f31cfcf7846fd6b54b53e.

Verdict: clean. The current complete diff achieves the PR Goal, Scope, and Definition of Done. Descriptor-aware targeted sync now admits the exact configured descriptor, source role/storage transitions advance the supervisor generation, and mismatched descriptors cannot adopt an existing generation. Required fixes: none. Non-blocking observations: none.

PR #832, opened 2026-07-18
OPT-1184: Prove source-processing liveness under churn

PORTALSURFER:
## Goal

Prove that source processing cannot silently idle with actionable work across churn, restart, playback, watcher recovery, and source lifecycle changes, while keeping source progress understandable in the real app.

## Scope and non-goals

- Add a deterministic native-runtime harness over real temporary sources and committed readiness generations.
- Add a liveness oracle with structured diagnostics for manifest/readiness generations, durable jobs, active work, retry/prerequisite waits, watcher stimulus, and resource state.
- Cover external create/nested create/same-size mutation/rename/delete/storm/overflow, committed internal mutation, playback, restart recovery, root disappearance/reappearance, and remove/re-add.
- Process one source at a time and expose the active source with an always-visible pulsing row highlight without disabling interaction.
- Treat actively growing recordings as unstable and defer them until their metadata is quiet.
- Recover from native watcher construction/failure stalls with bounded asynchronous startup and polling fallback.
- Remove playback cache generation from durable source readiness; audition WAV files through the playback runtime and retain only compact waveform cache data under a bounded budget.
- Prevent startup browser scans from being cancelled by source-replacement fencing; admission waits only on the background worker.
- Distinguish queued folder scans from admitted scan work so a waiting source never appears concurrently active.
- Suppress brief maintenance reconciliation from the source pulse, park capacity-blocked discovery instead of spinning, and avoid treating initial native watcher registration as runtime overflow.
- Fence watcher ingress until every root is registered, then re-arm authoritative manifest audits without dropping real post-ready events.
- Preserve valid in-flight processing when delayed UI projection or completed foreground-scan notifications arrive; only true filesystem invalidations cancel a generation.
- Add an explicit 10,000-file calibrated performance lane. Non-goal: run that profile in default CI.

## Definition of Done

- Active actionable deficits are always observable as dirty/scheduled, queued, in flight, retry-waiting, prerequisite-waiting, or explicit terminal/offline state.
- Only one source owns processing capacity at a time; its source row pulses without hover and remains usable.
- A source displays `(scanning)` only after its scan worker emits progress proving admission to the single lane; queued scans remain visually idle.
- Brief startup maintenance does not flash source rows or synthesize full foreground scans for every configured source.
- Source add, restart, watcher churn/failure recovery, playback, internal mutations, root loss/recovery, and remove/re-add converge without manual Process/Refresh.
- Exact committed manifest, analysis, embedding/aspect, similarity, and obsolete-membership state is asserted directly.
- Actively recorded files are ignored until stable instead of holding the queue open.
- WAV audition remains effectively immediate without decoded per-file playback artifacts or a 64 GiB playback cache.
- Startup source scans wait for the authoritative lifecycle generation rather than silently cancelling and repeating next launch.
- Delayed non-mutating source notifications cannot cancel and restart a current long-running finalizer.
- The large-source profile reports and enforces throughput, browser priority latency, CPU, memory, disk I/O, queue completion, and contention budgets.

## Validation

- `bash scripts/ci.sh agent` — passed at exact head `e8f9f5eca`: Radiant 1650 passed; Wavecrate 1856 passed, 20 ignored, 2 filtered; non-blocking architecture guard, UI-thread blocking guard, strict slow-handler diagnostics, and real-output playback stress passed.
- Source-processing supervisor, readiness, cache, watcher, and fast-audition focused suites — passed.
- Watcher suite — 19 passed, including idempotent startup sync, post-ready event delivery, and no synthetic startup overflow refresh.
- Stale scan-admission regressions — a scan waiting behind the replacement fence cannot resurrect a removed source and resolves only the authoritative lifecycle descriptor.
- Capacity-parking regression — a source blocked by an external scan completed at most two discovery sweeps instead of tight-looping.
- Brief-versus-sustained discovery progress regressions and explicit maintenance-versus-active source-row projection regressions — passed.
- Non-mutating source request regression — passed; the current cancel token and lifecycle generation remain intact while reconciliation is re-armed.
- Watcher-ready overlap regression — passed; a watcher-ready audit request arriving during an older in-flight audit survives that completion.
- Maintenance pulse-overlay regression — passed; manifest maintenance does not paint the transient source-row pulse.
- Sustained manifest-audit regression — passed; brief maintenance stays quiet, then a grace-surviving audit activates its source row.
- Active-recording quiet-window regressions — passed for both unhashed and already-hashed recently modified WAVs; downstream work resumes only after metadata is quiet.
- Manifest-audit browser refresh handoff regressions — passed; file-list changes transfer ownership to the foreground scan, a delivered handoff survives post-commit cancellation, mutations cannot restart coordinator discovery while that handoff is pending, scan-permit release alone does not re-arm work, only `SourceScanFinished` resumes normal reconciliation, missing roots release the handoff into explicit offline discovery, and content-generation-only changes still reconcile immediately.
- Retirement isolation, replacement, and shutdown regressions — passed; cleanup runs in a killable child process behind a dedicated joined worker, each pending retirement has its own cancellation token, source add/remove/re-add cancels blocked cleanup without waiting on stalled storage while the replacement fence is held only for admission and publication, active sources still converge, and blocked cleanup is cancelled and joined promptly during shutdown.
- Pre-I/O recovery feedback regression — passed; the selected source publishes `Opening source database` with its exact root before storage I/O begins.
- Durable target contract — playback is file-backed and compact waveform caches are explicitly best-effort outside source readiness.
- Deterministic ignored restart/churn/root-recovery liveness oracle — passed.
- 10,000-file profile — passed at 353 candidates/s with browser-priority p99 34 us, bounded CPU/memory/disk, and zero DB contention.
- `bash scripts/build.sh --release --variant OPT-1184` — signed exact-head app bundle built. The bundle launched, but final visual inspection was unavailable because the macOS session was locked.
- Live launch smoke on the preceding fix head — source recovery was immediately source-scoped and observable as `Opening source database`; removed-source retirement advanced independently on its dedicated joined worker. The log identified the current wait as the `Projects` source database itself, and an out-of-process one-byte read likewise transferred zero bytes before interruption.

## Known limitations

- The current local `Projects` `.wavecrate.db` is not returning file data to Wavecrate or a direct OS read, so live convergence of that user source cannot be completed until the storage layer becomes readable. The supervisor now identifies and pulses that exact source/phase instead of appearing idle; deterministic temporary-source convergence and the full runtime gate pass.

Linear: OPT-1184


src/test_support/source_processing_liveness/diagnostics.rs
```diff
@@ -0,0 +1,251 @@
...
+                matches!(
+                    entry.classification,
+                    ReadinessClassification::RetryableFailure { retry_at, .. } if retry_at > now
+                )
+            }));
+    let waiting_for_prerequisite = !has_schedulable_deficit
+        && snapshot.entries.iter().any(|entry| {
+            entry.target.stage == ReadinessStage::SimilarityLayout
+                && entry.classification != ReadinessClassification::Current
+                && !snapshot.prerequisites_are_current(&entry.target)
```
PORTALSURFER, line 75: Required fix — `queue_depth` and `readiness_queue_depth` come from supervisor-wide telemetry, but this oracle is evaluating one `source_id`. Work queued for source B therefore makes `observable_work` true while source A has actionable deficits and no dirty/queued/in-flight/prerequisite/retry state, so the claimed per-source silent-idle proof can pass in exactly the multi-source case the Goal covers. Track queued/readiness work per source (or derive it from source-keyed candidates) and add a two-source regression where unrelated queued work cannot mask an idle deficient source.

src/test_support/source_processing_liveness/harness.rs
```diff
@@ -0,0 +1,167 @@
...
+        let supervisor = SourceProcessingSupervisor::start(vec![source.clone()]);
+        Self {
+            source_parent,
+            source,
+            supervisor,
+            watcher_stimulus: WatcherStimulus::Startup,
+            watcher_events_committed: 0,
+            expected_source_generation: 1,
+        }
+    }
```
PORTALSURFER, line 39: Required fix — This harness calls `sync_source_database_paths` directly and then calls `wake_source`; the overflow path similarly calls `audit_source_and_record` directly. That bypasses watcher event delivery, watcher loss/overflow handling, and watcher restart/re-subscription entirely, even though the PR Goal/DoD require watcher churn and failure recovery (and the description claims a live watcher demonstration). Drive the real watcher-to-manifest path and force its overflow/loss/recovery behavior, then assert the same committed-generation convergence without manually reconciling or waking the supervisor.

src/test_support/source_processing_liveness/scenarios.rs
```diff
@@ -0,0 +1,291 @@
...
+    let materialization_throughput =
+        candidates.len() as f64 / discovery_elapsed.as_secs_f64().max(f64::EPSILON);
+    assert!(discovery_elapsed <= DISCOVERY_BUDGET);
+    assert!(materialization_throughput >= MIN_MATERIALIZATION_THROUGHPUT);
+    assert_eq!(candidates.len(), FILE_COUNT * 4 + 1);
+    assert_eq!(stats.readiness_queue_depth, FILE_COUNT * 4);
+    assert_eq!(stats.prerequisites_blocked, FILE_COUNT * 3);
+
+    let resources_after = process_resource_snapshot();
+    let memory_growth = resources_after
```
PORTALSURFER, line 365: Required fix — The 10,000-file lane keeps playback active, invokes `discover_source_candidates` directly, and then asserts `claimed == 0`. It only measures candidate materialization; no candidate is executed and no readiness queue ever completes. This misses the Definition of Done requirement that the large-source profile enforce queue completion. Add a resumed-processing phase that waits for the real supervisor to drain and asserts the exact 40,001-target completion/convergence (while retaining the paused browser-priority measurement).

PORTALSURFER (commented):
Required fixes identified: the current diff does not yet provide the stated per-source liveness proof, watcher-recovery coverage, or 10,000-file queue-completion measurement required by the Definition of Done. I cannot request changes on this PR because the authenticated GitHub identity is its author.

src/test_support/source_processing_liveness/diagnostics.rs
PORTALSURFER: Fixed in 0029c8758. The oracle now derives observable queue state from source-keyed telemetry and includes a two-source regression proving unrelated queued work cannot mask an idle deficient source. The deterministic liveness lane and full validation suite pass.

src/test_support/source_processing_liveness/harness.rs
PORTALSURFER: Fixed in 0029c8758. The harness now drives real watcher events, overflow/loss recovery, asynchronous watcher restart, and committed manifest convergence without manually reconciling or waking the supervisor. Native watcher and ignored restart/churn/root-recovery lanes pass.

src/test_support/source_processing_liveness/scenarios.rs
PORTALSURFER: Fixed in 0029c8758. The 10,000-file profile now resumes the real supervisor, drains the queue, and asserts exact completion of the maintained readiness target set while retaining the paused browser-priority measurements. The calibrated profile passed with zero DB contention.

PORTALSURFER (commented):
Required fixes are needed before this diff meets the PR Definition of Done. Inline threads cover stale lifecycle re-admission, dropped watcher mutations during startup suppression, and missing source activity feedback after the discovery grace period. I could not submit a formal request-changes review because this GitHub identity authored the PR.

src/native_app/sample_library/folder_scan_actions/worker.rs
```diff
@@ -80,14 +86,34 @@ impl NativeAppState {
...
-                let Some(permit) = expected_lifecycle_generation.and_then(|generation| {
-                    budget.acquire_scan_for_generation(&source_id, generation)
-                }) else {
+                let Some(source) = source else {
+                    let cancel = std::sync::Arc::new(std::sync::atomic::AtomicBool::new(true));
+                    return run_folder_scan_worker(request, events, cancel);
+                };
+                let generation = match match admission_generation {
+                    Some(generation) => Ok(generation),
+                    None => budget.register_source_for_scan_waiting(source),
```
PORTALSURFER, line 95: Required fix — When UI registration loses the replacement fence, this waits and then calls `register_source_for_scan_waiting(source)`. That helper adds the captured descriptor when it is absent. If the concurrent replacement removed this source, the stale scan now re-registers it after removal and can run against a source the browser has already retired, violating the required remove/re-add lifecycle fencing. The waiting path must only return the currently authoritative matching descriptor and generation after the fence releases; reject an absent or changed source instead of adding it. Add a regression that removes the source while this admission is waiting.

src/native_app/sample_library/source_watcher/handle.rs
```diff
@@ -198,7 +423,12 @@ fn run_source_watcher(
         }
 
         let mut watcher_failed = false;
+        let suppress_initial_ingress =
+            initial_ingress_quiet_until.is_some_and(|deadline| now < deadline);
         while let Ok(event) = event_rx.try_recv() {
+            if suppress_initial_ingress {
```
PORTALSURFER: Required fix — The initial quiet window drops every queued notify event, not just synthetic registration callbacks. A real create or mutation that lands after the startup manifest audit but during these 400 ms is consumed here, so no targeted refresh is published and the source can remain apparently settled until some unrelated audit. Preserve real post-registration events (or immediately run an authoritative audit after the suppression fence) while suppressing only the registration noise, and cover that interleaving.

src/native_app/app_chrome/view_models/library_sidebar.rs
```diff
@@ -178,9 +180,27 @@ impl LibrarySidebarViewModel {
...
                 state.ui.chrome.help_tooltips_enabled,
+                state
+                    .background
+                    .source_processing_progress
+                    .as_ref()
+                    .filter(|progress| {
+                        progress.active
+                            && !matches!(
+                                progress.stage.as_str(),
+                                "Checking pending work" | "Scanning source changes"
```
PORTALSURFER: Required fix — `DiscoveryProgressPublisher` emits `Checking pending work` only after its 250 ms UI grace threshold, but this filter suppresses the processing row for every such message. Consequently a genuinely long discovery owns source-processing capacity and shows active status-bar work without the required pulsing source row. Keep the brief-maintenance suppression separate from grace-surviving discovery progress, then add a view-model regression for an active `Checking pending work` update.

src/native_app/sample_library/folder_scan_actions/worker.rs
PORTALSURFER: Fixed at af9fcd990. Deferred scan registration now resolves only the authoritative configured descriptor and lifecycle generation after the replacement fence releases; it rejects absent, changed, inactive, or shutting-down sources. Added `background_scan_registration_cannot_readd_source_removed_behind_fence` plus updated waiting-path coverage. Full `bash scripts/ci.sh agent` passes at this head.

src/native_app/sample_library/source_watcher/handle.rs
PORTALSURFER: Fixed at af9fcd990. Watcher callback ingress stays fenced only while all roots are registered, opens once the watcher is installed, and emits a one-time ready boundary that re-arms authoritative manifest audits. Real post-ready events are no longer discarded. Added `filesystem_event_after_initial_watcher_ready_is_not_suppressed` and watcher-ready audit coverage; all 19 watcher tests and the full agent gate pass.

src/native_app/app_chrome/view_models/library_sidebar.rs
PORTALSURFER: Fixed at af9fcd990. Progress now carries an explicit `source_row_active` signal: grace-surviving discovery/readiness work pulses, while manifest maintenance does not. Added source-row projection and discovery-publisher regressions. The signed exact-head app showed no false maintenance pulse while all sources converged.

PORTALSURFER (commented):
Reviewed exact head `af9fcd990c0afeb57ac74b0c00a59b8f8507c60a` against the current PR description and complete merge-base diff. The diff does not yet achieve the watcher-gap recovery and maintenance-pulse requirements because of the two inline required fixes below. GitHub does not allow this identity to request changes on its own pull request, so this review is submitted as a comment review; both inline comments remain required fixes.

src/native_app/source_processing/supervisor.rs
```diff
@@ -1471,15 +1604,65 @@ fn run_coordinator(shared: Arc<Shared>) {
...
             drop(in_flight_work);
             shared.budgets().release(permit);
             shared.budget_wake.notify_all();
+            if matches!(&result, Ok(ExecutionOutcome::Completed))
+                && matches!(&candidate.task, RuntimeTask::ManifestAudit)
+            {
+                shared
+                    .control()
+                    .force_manifest_audit_sources
+                    .remove(candidate.source.id.as_str());
```
PORTALSURFER: Required fix — Do not let an older in-flight manifest audit clear a newer watcher-ready audit request. `request_manifest_audits` sets this same per-source flag while an audit can already be running, but this unconditional removal after that older candidate completes erases the newer request. If a filesystem change occurs after the audit snapshot but before watcher ingress becomes ready, the ready message marks the source dirty, this line clears the force flag, and rediscovery sees the just-updated audit timestamp and schedules no closing audit, so the change can remain silently absent until another event or the 24-hour sweep. Track an audit request generation/token (captured by the candidate) and clear only the request that candidate satisfied, or otherwise preserve a re-arm that arrived after the candidate began; add the overlap regression covering watcher-ready during an in-flight audit.

src/native_app/app_chrome/status_bar.rs
```diff
@@ -106,6 +108,35 @@ impl NativeAppState {
...
+    pub(in crate::native_app) fn paint_source_processing_source_pulse(
+        &mut self,
+        context: TransientOverlayContext<'_>,
+        primitives: &mut Vec<PaintPrimitive>,
+    ) {
+        let Some(progress) = self
+            .background
+            .source_processing_progress
+            .as_ref()
+            .filter(|progress| progress.active && !progress.source_id.is_empty())
```
PORTALSURFER: Required fix — Gate the transient row pulse on `progress.source_row_active`. Manifest-audit progress is deliberately published with `source_row_active: false` so maintenance stays visible in job details without pulsing a source row, and the retained row projection honors that flag, but this overlay filters only `active` and a non-empty source id. Any manifest audit that emits progress therefore still paints an animated row pulse, contradicting the stated maintenance-suppression behavior. Use the same activity predicate here and cover the overlay path with a maintenance-progress regression.

src/native_app/source_processing/supervisor.rs
PORTALSURFER: Fixed in `8fdf37f945fad305dcf7fb7eb2800d692418c265`. Audit completion now clears the forced manifest-audit flag only when no newer dirty request is pending, so a watcher-ready request that overlaps an older in-flight audit is preserved. Added `watcher_ready_request_survives_older_in_flight_audit_completion`; the focused regression and full `bash scripts/ci.sh agent` gate pass.

src/native_app/app_chrome/status_bar.rs
PORTALSURFER: Fixed in `8fdf37f945fad305dcf7fb7eb2800d692418c265`. The transient pulse overlay now requires `source_row_active`, matching the normal row rendering path, so manifest maintenance cannot pulse a source. Added the non-vacuous `manifest_maintenance_does_not_paint_source_row_pulse_overlay` regression; the focused regression and full `bash scripts/ci.sh agent` gate pass.

src/native_app/source_processing/supervisor.rs
```diff
@@ -1829,6 +2115,7 @@ fn publish_source_processing_progress(
             source_id: candidate.source.id.as_str().to_string(),
             lifecycle_generation,
             active: true,
+            source_row_active: !matches!(candidate.task, RuntimeTask::ManifestAudit),
```
PORTALSURFER, line 2323: Required fix: This suppresses the source-row pulse for every manifest audit, not just brief maintenance. A full audit can scan for seconds or minutes, and the audit progress callback also publishes `source_row_active: false`, so the only active source has no row-level indicator throughout sustained work. That contradicts the Scope/DoD promise that only brief startup maintenance is suppressed while the active source pulses. Keep the grace-period suppression, but promote a manifest audit to `source_row_active` once it survives the brief-maintenance threshold, with a sustained-audit regression.

docs/TARGET.md
```diff
@@ -338,6 +338,21 @@ One revision-aware reconciler owns source completeness. It compares the complete
 
 Detection and execution stay off UI, render, and latency-sensitive read paths. UI and read paths may observe a side-effect-free readiness snapshot and request priority, repair, or retry, but they must not discover a missing artifact by directly enqueueing hidden work. Lifecycle writers publish committed desired generations and wake the background coordinator only after their authoritative transaction succeeds.
 
+The readiness contract must be protected by a deterministic native-runtime
```
PORTALSURFER, line 338: Required fix: The new liveness contract is being added beside a still-current readiness definition that requires a playback-summary target/artifact/cache reference and says cache-budget pruning invalidates readiness (lines 319 and 325-331). This PR instead removes playback from the target matrix and explicitly asserts zero playback readiness rows. Update the durable target contract to make compact waveform/cache data best-effort and outside source readiness (and renumber the stage list), otherwise the implementation and the repository’s authoritative product contract disagree.

PORTALSURFER (commented):
Reviewed exact head `8fdf37f945fad305dcf7fb7eb2800d692418c265` against the current PR description and complete merge-base diff. Two required fixes remain: sustained manifest audits lose the promised active-source row pulse, and the durable target document still defines playback-cache residency as readiness despite this PR removing it. No checks are currently reported for the PR, and there were no unresolved non-outdated review threads before this review.

src/native_app/source_processing/supervisor.rs
PORTALSURFER: Fixed in `09a901819a589e2bc8999edc3fe76027728a2a4f`. Manifest-audit progress now keeps the row quiet during the 250 ms maintenance grace period and sets `source_row_active` once the audit survives that threshold. Added `sustained_manifest_audit_activates_source_row_after_grace_period`; the brief-audit and overlay regressions also pass, as does the full agent gate.

docs/TARGET.md
PORTALSURFER: Fixed in `09a901819a589e2bc8999edc3fe76027728a2a4f`. `docs/TARGET.md` now defines the four durable readiness stages without playback, makes compact waveform/browser caches best-effort outside readiness, and documents file-backed WAV audition with no persistent decoded library cache. Related cache, scan, retirement, and instant-playback sections were aligned with that contract.

PORTALSURFER (commented):
Reviewed exact head 09a901819a589e2bc8999edc3fe76027728a2a4f against the current PR description and complete merge-base diff. One required fix remains: the recording quiet-window deferral stops applying as soon as a content hash exists, even though the scan/audit path can hash a small file before it has been metadata-quiet for five seconds. This leaves the active-recording Definition of Done unmet. No current checks are reported; the PR has no unresolved current review threads apart from this review.

src/native_app/source_processing/supervisor.rs
```diff
@@ -2369,6 +2723,68 @@ fn discover_source_candidates_with_connection(
...
+        .saturating_sub(ACTIVE_RECORDING_QUIET_SECONDS)
+        .saturating_mul(NANOS_PER_SECOND);
+    let mut statement = connection
+        .prepare(
+            "SELECT file_identity, modified_ns
+             FROM wav_files
+             WHERE missing = 0
+               AND file_identity IS NOT NULL
+               AND TRIM(file_identity) != ''
+               AND (content_hash IS NULL OR TRIM(content_hash) = '')
```
PORTALSURFER: Required fix: This quiet-window query excludes every recently modified file once it has a non-empty `content_hash`. The manifest audit/quick-scan path can hash a small WAV after only a momentarily stable read, so a still-open recording can acquire a hash before five seconds of metadata quiet and immediately enqueue analysis or embedding. That violates the stated scope and Definition of Done and can reintroduce cancellation churn. Apply the recent-mtime deferral regardless of hash state (or otherwise fence hash publication and downstream scheduling until the quiet deadline), and add a regression with a recently modified, already-hashed small WAV proving that no downstream candidate runs before the quiet deadline.

PORTALSURFER: Fixed in `cce74f59fada4eb95153bffeb0b9237ebedb662f`. Recent-mtime quiet-window deferral no longer depends on `content_hash`, so a quick/audit hash cannot admit a still-open WAV to analysis or embedding. Added `recently_modified_hashed_wav_is_parked_until_its_quiet_deadline`; both hashed and unhashed recording regressions pass, as do the full agent gate, signed build, and six-source live startup smoke.

src/native_app/source_processing/supervisor.rs
```diff
@@ -2369,6 +2723,67 @@ fn discover_source_candidates_with_connection(
...
+        .saturating_sub(ACTIVE_RECORDING_QUIET_SECONDS)
+        .saturating_mul(NANOS_PER_SECOND);
+    let mut statement = connection
+        .prepare(
+            "SELECT file_identity, modified_ns
+             FROM wav_files
+             WHERE missing = 0
+               AND file_identity IS NOT NULL
+               AND TRIM(file_identity) != ''
+               AND modified_ns BETWEEN ?1 AND ?2",
```
PORTALSURFER, line 2974: Required fix: `now` comes from `now_epoch_seconds()`, so `now_ns` is exactly the start of the current second, while scanner `modified_ns` preserves the filesystem timestamp’s sub-second component. A file modified normally during that same second (for example, `now * 1e9 + 500_000_000`) is therefore excluded by this `BETWEEN ... AND now_ns` predicate and can immediately enqueue hashing/analysis instead of observing the five-second quiet window. Remove or correctly widen the upper bound while handling genuinely future timestamps safely, and add a regression using an unhashed or already-hashed WAV whose `modified_ns` is within the current second but greater than `now_ns`; it must produce no downstream candidate until the quiet deadline.

PORTALSURFER (commented):
Reviewed exact head cce74f59fada4eb95153bffeb0b9237ebedb662f against the current PR description and complete merge-base diff. One required fix remains: the recording quiet-window query compares nanosecond file mtimes against a whole-second upper bound, so ordinary current-second writes can bypass deferral and enter readiness before five seconds of metadata quiet. This leaves the active-recording Definition of Done unmet. No current checks are reported; the PR had no unresolved current review threads before this review.

src/native_app/source_processing/supervisor.rs
PORTALSURFER: Fixed at exact head 726c14364442dc898a3f6ec3f592beb278591373. The active-recording upper bound now includes the end of the current second, and the regression uses an already-hashed WAV with modified_ns = now * 1e9 + 500_000_000 to prove it remains parked until the quiet deadline. The full agent gate passes at this head.

src/native_app/sample_library/folder_scan_actions/filesystem_refresh.rs
```diff
@@ -158,27 +158,33 @@ impl NativeAppState {
...
+                    created = committed_delta.created.len(),
+                    changed = committed_delta.changed.len(),
+                    moved = committed_delta.moved.len(),
+                    deleted = committed_delta.deleted.len(),
+                    "Refreshing browser projection after periodic source audit"
+                );
+                // The folder refresh writes the same source database that
+                // discovery reads. Its completion queues SourceScanFinished
+                // reconciliation, so wait until the refresh releases the DB.
+                self.queue_filesystem_source_refresh(source_id, Instant::now(), context);
```
PORTALSURFER, line 186: Required fix: preserve the promised refresh-before-reconciliation ordering. This branch only queues the GUI-side folder refresh, but after `execute_candidate` returns `Completed` the coordinator still classifies every completed manifest audit as `should_refresh` and immediately marks this source dirty (`supervisor.rs` lines 1852-1871), after releasing its budget. It can therefore rediscover the source before the GUI handles `SourceManifestAuditCommitted` and acquires the foreground scan permit, contending on the same DB or reconciling ahead of the refreshed projection. Transfer completion ownership to the foreground refresh for file-list/metadata deltas (and re-arm only from `SourceScanFinished`, while retaining immediate reconciliation for content-only deltas), with an end-to-end regression that proves coordinator discovery does not resume before the scan permit is released; the current enum-classification tests do not exercise this race.

src/native_app/source_processing/supervisor.rs
```diff
@@ -1673,26 +1921,117 @@ fn run_coordinator(shared: Arc<Shared>) {
...
-    };
+            .unwrap_or_else(|poison| poison.into_inner());
+        if control.shutdown {
+            return;
+        }
+        if !source_retirement_is_ready(&control, now_epoch_seconds()) {
+            continue;
+        }
+        drop(control);
+        process_ready_source_retirements(shared.as_ref());
```
PORTALSURFER, line 2021: Required fix: make an in-progress retirement interruptible before synchronously joining this worker. Once this call starts, neither `process_ready_source_retirements` nor `retire_source_derived_state` observes `shared.cancel`/`control.shutdown`; the latter opens and queries the removed source DB directly. `shutdown()` then unconditionally joins this worker, so a stalled source read—the storage failure explicitly exercised in the PR description—can hang app shutdown/drop indefinitely. Add cancellation or a bounded, joinable retirement I/O design and cover shutdown while cleanup itself is blocked. The current isolation test uses a dormant supervisor with no retirement worker, releases its mutex blocker before `shutdown`, and therefore does not prove this join path.

PORTALSURFER (commented):
Fresh stateless review of exact head `726c14364442dc898a3f6ec3f592beb278591373`: changes requested. Two blocking lifecycle mismatches remain against the PR Goal/Definition of Done: the manifest-audit browser-refresh ordering is not actually fenced from coordinator rediscovery, and an in-progress retirement storage stall cannot be interrupted before the joined shutdown. Submitted as `COMMENT` because this GitHub identity owns the PR.

src/native_app/sample_library/folder_scan_actions/filesystem_refresh.rs
PORTALSURFER: Fixed at 41eb892a7. File-list/metadata audit deltas now return an explicit foreground-refresh handoff outcome, the coordinator records that source as awaiting the foreground refresh and filters every intervening dirty/safety wake, and only the applied folder-scan completion clears the fence and re-arms discovery. Content-generation-only deltas retain immediate reconciliation. The end-to-end regression `manifest_projection_handoff_defers_discovery_until_external_scan_releases` injects another source mutation during the handoff, holds and releases the external scan permit, proves discovery remains unchanged after release, then proves the explicit scan-finished signal resumes discovery. Full `bash scripts/ci.sh agent` and the signed OPT-1184 build pass.

src/native_app/source_processing/supervisor.rs
PORTALSURFER: Fixed at 41eb892a7. Removed-source database cleanup now executes through the existing killable child-process boundary; the dedicated retirement thread owns and joins that child, while supervisor shutdown sets the shared cancellation token, terminates the child, and joins the retirement thread. Best-effort orphan-cache pruning was moved inside the same child so no removed/retained-source database I/O remains on the in-process retirement thread. The regression `shutdown_cancels_blocked_retirement_cleanup_and_joins_worker` starts the real retirement worker, blocks cleanup itself, invokes shutdown without releasing the blocker, and proves cancellation plus join completes in under one second. Full `bash scripts/ci.sh agent` and the signed OPT-1184 build pass.

PORTALSURFER (commented):
Fresh stateless review of exact head `e8f9f5eca15229e9ac2fb217d4b9fd67ebe20ecc`: clean. The current complete diff satisfies the PR Goal and Definition of Done. In particular, a delivered manifest refresh handoff survives post-commit cancellation; missing-root refreshes release the fence into explicit offline discovery; blocked retirement I/O runs outside the source-replacement fence and is superseded through per-retirement cancellation on re-add; and shutdown cancels those operations before joining the dedicated retirement worker. No blocking goal mismatch found. Submitted as `COMMENT` because this GitHub identity owns the PR.

PR #833, opened 2026-07-19
OPT-1203 Detect same-path source root replacement

PORTALSURFER:
## Goal

Detect when a configured source root is replaced or remounted at the same path so Wavecrate fences the retired watcher, registers the current filesystem object, and performs an authoritative source reconciliation before the daily audit is relevant.

## Scope and non-goals

- Track platform-stable filesystem identity for each successfully watched root.
- Compare current root identity on root events and bounded periodic root refreshes.
- Treat destructive/name root events and identity changes as targeted watcher invalidation plus overflow reconciliation.
- Back off identity-read uncertainty from 30 seconds to one hour while still failing toward periodic full reconciliation.
- Preserve low-latency child-path handling and suppression of metadata-like live-root echoes.
- Extend the liveness scenario with a same-path replacement and document the durable recovery contract.
- Do not move processing into watcher callbacks, redesign source descriptors/classification, or replace the notify backend.

## Definition of Done

- Same-path delete/recreate and atomic replacement are detected by filesystem identity.
- The affected source is overflow-reconciled and the watcher is re-registered with ingress fencing.
- Root disappearance/reappearance continues to recover.
- Metadata-like root echoes on an unchanged live root remain suppressed.
- Identity-read failure triggers bounded, backoff-limited full reconciliation rather than silent acceptance or a tight scan loop.
- Deterministic watcher regressions, compile/smoke gates, and a real desktop replacement smoke pass.

## Validation

Passed:

- `cargo test -p wavecrate --bin wavecrate source_watcher` — 26 passed.
- `cargo test -p wavecrate-library filesystem_identity` — 2 passed.
- `cargo check -p wavecrate --all-targets`.
- `bash scripts/ci.sh smoke`.
- Real desktop smoke using one freshly built app process and a temporary source: renamed the configured root aside, created a different directory at the identical path containing `new-pop.wav`, and observed Wavecrate automatically recreate source artifacts and update the live browser to `new-pop` without Refresh or Process.

Not green before the new replacement checkpoint:

- `cargo test -p wavecrate --bin wavecrate source_processing_liveness_harness_converges_restart_churn_and_root_recovery -- --ignored --nocapture` was run twice. The existing harness became silently idle during startup/its first targeted reconciliation with a pending readiness job and no queued or in-flight work, before reaching this PR's new same-path replacement scenario.
- `bash scripts/agent.sh request` completed its compile-backed architecture checks but reported existing facade-guard violations outside this diff.

## Known limitations

The full ignored liveness harness cannot currently provide end-to-end automated evidence for the new checkpoint because it reproducibly stops in pre-existing source-processing scheduler state before that checkpoint. The focused coordinator regressions and real desktop smoke directly prove the watcher replacement behavior.


PORTALSURFER (commented):
Reviewed current head `f35ee15e1b544d21df1443d3fd0d82b7f4bf06d3` against the PR goal and Definition of Done. No required fixes found: the diff detects same-path replacement by filesystem identity, fences and re-registers the affected watcher, reconciles overflowed sources, preserves live-root echo suppression, and bounds identity-read recovery.

PR #840, opened 2026-07-19
OPT-1202: Keep source processing pulse cadence stable during playback

PORTALSURFER:
## Goal

Keep the active source-processing row pulse at the same visible rate whether audio playback is idle or active.

## Scope and non-goals

- Replace the retained source row's frame-count-driven pulse with a stable base highlight.
- Keep the existing monotonic-animation-time transient overlay as the single pulse clock.
- Preserve the processing marker, row actions, and manifest-maintenance suppression.
- Add regression coverage proving frame-count tick changes do not change pulse phase.
- Non-goals: source-processing scheduling, playback/audio behavior, and unrelated progress animations.

## Definition of Done

- The source-processing pulse advances from elapsed animation time rather than UI frame count.
- Playback frame cadence cannot speed up the source-row pulse.
- The processing row remains visibly highlighted and interactive.
- Manifest maintenance still does not pulse a source row.
- Focused tests, the standard agent gate, and a freshly built signed-app smoke are green.

## Validation

- `cargo test -p wavecrate --bin wavecrate source_processing_source_pulse_uses_animation_time_not_frame_count` — passed (1 test).
- `cargo test -p wavecrate --bin wavecrate processing_source_row_keeps_actions_enabled_and_paints_activity_marker` — passed (1 test).
- `cargo test -p wavecrate --bin wavecrate manifest_maintenance_does_not_paint_source_row_pulse_overlay` — passed (1 test).
- `bash scripts/ci.sh smoke` — passed.
- `bash scripts/ci.sh agent` — passed: 1,650 Radiant tests, 50 app-runtime API tests, and 1,856 Wavecrate tests; 20 intentional ignores.
- `cargo fmt --all --check` and `git diff --check` — passed.
- `bash scripts/build.sh --release` — built and signed `target/app-bundles/opt1202/Wavecrate OPT-1202.app`.
- Signed-app runtime smoke — source processing remained visible while playback was active and after stopping playback; user confirmed the pulse now appears correct.

## Known limitations

None.

### DIFF
--- a/src/native_app/app_chrome/library_browser/library_sidebar/source_section/rows.rs
+++ b/src/native_app/app_chrome/library_browser/library_sidebar/source_section/rows.rs
@@ -25,8 +25,7 @@ const SOURCE_ROW_OUTLINE_INSET: f32 = 0.5;
 const SOURCE_ROW_OUTLINE_WIDTH: f32 = 1.0;
 const SOURCE_ROW_OUTLINE_COLOR: ui::Rgba8 = ui::Rgba8::new(255, 255, 255, 30);
 const SOURCE_PROCESSING_MARKER_COLOR: ui::Rgba8 = ui::Rgba8::new(255, 151, 72, 230);
-const SOURCE_PROCESSING_FILL_MIN_ALPHA: u8 = 48;
-const SOURCE_PROCESSING_FILL_MAX_ALPHA: u8 = 112;
+const SOURCE_PROCESSING_FILL_ALPHA: u8 = 48;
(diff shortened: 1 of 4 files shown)


PORTALSURFER (commented):
Clean review — the merge-base-to-current-head diff satisfies the stated goal and Definition of Done: the retained source row no longer derives its pulse phase from frame count, the transient overlay remains animation-time-driven, and focused regression coverage verifies frame-tick independence while preserving processing-row behavior and manifest-maintenance suppression. Reviewed head: `c31eb408d9333f151ad961e36c3588d0bef457f4`.

PR #849, opened 2026-07-20
Wavecrate: target architecture guardrail test binaries directly

PORTALSURFER:
## Goal

Make the required non-blocking architecture lane launch only the Cargo test targets that own its focused assertions.

Linear: OPT-1218

## Scope and non-goals

- Target the Radiant synthetic fixture with `--lib`.
- Target the Wavecrate app-facing assertion with `--test gui_boundary`.
- Target the Wavecrate slow-handler assertion with `--lib`.
- Preserve `--no-default-features`, failure behavior, static architecture scans, CI wiring, and Bash/PowerShell parity.
- Pin all six Bash/PowerShell selector fragments in the existing deterministic script guardrail contract.
- Do not change production architecture, allowlists, Cargo caching, or broader preflight ownership.

The issue suggested `--bin wavecrate` for the slow-handler test, but current target enumeration proves that binary runs zero matching tests. The test is owned by Wavecrate's library target, so this PR uses `--lib` to preserve the intended assertion.

## Definition of Done

- [x] Each focused command launches exactly one owning test binary.
- [x] Each intended test executes exactly once.
- [x] Deterministic script coverage requires the selectors in both Bash and PowerShell.
- [x] A synthetic blocking violation still fails the targeted lane with precise diagnostics.
- [x] Warm-build evidence shows materially fewer launched binaries and lower wall time.
- [x] The architecture and script guardrail lanes pass.
- [x] The full agent lane result is calibrated against detached `main`.

## Validation

Exact commit: `92bdd66bf8bfb628be8dc14b38d68d9d963892a5`

- `cargo test --manifest-path vendor/radiant/Cargo.toml --lib guardrail_reports_file_line_and_guidance_for_blocking_tokens` — one binary, one test passed.
- `cargo test -p wavecrate --no-default-features --test gui_boundary native_app_ui_update_paths_do_not_call_blocking_business_apis` — one binary, one test passed.
- `cargo test -p wavecrate --no-default-features --lib rapid_navigation_harness_keeps_ui_responsive_while_business_work_is_slow` — one binary, one test passed.
- Temporary synthetic `std::thread::sleep` in a scanned native-app file — targeted GUI command exited 101 and reported the exact file, line, token, and offloading guidance; fixture removed afterward.
- `bash scripts/internal/check/check_non_blocking_architecture.sh` — passed; four total intended test binaries launched.
- `bash scripts/internal/check/check_script_guardrails.sh` — passed, including all six new selector assertions.
- `bash scripts/ci.sh agent` — optimized architecture lane, 1,650 Radiant library tests, 50 Radiant runtime API tests, and five legacy-controller integration tests passed. The Wavecrate library phase then reproduced the current baseline result: 3,692 passed, 51 failed, 23 ignored.
- Detached `main` ran the exact same Wavecrate library command and reproduced exactly 3,692 passed, 51 failed, and 23 ignored, confirming those failures are independent of this script-only diff.

Warm architecture-lane comparison on the same build artifacts:

| | Before | After | Change |
| --- | ---: | ---: | ---: |
| Launched test binaries | 41 | 4 | -90.2% |
| Wall time | 1.45 s | 1.17 s | -19.3% |

## Known limitations

- PowerShell was not installed on this macOS host, so PowerShell parity is enforced by the same deterministic required-fragment contract rather than a local PowerShell execution.
- The unrelated 51-test Wavecrate library baseline remains outside this PR.

### DIFF
--- a/scripts/internal/check/check_non_blocking_architecture.ps1
+++ b/scripts/internal/check/check_non_blocking_architecture.ps1
@@ -47,7 +47,7 @@ try {
   Enable-WavecrateCargoCache
 
   Invoke-NativeStep -Label "Radiant synthetic blocking-token fixture" -Command {
-    Invoke-WavecrateCargo test --manifest-path vendor/radiant/Cargo.toml guardrail_reports_file_line_and_guidance_for_blocking_tokens
+    Invoke-WavecrateCargo test --manifest-path vendor/radiant/Cargo.toml --lib guardrail_reports_file_line_and_guidance_for_blocking_tokens
   }
 
   Invoke-NativeStep -Label "Radiant app/runtime/example guardrails" -Command {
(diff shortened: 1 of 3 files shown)


PORTALSURFER (commented):
Reviewed head `92bdd66bf8bfb628be8dc14b38d68d9d963892a5` against the PR description and complete merge-base diff. **clean** — the selector changes target each owning Cargo binary exactly once, preserve the existing guardrails and Bash/PowerShell parity, and pin the corrected library-owned slow-handler test.

PR #853, opened 2026-07-20
Show session-local played waveform ranges

PORTALSURFER:
## Goal

Show which parts of the loaded sample have been auditioned during the current load session, with playback and the visual rail advancing together from the actual audible position.

## Scope and non-goals

- Add a subtle neutral-grey 4 px rail along the waveform bottom for merged played ranges.
- Update the rail on paint-only playback frames, including navigation playback.
- Keep history local to the loaded waveform and reset it when another sample is loaded.
- Stream supported WAV navigation from the original file at frame zero so waveform promotion does not splice a short preview into a second full playback source.
- Use a minimal 3 ms crossfade for rapid source replacement and stop play-selection mode at terminal one-shot progress.
- Do not persist played ranges as sample metadata or change extraction markers.

## Definition of Done

- Played portions accumulate as a subtle bottom rail while audio is actually playing.
- The rail grows smoothly in realtime and starts at the true navigation playback start.
- Navigation neither replays the audition prefix nor creates a preview-to-full transition click.
- Rapid navigation avoids hard source cuts without materially softening kick transients.
- A completed one-shot play selection leaves playback stopped, so a later selection does not auto-play.
- Played history survives repaint and selection changes for the current load only, and resets for a different loaded sample.

## Validation

- `cargo fmt --all`
- `git diff --check`
- `cargo test -p reson --lib --no-run`
- `cargo test -p wavecrate --lib --no-run`
- `bash scripts/ci.sh smoke`
- `bash scripts/build.sh --release`
- Signed runtime artifact: `target/app-bundles/opt1194/Wavecrate OPT-1194.app`
- Live runtime: user confirmed continuous progress/played-rail behavior and click-free rapid navigation with the 3 ms crossfade.

The local debug test executables compile but stall before harness entry under this machine setup, so focused test execution is represented by compile coverage plus the established smoke gate and signed-app runtime verification.

## Known limitations

None.

### DIFF
--- a/src/native_app/audio/sample_load_actions/completion/loaded.rs
+++ b/src/native_app/audio/sample_load_actions/completion/loaded.rs
@@ -140,21 +140,24 @@ impl NativeAppState {
         if !self.audio.active_sample_playback_is_streamable(path) {
             return false;
         }
-        let Some(progress) = self
+        let pending_runtime = self.audio.active_sample_playback_pending_runtime();
+        let progress = self
             .audio
             .active_sample_playback_progress(path)
             .filter(|progress| progress.active)
-            .and_then(|progress| progress.progress)
-        else {
+            .and_then(|progress| progress.progress);
+        if progress.is_none() && !pending_runtime {
             return false;
-        };
+        }
         if !self.audio.promote_sample_playback_session_to_waveform(path) {
             return false;
         }
-        self.waveform
-            .current
-            .start_playback_after_audition_handoff(progress, 0.0, true);
-        self.audio.current_playback_span = Some((0.0, 1.0));
+        if let Some(progress) = progress {
+            self.waveform
+                .current
+                .start_playback_after_audition_handoff(progress, 0.0, true);
+            self.audio.current_playback_span = Some((0.0, 1.0));
+        }
         self.record_current_playback_history(0.0, 1.0);
         self.record_sample_last_played(path.to_owned(), context);
         self.ui.status.sample = format!("Playing {file_name}");
@@ -162,7 +165,11 @@ impl NativeAppState {
             "browser.sample_load.finish",
             Some("browser"),
             Some(file_name),
-            "waveform_ready_playback_continued",
+            if pending_runtime {
+                "waveform_ready_playback_pending_start"
+            } else {
+                "waveform_ready_playback_continued"
+            },
             started_at,
             None,
         );
--- a/src/native_app/audio/sample_load_actions/completion/loaded.rs
+++ b/src/native_app/audio/sample_load_actions/completion/loaded.rs
@@ -148,6 +148,9 @@ impl NativeAppState {
         else {
             return false;
         };
+        if !self.audio.promote_sample_playback_session_to_waveform(path) {
+            return false;
+        }
         self.waveform
             .current
             .start_playback_after_audition_handoff(progress, 0.0, true);
--- a/src/native_app/audio/sample_load_actions/completion/loaded.rs
+++ b/src/native_app/audio/sample_load_actions/completion/loaded.rs
@@ -73,6 +73,9 @@ impl NativeAppState {
         if self.start_pending_sample_playback(&path, &file_name, started_at, context) {
             return;
         }
+        if self.finish_completed_streamable_sample_playback_load(&path, &file_name, started_at) {
+            return;
+        }
         if !autoplay {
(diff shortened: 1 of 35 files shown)


src/native_app/audio/sample_load_actions/completion/loaded.rs
```diff
@@ -108,8 +110,17 @@ impl NativeAppState {
         if !self.audio.active_sample_playback_is_streamable(path) {
             return false;
         }
-        let progress = self.audio.playback_progress.progress.unwrap_or(0.0);
-        self.waveform.current.start_playback(progress);
+        let Some(progress) = self
```
PORTALSURFER: Required fix: A matching streamable navigation session that reaches terminal progress before waveform loading completes falls through this active-only continuation check and restarts full playback from 0.0. Treat terminal streamed progress as a completed audition: transfer the terminal played prefix/history to the loaded waveform, keep playback stopped, and suppress generic autoplay. Add a regression test covering inactive terminal progress with autoplay enabled.

PORTALSURFER: Fixed in 598c2072f. Terminal streamable navigation is now captured before transient-session cleanup, consumed only by the matching waveform load, recorded as the played prefix, and leaves playback stopped without queuing a second start. Added the inactive-terminal/autoplay regression. Exact-head validation: wavecrate test target compiled, smoke gate passed, signed release build passed.

src/native_app/audio/sample_load_actions/completion/loaded.rs
```diff
@@ -108,8 +140,17 @@ impl NativeAppState {
...
+            .audio
+            .active_sample_playback_progress(path)
+            .filter(|progress| progress.active)
+            .and_then(|progress| progress.progress)
+        else {
+            return false;
+        };
+        self.waveform
```
PORTALSURFER: Required fix: Active file-backed playback is attached to the loaded waveform here without promoting the existing runtime session from Transient/AudibleTransient to Waveform/WaveformVisible. Natural completion then takes transient cleanup, skips the generic waveform stop path, and can leave transport logically playing. Promote the same session before attaching the waveform playhead and add the active-load-first/terminal-progress-second regression.

PORTALSURFER: Fixed in 940aaa50e. Active streamable navigation is now promoted to Waveform/WaveformVisible before its playhead is attached, so terminal progress follows the waveform completion path and stops transport/session/span together. Added the load-first/terminal-second regression. Exact-head validation: wavecrate test target compiled, smoke gate passed, signed release build passed.

src/native_app/audio/sample_load_actions/completion/loaded.rs
```diff
@@ -108,8 +140,20 @@ impl NativeAppState {
         if !self.audio.active_sample_playback_is_streamable(path) {
```
PORTALSURFER, line 140: Required fix: A matching file-backed session can still be RuntimePending when waveform completion reaches this branch (runtime Started event not yet drained). Requiring published active progress then falls through and submits a duplicate full-sample request from frame zero. Retain and promote the original pending session/request ID, let its eventual Started event initialize waveform playback, and add the waveform-completes-before-Started regression.

PORTALSURFER: Fixed in b2efde509. Matching file-backed RuntimePending sessions are now promoted in intent while retaining their original runtime request ID and pending state. Waveform completion no longer submits a replacement; the eventual Started event initializes the waveform on the retained source. Added the waveform-completes-before-Started regression alongside the existing opposite ordering. Exact-head validation: wavecrate test target compiled, smoke gate passed, signed release build passed.

PORTALSURFER:
Review summary — clean at `b2efde509ddee876ee71d7e4769077eaf32ad0fb`.

The fresh ordinary and realtime/audio reviews found no blocking mismatch with the PR Goal or Definition of Done. All required-fix threads from earlier passes are resolved.

Non-blocking observation: rapid navigation now uses 3 ms `FadeOutPrevious` overlap. The callback-owned source list is currently an unbounded `Vec`, so a future realtime-hardening change could replace it with preallocated bounded storage or enforce a fixed overlap ceiling. Runtime request coalescing and the 3 ms fade mean the current diff does not demonstrate a capacity or callback-safety failure.

PR #855, opened 2026-07-21
OPT-1217 Show global database and cache usage

PORTALSURFER:
## Goal

Give users clear, read-only insight into the disk footprint of Wavecrate's global database and rebuildable cache from General Settings.

Linear: [OPT-1217](https://linear.app/boostnlvp/issue/OPT-1217/add-insight-into-the-size-of-the-global-wavecrate-databasecache-to-the)

## Scope and non-goals

- Report `library.db` plus active SQLite `-wal` / `-shm` sidecars.
- Recursively report regular files under the global rebuildable `cache/` root without following symbolic links.
- Refresh asynchronously whenever Settings opens and after cache cleanup; fence stale completions.
- Render a human-readable total and database/cache breakdown in General Settings.
- Excludes source-local `.wavecrate.db` files, logs, config, handoff staging, cleanup policy, and storage limits.

## Definition of Done

- General Settings shows the current global database/cache total and breakdown.
- Filesystem traversal does not block the GUI thread.
- Missing database/cache paths report zero and transient deletion races do not fail the measurement.
- Cache cleanup refreshes the displayed value.
- Formatting, scope, async routing, and stale-result behavior have regression coverage.
- The exact branch build is exercised in a uniquely identified sandbox app.

## Validation

- `CARGO_TARGET_DIR=/tmp/wavecrate-opt-1217.LITP1d CARGO_INCREMENTAL=0 cargo check -j2 -p wavecrate --all-targets` — passed.
- `CARGO_TARGET_DIR=/tmp/wavecrate-opt-1217.LITP1d CARGO_INCREMENTAL=0 cargo test -p wavecrate-library --lib app_dirs::storage_usage` — 5 passed.
- Focused Wavecrate tests for projection/formatting, stale completion, Settings-open routing, and cache-clear refresh — passed.
- `bash scripts/ci.sh agent` — smoke, Radiant/Wavecrate architecture guardrails, Radiant tests, and 3,719 Wavecrate tests passed; the gate remains red on 52 unrelated pre-existing Wavecrate failures. A representative exact failure (`filter_section_projects_curation_scope_dropdown_and_dispatches_changes`) reproduces unchanged on pristine `origin/main` at `5c21e04dc`.
- `scripts/build.sh --release --variant OPT-1217 --no-build` — signed branch-specific app bundle passed code-sign verification.
- Manual sandbox runtime — General Settings rendered `Database + Cache`, `152.0 KiB total`, and the database/cache breakdown without clipping.

## Known limitations

- The repository-wide agent gate is not green because of the confirmed `origin/main` baseline failures described above; no OPT-1217-focused validation failed.

### DIFF
--- a/crates/wavecrate-library/src/app_dirs/storage_usage.rs
+++ b/crates/wavecrate-library/src/app_dirs/storage_usage.rs
@@ -50,6 +50,15 @@ fn global_storage_usage_at(root: &Path) -> Result<GlobalStorageUsage, String> {
 }
 
 fn directory_regular_file_size(root: &Path) -> Result<u64, String> {
+    let root_metadata = match fs::symlink_metadata(root) {
+        Ok(metadata) => metadata,
+        Err(error) if error.kind() == io::ErrorKind::NotFound => return Ok(0),
+        Err(error) => return Err(read_error("metadata", root, error)),
+    };
+    if !root_metadata.file_type().is_dir() {
+        return Ok(0);
+    }
+
     let mut total = 0_u64;
     let mut pending = vec![root.to_path_buf()];
     while let Some(directory) = pending.pop() {
@@ -152,5 +161,21 @@ mod tests {
         let usage = global_storage_usage_at(root.path()).expect("measure linked cache");
 
         assert_eq!(usage.cache_bytes, 5);
+    }
+
+    #[cfg(unix)]
+    #[test]
+    fn cache_traversal_does_not_follow_symbolic_cache_root() {
+        use std::os::unix::fs::symlink;
+
+        let root = tempfile::tempdir().expect("create storage root");
+        let external = tempfile::tempdir().expect("create external cache root");
+        fs::write(external.path().join("external.cache"), [0_u8; 61])
+            .expect("write external payload");
+        symlink(external.path(), root.path().join(CACHE_DIR_NAME)).expect("link cache root");
+
+        let usage = global_storage_usage_at(root.path()).expect("measure linked cache root");
+
+        assert_eq!(usage.cache_bytes, 0);
     }
 }
--- a/crates/wavecrate-library/src/app_dirs/storage_usage.rs
+++ b/crates/wavecrate-library/src/app_dirs/storage_usage.rs
@@ -2,6 +2,8 @@
 
 use std::{fs, io, path::Path};
 
+use cap_primitives::{ambient_authority, fs as capability_fs};
+
 use crate::sample_sources::LIBRARY_DB_FILE_NAME;
 
 use super::app_root_dir;
@@ -27,8 +29,8 @@ impl GlobalStorageUsage {
 
 /// Measure the current global library database and rebuildable cache footprint.
 ///
-/// The traversal does not follow symbolic links, so a link below the cache root
-/// cannot make the reported size escape the app-owned storage boundary.
+/// The traversal does not follow symbolic links, so a linked cache root or
+/// descendant cannot make the reported size escape the app-owned boundary.
 pub fn global_storage_usage() -> Result<GlobalStorageUsage, String> {
     let root = app_root_dir().map_err(|error| error.to_string())?;
     global_storage_usage_at(&root)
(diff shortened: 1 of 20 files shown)


crates/wavecrate-library/src/app_dirs/storage_usage.rs
```diff
@@ -0,0 +1,156 @@
...
+        database_bytes,
+        cache_bytes,
+    })
+}
+
+fn directory_regular_file_size(root: &Path) -> Result<u64, String> {
+    let mut total = 0_u64;
+    let mut pending = vec![root.to_path_buf()];
+    while let Some(directory) = pending.pop() {
+        let entries = match fs::read_dir(&directory) {
```
PORTALSURFER: **Required fix — High**

`pending` is seeded with `root` without first checking its file type, so `fs::read_dir` follows a symbolic-link `cache/` root. That traverses and counts an arbitrary external directory, contrary to the stated requirement to recursively report cache files **without following symbolic links**. It can also turn opening Settings into an unexpectedly expensive scan of external storage.

Before enqueuing/traversing the root, use `symlink_metadata` (or an equivalent no-follow directory open) and treat a missing, non-directory, or symlink cache root as zero. Add a regression test where `cache/` itself is a symlink to a populated external directory and assert the cache contribution remains zero.

PORTALSURFER (commented):
## Required fixes

1. **High — cache-root symlink is followed.** The measurement must not traverse symbolic links anywhere in `cache/`, including the root itself.

Verdict: **changes requested**. I could not submit a formal request-changes review because this GitHub identity is the PR author.

crates/wavecrate-library/src/app_dirs/storage_usage.rs
PORTALSURFER: Fixed in 2d0cc61e3. The cache root is now checked with no-follow `symlink_metadata` before traversal; missing, non-directory, and symlink roots contribute zero. Added `cache_traversal_does_not_follow_symbolic_cache_root`; all 4 storage-usage tests pass, and `cargo check -p wavecrate --all-targets` passes.

PORTALSURFER (commented):
## Required fixes

- **P1 — Required fix:** The cache walk can follow a cache entry (or the cache root) that is replaced by a symlink after its `symlink_metadata`/`file_type` check and before the later `read_dir` call. This breaks the stated no-symlink traversal boundary and can report arbitrary external storage or perform an unexpectedly large scan. Please make the traversal descriptor-relative with no-follow semantics (or an equivalent race-safe approach) so every opened directory is guaranteed not to be a symlink, including the root; add regression coverage for the replacement race.

I could not submit a formal “request changes” event because GitHub identifies this as my own pull request. Reviewed only the current PR description and merge-base-to-head diff at `2d0cc61e35d895dc3088653c0f0560c9e45bb85a`.

crates/wavecrate-library/src/app_dirs/storage_usage.rs
```diff
@@ -0,0 +1,181 @@
...
+    let root_metadata = match fs::symlink_metadata(root) {
+        Ok(metadata) => metadata,
+        Err(error) if error.kind() == io::ErrorKind::NotFound => return Ok(0),
+        Err(error) => return Err(read_error("metadata", root, error)),
+    };
+    if !root_metadata.file_type().is_dir() {
+        return Ok(0);
+    }
+
+    let mut total = 0_u64;
```
PORTALSURFER, line 86: **Required fix — P1:** `entry.file_type()` only classifies this pathname at this instant. If `path` is replaced with a symlink before the later `fs::read_dir(&directory)` at line 63, `read_dir` follows that link and the walk counts files outside the global `cache/` root. The same check-then-open gap exists for the root after its metadata check. That contradicts the PR’s explicit requirement to recurse without following symbolic links and can make a cache-size refresh scan/report arbitrary external storage. Traverse using directory descriptors opened with no-follow semantics (or an equivalent race-safe mechanism), and cover a replacement-after-classification case; a second path metadata check alone remains racy.

PORTALSURFER: Fixed in 245b8ae6d. Cache traversal now opens the root and every child directory descriptor-relative with `cap_primitives::fs::open_dir_nofollow`, then enumerates each already-open handle. Files are measured with no-follow entry metadata. Added a deterministic replacement test that swaps an enumerated directory for an external symlink immediately before descent; all 5 storage tests, focused UI/routing tests, macOS all-targets check, and the smoke gate pass. A Windows cross-check reached the existing bundled-SQLite C build and was blocked by the missing Windows C SDK (`stdlib.h`), after the new capability dependency itself compiled for the target.

PORTALSURFER (commented):
Reviewed the PR description and complete merge-base-to-head diff at `245b8ae6d60ea57c5992217e4f94616abcf35eb2`. The implementation correctly delivers asynchronous global database/cache usage reporting, the specified accounting scope, stale-completion fencing, and post-cleanup refresh coverage. No required fixes found.

PR #858, opened 2026-07-21
OPT-1224: batch waveform cache directory pruning

PORTALSURFER:
## Goal

Avoid a full waveform-cache directory scan after every successful cache write while preserving bounded cache growth and crash/startup recovery.

## Scope and non-goals

- Move cache reconciliation from the individual write transaction onto the existing single background store worker.
- Reconcile at worker startup, after 64 successful writes, after 16 MiB of conservatively counted cache artifacts, or after 250 ms of writer idleness.
- Keep the on-disk scan authoritative; do not add a durable cache index or change the cache format.
- Add scan/metadata instrumentation and a repeatable filesystem benchmark.
- Non-goal: redesign cache identity, invalidation, remapping, or playback payload policy.

## Definition of Done

- A burst of cache writes performs a bounded number of full directory scans instead of one scan per write.
- The 2 GiB cache limit and its transient overshoot tolerance are documented and enforced by threshold or idle reconciliation.
- Startup reconciliation removes interrupted/stale artifacts without persistent bookkeeping.
- Publication and pruning are serialized on the cache writer worker; shutdown flushing includes pending pruning.
- Diagnostics expose scan cadence and metadata work.

## Validation

- PASS: cargo test -p wavecrate --lib waveform_cache -- --test-threads=1 (48 passed, 1 ignored)
- PASS: cargo test -p wavecrate --lib benchmark_batched_waveform_cache_pruning_reduces_directory_metadata_work -- --ignored --nocapture --test-threads=1
  - 1,024 populated entries and 128 burst writes
  - scans: 128 baseline vs 3 batched
  - metadata probes: 417,984 baseline vs 9,792 batched
  - wall time: 1,063 ms baseline vs 33 ms batched
- BLOCKED (unrelated baseline): bash scripts/ci.sh agent reaches an existing invalid format string in unchanged tests/release_workflow_helpers.rs:936.
- NOT RUN: manual sandbox source-wide warming; the deterministic populated-cache benchmark covers scan cadence and disk-threshold behavior for this backend-only change.

## Known limitations

- The repository-wide agent gate remains blocked by the unchanged baseline compile error noted above.
- Manual sandbox source-wide warming remains for review smoke testing.

Linear: OPT-1224


src/native_app/waveform/audio_file/waveform_cache/store_queue.rs
```diff
@@ -184,22 +191,64 @@ impl BackgroundStoreQueue {
...
-            let job = self.next_job();
-            let cache_path = job.cache_path.clone();
-            let outcome = write_cached_waveform_file_now(job);
-            log_store_completion(&cache_path, outcome);
-            self.finish_job(&cache_path);
+            match self.next_worker_action(&prune_schedule) {
+                StoreWorkerAction::Write(job) => {
+                    let cache_path = job.cache_path.clone();
+                    let outcome = write_cached_waveform_file_now(job);
+                    let cache_written = matches!(&outcome, StoreWriteOutcome::Completed(_));
```
PORTALSURFER, line 201: Required fix: This threshold records only `cache_path` (`.wfc`). Reconciliation, however, probes the associated playback sidecar and descriptor when computing cache usage. A write with a small `.wfc` and large companion data can therefore stay below 16 MiB throughout a continuous burst, postponing reconciliation until 64 writes and exceeding the documented transient cache-growth bound. Account conservatively for all published artifacts included in cache usage (or mark an incomplete measurement as unknown so it reconciles immediately), and add a regression test that exercises a large companion artifact.

PORTALSURFER (commented):
Review finding: required cache-limit accounting fix below. I could not submit a formal changes request because this GitHub identity owns the pull request.

src/native_app/waveform/audio_file/waveform_cache/store_queue.rs
PORTALSURFER: Fixed in 86f9f2469. The scheduler now sums the published .wfc, playback sidecar, and descriptor sizes; any missing required file or non-NotFound companion metadata failure is treated as unknown, which forces immediate reconciliation. Added a sparse large-descriptor regression. Validation: waveform_cache suite 48 passed/1 ignored; benchmark remains 128 baseline scans vs 3 batched.

PORTALSURFER (commented):
Reviewed current head `86f9f24693fb7917e9e317c32abd2fd236798355` against the PR description and complete merge-base diff. The worker-owned startup, threshold, and idle reconciliation keeps pruning serialized with publication; the documented bounded-overage policy, recovery scan, shutdown drain, diagnostics, and focused coverage are all present. No required fixes found. **Verdict: clean.**

PR #859, opened 2026-07-21
Fix Windows settings replacement failure safety

PORTALSURFER:
## Goal

Preserve an existing valid Wavecrate settings file when Windows cannot atomically replace it.

Linear: OPT-1223

## Scope and non-goals

- Replace the Windows delete-then-rename fallback with one native `MoveFileExW` call using replace-existing and write-through flags.
- Keep the completed temporary-file write and cleanup behavior.
- Add deterministic coverage for replacement failure, post-replacement durability failure, cleanup, and retry.
- Add a Windows-only locked-destination regression using real Windows sharing semantics.
- No settings schema, migration, or public configuration API changes.

## Definition of Done

- A failed Windows replacement leaves the previous complete settings file in place.
- A successful replacement retains atomic visibility and write-through intent.
- Failed temporary files are cleaned up, and retry succeeds without recovery artifacts.
- A failure after replacement leaves the complete new settings file recoverable.
- Existing Unix atomic-save coverage remains green.

## Validation

- `cargo test --lib sample_sources::config::config_io -- --nocapture` — passed (29 tests).
- `bash scripts/ci.sh agent` — blocked by an unchanged baseline format-string error in `tests/release_workflow_helpers.rs:936`.
- `cargo check --tests --target x86_64-pc-windows-msvc -j 2` — reached Windows-target dependencies, then stopped in unchanged dependency `ring` because this macOS host lacks the Windows C headers/MSVC toolchain (`assert.h` not found).
- Independent review of head `8d3a7a323cf801bbf079632b043d18b09ef11523` — clean.

## Known limitations

- This repository has no PR CI workflow, so the Windows-only locked-destination test still requires execution on a Windows host.
- The repository-wide agent gate and strict clippy run have unrelated baseline blockers noted in the validation record.

### DIFF
--- a/src/sample_sources/config_io/save.rs
+++ b/src/sample_sources/config_io/save.rs
@@ -139,6 +139,14 @@ pub(super) fn save_settings_to_path(
 }
 
 fn atomic_write(path: &Path, data: &[u8]) -> Result<(), ConfigError> {
+    atomic_write_with(path, data, publish_temp_file)
+}
+
+fn atomic_write_with(
+    path: &Path,
+    data: &[u8],
+    publish: impl Fn(&Path, &Path, &Path) -> Result<(), ConfigError>,
+) -> Result<(), ConfigError> {
     use rand::TryRngCore;
     let dir = path.parent().ok_or_else(|| ConfigError::Write {
         path: path.to_path_buf(),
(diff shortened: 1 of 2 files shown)


PORTALSURFER (commented):
Clean review — reviewed merge-base-to-current-head diff at . The implementation meets the stated goal and Definition of Done: Windows replacement no longer deletes the existing settings file before publishing, failure cleanup and retry behavior are covered, and the Windows locked-destination regression verifies native sharing semantics. No required fixes found.

PR #862, opened 2026-07-21
OPT-1227 Bound macOS parent-workspace validation stalls

PORTALSURFER:
## Goal

Make supported Wavecrate macOS parent-workspace validation complete reliably or fail within a documented bound with actionable diagnostics and validation process-group cleanup.

## Root cause

The reproduced compiler was not blocked in Radiant code or Cargo scheduling. A macOS `sample` captured `rustc` in `rustc_metadata::creader -> dyld -> fcntl` while loading a proc-macro dylib. The shared `target/debug/deps` directory had grown to 40,610,912 bytes of directory metadata with its link count saturated at 65,535; an independent reader was simultaneously blocked in `getdirentries64`. A standalone Radiant target avoided that accumulated flat directory.

## Scope

- Route supported macOS Bash `scripts/agent.sh` and `scripts/ci.sh` lanes through a Rust-host/release and `Cargo.lock` keyed validation target, overriding any inherited `CARGO_TARGET_DIR`.
- Atomically quarantine a validation target before Cargo starts if its own `debug/deps` metadata becomes pathological.
- Serialize each keyed target with a bounded PID-plus-start-identity lease that safely recovers stale, zombie, and reused-PID owners.
- Add an owned-process-group watchdog that uses process-tree changes and aggregate CPU time as progress.
- After five idle minutes, capture command, process tree, Cargo/Rust/macOS versions, and macOS compiler samples within one 30-second global budget; allow a full two further minutes, then terminate the validation process group and exit 124. Including polling and termination grace, the default scheduled bound is 7 minutes 50 seconds after last observed progress.
- Install cancellation handling and cleanup coverage before spawning the validation group, and use that cleanup path for cancellation, ordinary failures, and background children that remain in the group after their parent exits.
- Document the validation, diagnostics, target-cache, and troubleshooting contracts.

Non-goals: no Radiant activation changes, no guardrail weakening, no dependency/toolchain pin change, no broad deletion of the existing shared target, and no attempt to kill arbitrary programs that deliberately detach into a separate session.

## Definition of Done

- [x] Clean/dependency-rebuilding parent compilation advances through the formerly stalled Radiant and Wavecrate architecture stages.
- [x] Warm parent validation reuses the isolated target and advances deterministically.
- [x] Genuine idle work produces bounded diagnostics and exit 124.
- [x] Success, failure, cancellation, and stall paths leave no process in the watchdog-created group while unrelated processes survive.
- [x] Existing guardrails keep their assertions and normal failure behavior.
- [x] Standalone locked Radiant validation still passes.

## Validation

- `bash scripts/internal/validation/test_validation_watchdog.sh` — pass (CPU progress, idle classification, globally bounded slow multi-process diagnostics, immediate pre-spawn and post-spawn cancellation, success/failure process-group cleanup, unrelated-process isolation, ambient target override, concurrent target rotation, stale/reused-PID lease recovery, and live-owner timeout).
- `WAVECRATE_VALIDATION_WATCHDOG_ACTIVE=1 bash scripts/internal/validation/test_validation_watchdog.sh` — pass through the nested-wrapper path.
- `bash scripts/internal/check/check_script_guardrails.sh` — pass.
- `bash scripts/agent.sh preflight` from the new empty validation target — the formerly stalled Radiant guardrails, Wavecrate GUI boundary compile, and slow-handler harness pass; lane then fails normally at pre-existing facade violations on `main`.
- Repeated warm `bash scripts/agent.sh preflight` — the same Rust stages pass using the isolated target; same pre-existing facade guardrail failure. The keyed lease is released and no Cargo/rustc/watchdog process remains.
- `CARGO_TARGET_DIR=target bash scripts/agent.sh request` — the ambient target is overridden with the keyed validation target; Rust stages advance normally to the same pre-existing facade failure, with clean lease/process cleanup.
- `bash scripts/ci.sh agent` — parent dependency/feature rebuild advances without stalling; smoke stops normally at the pre-existing invalid format string in `tests/release_workflow_helpers.rs:936`.
- Standalone Radiant archive using Radiant's own lockfile:
  - `cargo test --locked --lib --no-default-features` — 1,663 passed.
  - `cargo test --locked --test app_runtime_api --no-default-features` — 50 passed.
- Exact reproduction metadata: rustc 1.93.0, Cargo 1.93.0, macOS 26.5.2 (25F84), aarch64-apple-darwin.
- Post-run process checks found no owned Cargo/rustc/watchdog process.

## Known limitations

The current `main` branch has unrelated facade-guardrail violations and an unrelated format-string compile error in `tests/release_workflow_helpers.rs:936`. This PR preserves those failures and does not expand scope to fix them.


scripts/internal/validation/test_validation_watchdog.sh
```diff
@@ -0,0 +1,100 @@
...
+kill -0 "$owned_child_pid" 2>/dev/null && fail "owned child survived cancellation"
+echo "[validation_watchdog_test] PASS: cancellation cleanup"
+
+target_root="$FIXTURE_DIR/targets"
+export WAVECRATE_VALIDATION_TARGET_ROOT="$target_root"
+export WAVECRATE_VALIDATION_MAX_DEPS_METADATA_BYTES=1
+unset CARGO_TARGET_DIR
+# shellcheck source=scripts/internal/validation/use_validation_target.sh
+source "$ROOT_DIR/scripts/internal/validation/use_validation_target.sh"
+wavecrate_use_validation_target "$ROOT_DIR"
```
PORTALSURFER, line 162: Required fix: this macOS-only target-selection fixture runs unconditionally. On Linux/WSL `wavecrate_use_validation_target` returns without exporting `CARGO_TARGET_DIR`, and the next line exits under `set -u`; because script guardrails invoke this fixture on every platform, the existing Bash guardrail lane regresses. Gate or explicitly emulate the Darwin-only portion and assert the non-Darwin success path.

scripts/internal/validation/use_validation_target.sh
```diff
@@ -0,0 +1,41 @@
...
+  deps_dir="$target_dir/debug/deps"
+
+  local max_metadata_bytes="${WAVECRATE_VALIDATION_MAX_DEPS_METADATA_BYTES:-8388608}"
+  if [[ -d "$deps_dir" ]]; then
+    local metadata_bytes link_count
+    metadata_bytes="$(stat -f '%z' "$deps_dir")"
+    link_count="$(stat -f '%l' "$deps_dir")"
+    if (( metadata_bytes > max_metadata_bytes || link_count >= 60000 )); then
+      local quarantine="$target_root/stale-$(date -u +%Y%m%dT%H%M%SZ)-$$"
+      mv "$target_dir" "$quarantine"
```
PORTALSURFER, line 137: Required fix: inspection and `mv` are unsynchronized. Concurrent supported validations can both observe the same pathological target, race the move, or rotate a target another run already uses. Serialize inspection/rotation per keyed target, recheck after acquiring ownership, and cover two concurrent rotations: both callers succeed, select one fresh target, and produce exactly one quarantine.

PORTALSURFER (commented):
Independent review at 91967518e found two required fixes before the PR meets its validation reliability contract.

scripts/internal/validation/test_validation_watchdog.sh
PORTALSURFER: Fixed in 568bb12d2. The fixture now forces and asserts the non-Darwin no-op path before running Darwin-specific target tests. It also passes both directly and nested under an already-active watchdog via the integrated script guardrail.

scripts/internal/validation/use_validation_target.sh
PORTALSURFER: Fixed in 568bb12d2. A per-keyed-target PID lease now covers inspection, rotation, and the full validation command; concurrent callers wait, dead owners are recovered, and nested commands release a newly acquired lease. Repeated concurrent fixtures prove both callers select the same fresh target, exactly one quarantine is created, and no lease remains.

scripts/internal/validation/use_validation_target.sh
```diff
@@ -0,0 +1,86 @@
...
+  owner_pid="${BASHPID:-$$}"
+  mkdir -p "$target_root"
+
+  local missing_owner_attempts=0
+  while ! mkdir "$lease_dir" 2>/dev/null; do
+    local current_owner=""
+    IFS= read -r current_owner 2>/dev/null < "$lease_dir/pid" || current_owner=""
+    if [[ "$current_owner" =~ ^[0-9]+$ ]] && kill -0 "$current_owner" 2>/dev/null; then
+      missing_owner_attempts=0
+      sleep 0.1
```
PORTALSURFER, line 113: Required fix: `kill -0` alone treats zombies and reused PIDs as the live lease owner, then waits forever before the watchdog starts. Record and validate a non-reusable process identity (PID plus start identity/state), treat mismatches/zombies as stale, and bound the wait for a genuinely live owner so this entrypoint still exits predictably without disturbing that owner. Add deterministic stale-identity recovery and live-owner timeout coverage.

PORTALSURFER (commented):
Fresh stateless review at 568bb12d2 found one remaining required fix in pre-watchdog lease recovery.

scripts/internal/validation/use_validation_target.sh
PORTALSURFER: Fixed in 7f2666979. Leases now record PID plus process start identity, reject identity mismatches and zombie owners as stale, and bound waits for a genuinely live owner (default 480s, exit 124 without disturbing the owner). The fixture deterministically covers reused-PID recovery and a one-second live-owner timeout, and the exact-head warm parent preflight released its lease cleanly after reaching the known unrelated facade guardrail failure.

scripts/internal/validation/test_validation_watchdog.sh
```diff
@@ -0,0 +1,168 @@
...
+export WAVECRATE_VALIDATION_MAX_DEPS_METADATA_BYTES=1
+unset CARGO_TARGET_DIR
+# shellcheck source=scripts/internal/validation/use_validation_target.sh
+source "$ROOT_DIR/scripts/internal/validation/use_validation_target.sh"
+export WAVECRATE_VALIDATION_TEST_PLATFORM=Linux
+wavecrate_use_validation_target "$ROOT_DIR"
+[[ -z "${CARGO_TARGET_DIR:-}" ]] || fail "non-Darwin target selection changed CARGO_TARGET_DIR"
+echo "[validation_watchdog_test] PASS: non-Darwin target selection is a no-op"
+
+export WAVECRATE_VALIDATION_TEST_PLATFORM=Darwin
```
PORTALSURFER, line 159: Required fix: the fixture forces the Darwin target path on every host, but target metadata still uses BSD-only `stat -f`. The unconditionally integrated script guardrail therefore fails on GNU/Linux. Make metadata inspection portable or inject it, while retaining the explicit non-Darwin no-op assertion.

scripts/internal/validation/run_with_progress_watchdog.py
```diff
@@ -0,0 +1,302 @@
...
+        "pid\tppid\tcpu_seconds\tstate\tcommand\n" + "\n".join(process_lines) + "\n",
+        encoding="utf-8",
+    )
+    run_and_record(destination / "rustc-version.txt", ["rustc", "-Vv"])
+    run_and_record(destination / "cargo-version.txt", ["cargo", "-Vv"])
+    if platform.system() == "Darwin":
+        run_and_record(destination / "macos-version.txt", ["sw_vers"])
+        sample_seconds = int(float(os.environ.get("WAVECRATE_VALIDATION_SAMPLE_SECONDS", "3")))
+        if sample_seconds > 0:
+            for item in processes.values():
```
PORTALSURFER, line 174: Required fix: compiler samples are individually timed out but unbounded in count, so total diagnostics can exceed the documented stall bound; the recovery timer also starts before collection and can expire during sampling. Apply one global collection budget, start the full recovery grace after collection, document the maximum bound, and add deterministic slow multi-sample coverage.

PORTALSURFER (commented):
Fresh exact-head review found two required fixes.

scripts/internal/validation/test_validation_watchdog.sh
PORTALSURFER: Fixed in ac4e1cdc8. Validation-target directory metadata now comes from Python `os.stat` (`st_size`/`st_nlink`), so the explicitly emulated Darwin target-rotation fixture no longer depends on BSD `stat -f` and remains runnable under GNU/Linux. The full script guardrail integration passes.

scripts/internal/validation/run_with_progress_watchdog.py
PORTALSURFER: Fixed in ac4e1cdc8. Toolchain metadata and all compiler samples now share one 30-second global collection deadline; the full 120-second recovery grace begins after collection, and docs state the 7m50 default scheduled bound including polling and termination grace. A deterministic fixture presents three compiler candidates to a deliberately unresponsive sample command, proves the shared budget and post-collection grace, and the integrated guardrail passes.

scripts/internal/validation/run_with_progress_watchdog.py
```diff
@@ -0,0 +1,338 @@
...
+    return False
+
+
+def terminate_owned_group(child: subprocess.Popen[bytes], grace_seconds: float) -> None:
+    if child.poll() is not None and not process_group_exists(child.pid):
+        return
+    try:
+        os.killpg(child.pid, signal.SIGTERM)
+    except ProcessLookupError:
+        return
```
PORTALSURFER, line 227: Required fix: cleanup is intentionally bounded to the watchdog-created process group, but the PR description promises cleanup of every owned child, including a child that deliberately detaches into another session. Either safely track and clean detached descendants or narrow the PR contract to the process-group boundary required for Cargo/rustc so it does not overclaim arbitrary cross-session cleanup.

PORTALSURFER (commented):
Fresh exact-head review found one contract mismatch.

scripts/internal/validation/run_with_progress_watchdog.py
PORTALSURFER: Resolved as a PR-contract correction at unchanged head ac4e1cdc8. The Goal, Scope, Definition of Done, validation wording, and non-goals now consistently define cleanup at the watchdog-created process-group boundary required for Cargo/rustc. The PR no longer claims it can safely kill arbitrary programs that deliberately detach into a separate session.

PORTALSURFER (commented):
Fresh exact-head review found two required fixes.

scripts/internal/validation/run_with_progress_watchdog.py
```diff
@@ -0,0 +1,338 @@
...
+    term_grace = env_seconds("WAVECRATE_VALIDATION_TERM_GRACE_SECONDS", 10)
+    poll_seconds = env_seconds("WAVECRATE_VALIDATION_POLL_SECONDS", 5)
+    diagnostics_root = Path(
+        os.environ.get(
+            "WAVECRATE_VALIDATION_DIAGNOSTICS_DIR",
+            str(Path.cwd() / "target" / "validation-diagnostics"),
+        )
+    )
+
+    child = subprocess.Popen(command, start_new_session=True)
```
PORTALSURFER: Required fix: the detached child is spawned before cancellation handlers and cleanup coverage are established. A signal in that interval can exit the watchdog and orphan its validation process group. Install handlers and cleanup state before spawn, guard pre-spawn cancellation, and add deterministic immediate-cancellation coverage.

scripts/internal/validation/use_validation_target.sh
```diff
@@ -0,0 +1,130 @@
...
+PY
+}
+
+wavecrate_use_validation_target() {
+  local root_dir="$1"
+  local platform="${WAVECRATE_VALIDATION_TEST_PLATFORM:-$(uname -s)}"
+  if [[ "$platform" != "Darwin" ]]; then
+    return 0
+  fi
+  if [[ -n "${CARGO_TARGET_DIR:-}" ]]; then
```
PORTALSURFER: Required fix: an inherited `CARGO_TARGET_DIR` silently bypasses the keyed isolated target and can select the original pathological cache. Supported macOS entrypoints must override or reject ambient targets, while preserving the wrapper's own nested keyed target without reacquiring its lease. Add regression coverage.

scripts/internal/validation/run_with_progress_watchdog.py
PORTALSURFER: Fixed in 2c0528f65. Signal handlers, nullable child state, and the cleanup `try/finally` are established before spawn; a recorded pre-spawn signal returns 128+signal without creating a child, while any post-spawn signal reaches group termination and wait. A deterministic ready-file fixture cancels inside the pre-spawn window and proves the command never starts.

scripts/internal/validation/use_validation_target.sh
PORTALSURFER: Fixed in 2c0528f65. Supported macOS selection now overrides ambient `CARGO_TARGET_DIR` with the keyed target and documents `WAVECRATE_VALIDATION_TARGET_ROOT` as the customization point. Only a nested call already marked inside the active watchdog may reuse the exact selected target without reacquiring its parent lease. The fixture and a real `CARGO_TARGET_DIR=target bash scripts/agent.sh request` run both prove the override.

scripts/internal/validation/run_validation_command.sh
```diff
@@ -0,0 +1,30 @@
...
+
+if [[ "$(uname -s)" != "Darwin" ]]; then
+  exec "$@"
+fi
+
+# shellcheck source=scripts/internal/validation/use_validation_target.sh
+source "$ROOT_DIR/scripts/internal/validation/use_validation_target.sh"
+wavecrate_use_validation_target "$ROOT_DIR"
+
+if [[ "${WAVECRATE_VALIDATION_WATCHDOG_ACTIVE:-0}" == "1" ]]; then
```
PORTALSURFER: Required fix: an inherited `WAVECRATE_VALIDATION_WATCHDOG_ACTIVE=1` is currently trusted without proving an enclosing watchdog, so a top-level supported lane can bypass diagnostics, timeout, and process-group cleanup. Verify the enclosing watchdog via lease owner PID/start identity and ancestry (and the watchdog command), otherwise treat the marker as untrusted and start normal supervision. Add a preset-marker regression.

PORTALSURFER (commented):
Fresh exact-head review found one required fix.

scripts/internal/validation/run_validation_command.sh
PORTALSURFER: Fixed in 35e3ba106. Nested bypass now requires a live enclosing watchdog whose PID/start identity matches the lease record, whose command is `run_with_progress_watchdog.py`, and whose PID is an ancestor of the current wrapper. An ambient marker fails that proof, acquires a normal top-level lease, and starts supervision. The fixture proves a preset marker still captures diagnostics, exits 124, and releases its lease; a real preset-marker `scripts/agent.sh request` run also advanced under the keyed target to the known facade failure with clean cleanup.

PORTALSURFER (commented):
Fresh stateless review of the current PR description and complete `origin/main...35e3ba106` diff: clean. The diff achieves the stated Goal and Definition of Done; no required fixes or non-blocking observations found.

PR #863, opened 2026-07-21
OPT-1219: Prevent starmap audition UI flashes

PORTALSURFER:
## Goal
Prevent the whole Wavecrate window from flashing while dragging across starmap nodes, releasing an active drag, or selecting another node immediately afterward.

## Scope and non-goals
- Keep the existing declarative scene retained across the starmap gesture and asynchronous audition handoff.
- Downgrade eligible background progress and sample-loading messages to paint-only while that retained-scene window is active.
- Restore the normal correctness-first surface refresh after the starmap audition session ends.
- Preserve existing starmap selection, waveform restoration, and audition behavior.
- Non-goal: change starmap hit testing, playback scheduling, or waveform restore semantics.

## Definition of Done
- Pointer-down, active drag, audio replacement, pointer-up, queued load, and audible handoff frames do not rebuild the full scene.
- Background source-processing progress cannot force a starmap scene reprojection during audition.
- Background source-processing progress resumes normal projection once the starmap audition session ends.n- Normal non-starmap projection refresh remains unchanged.
- The signed macOS app no longer flashes in the reported drag-release-click sequence and audition remains immediate.

## Validation
- `cargo test -p wavecrate --lib starmap_drag_begin_and_release_frames_stay_paint_only -- --nocapture` — passed
- `cargo test -p wavecrate --lib starmap_audition_handoff_stays_paint_only -- --nocapture` — passed
- `cargo test -p wavecrate --lib scene_starmap_drag_frame_uses_paint_only_repaint_scope_during_background_progress -- --nocapture` — passed
- `cargo test -p wavecrate --lib source_processing_progress_does_not_reproject_during_starmap_audition -- --nocapture` — passed
- `cargo check --release -p wavecrate --bin wavecrate` — passed
- `git diff --check` — passed
- `bash scripts/build.sh --release --variant OPT-1219` — passed; signed app at `target/app-bundles/opt1219/Wavecrate OPT-1219.app`
- Real-app macOS validation with the configured 57-node starmap — drag, release, and immediate follow-up selections were clean in the quiet release build; audition latency remained immediate.

## Known limitations
- The repository-wide smoke lane is currently blocked on `main` by the pre-existing invalid format string in `tests/release_workflow_helpers.rs:936`; this PR does not modify that test.
- Release-mode library tests are currently blocked on `main` because debug-only source DB test helpers are unavailable under release assertions.

### DIFF
--- a/src/native_app/audio/playback/progress/telemetry.rs
+++ b/src/native_app/audio/playback/progress/telemetry.rs
@@ -6,7 +6,9 @@ use radiant::runtime::{RepaintScope, SurfaceRevisions};
 
 use super::super::diagnostics::PlayheadFrameMessageDiagnostics;
 use crate::native_app::{
-    app::{NativeAppState, SamplePlaybackSession},
+    app::{
+        NativeAppState, SamplePlaybackIntent, SamplePlaybackSession, SamplePlaybackSessionState,
+    },
     starmap_audition_telemetry::{
         self as starmap_telemetry, StarmapAuditionCounter, StarmapAuditionDuration,
     },
@@ -22,7 +24,7 @@ thread_local! {
 pub(in crate::native_app) struct FrameSurfaceRevisionTracker {
     last: Option<FrameSurfaceRevisionInputs>,
     revisions: SurfaceRevisions,
-    playback_fast_path_active: bool,
+    transient_fast_path_active: bool,
 }
 
 #[derive(Clone, Copy, Debug, PartialEq, Eq)]
@@ -64,19 +66,20 @@ struct FrameProjectionState {
 
 #[derive(Clone, Copy, Debug, PartialEq, Eq)]
 pub(in crate::native_app) struct FrameSurfaceRevisionGuard {
-    before: Option<PlaybackFrameRevisionInputs>,
+    before: Option<TransientFrameRevisionInputs>,
     forced_surface: bool,
+    starmap_retained_scene_active_before_message: bool,
 }
 
 #[derive(Clone, Copy, Debug, PartialEq, Eq)]
-struct PlaybackFrameRevisionInputs {
-    structure: PlaybackFrameStructureState,
+struct TransientFrameRevisionInputs {
+    structure: TransientFrameStructureState,
     layout: FrameLayoutState,
-    projection: PlaybackFrameProjectionState,
+    projection: TransientFrameProjectionState,
 }
 
 #[derive(Clone, Copy, Debug, PartialEq, Eq)]
-struct PlaybackFrameStructureState {
+struct TransientFrameStructureState {
     drag_hover_auto_expand_pending: bool,
     source_cache_progress_active: bool,
     sample_loading: bool,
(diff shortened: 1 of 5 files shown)


src/native_app/audio/playback/progress/telemetry.rs
```diff
@@ -187,6 +196,27 @@ impl NativeAppState {
...
+    }
+
+    pub(in crate::native_app) fn starmap_retained_scene_active(&self) -> bool {
+        self.ui.chrome.starmap_audition_drag.is_some()
+            || self
+                .audio
+                .sample_playback_session
+                .as_ref()
+                .is_some_and(|session| {
+                    session.request.intent == SamplePlaybackIntent::StarmapDrag
```
PORTALSURFER, line 210: Required fix — `starmap_retained_scene_active` stays true for every `StarmapDrag` playback session except `Failed(_)`. That has no success/settled boundary, so any retained successful session continues to take the transient fast path and makes background progress paint-only after the handoff should have returned to normal correctness-first projection. Restrict this predicate to the explicit drag/queued/resolving/audible states (or explicitly end the retained-session marker at successful settlement), and add a regression that a settled starmap session allows `SourceProcessingProgress` to request `Projection` again.

PORTALSURFER (commented):
Review outcome: changes requested. I cannot submit a formal request-changes review on my own pull request, so the required fix is recorded inline.

src/native_app/audio/playback/progress/telemetry.rs
PORTALSURFER, line 210: Required fix — `starmap_retained_scene_active` stays true for every `StarmapDrag` playback session except `Failed(_)`. That has no success/settled boundary, so any retained successful session continues to take the transient fast path and makes background progress paint-only after the handoff should have returned to normal correctness-first projection. Restrict this predicate to the explicit drag/queued/resolving/audible states (or explicitly end the retained-session marker at successful settlement), and add a regression that a settled starmap session allows `SourceProcessingProgress` to request `Projection` again.

PORTALSURFER: Addressed at 1a0c97e22. The full StarmapDrag playback session is intentionally retained because it can reach AudibleTransient before pointer-up; ending the guard at handoff would reopen the observed release flash. The focused regression now proves source progress stays PaintOnly through that audible post-drag session and returns to Projection immediately after clear_sample_playback_session(). The PR contract now names the session end as the exit condition.

PORTALSURFER (commented):
Clean review at `1a0c97e22d7a37505f454611be9dc3ee676a8d54`: the current merge-base-to-head diff satisfies the stated Goal and Definition of Done. It retains the scene for drag/queued/audible starmap audition work, keeps progress paint-only while the session is active, and restores progress projection after the session is cleared. No required fixes found.

src/native_app/audio/playback/progress/telemetry.rs
PORTALSURFER: Addressed at 1a0c97e22. The retained boundary is the full StarmapDrag playback session because the session can be AudibleTransient before pointer-up; ending it at handoff would reopen the observed release flash. The regression now proves source progress stays PaintOnly through that audible post-drag session and returns to Projection immediately after clear_sample_playback_session(). The current PR contract explicitly names session end as the exit condition, and the fresh exact-head review is clean.

PR #865, opened 2026-07-21
OPT-1220 Make destructive waveform edits failure-atomic

PORTALSURFER:
## Goal

Make every destructive waveform edit failure-atomic so write, finalize, replace, or rollback errors cannot silently destroy the user's original audio.

Linear: [OPT-1220](https://linear.app/boostnlvp/issue/OPT-1220/wavecrate-make-destructive-waveform-edits-failure-atomic)

## Scope

- Render edits to a unique sibling temporary WAV.
- Finalize and sync the rendered WAV before publishing it.
- Prepare and sync the redo snapshot before changing the source.
- Atomically replace the source on macOS/Linux and use `MoveFileExW(MOVEFILE_REPLACE_EXISTING | MOVEFILE_WRITE_THROUGH)` on Windows.
- Restore from a separately staged copy after a failed replace or directory sync.
- Retain and report the exact recovery-copy path if restore also fails.
- Preserve original file permissions and existing edit/undo/readiness behavior.
- Add deterministic write, finalize, snapshot, replace, restore, and success fixtures plus a debug-only injection seam for live-app smoke testing.

## Non-goals

- No DSP, selection math, sample-format, protected-source policy, undo UX, or transaction-history redesign.
- No changes to committed-mutation/readiness invalidation semantics.

## Definition of Done

- [x] Write or finalize failures leave the original byte-for-byte unchanged.
- [x] Replace failure restores the original or retains and reports a recovery copy.
- [x] Failed edits do not leave a truncated/corrupt original.
- [x] Focused tests cover write, finalize, replace, rollback, success, and undo/redo compatibility.
- [x] The real app reports an injected failure without losing the selected source file.

## Validation

Passed:

- `cargo test -p wavecrate --lib native_app::waveform_edits` — 6 passed
- `cargo test -p wavecrate --lib crop_request_rewrites_file_and_undo_restores_original_audio`
- `cargo test -p wavecrate --lib extract_and_trim_request_extracts_selection_trims_source_and_undo_redo_roundtrips`
- `cargo test -p wavecrate --bin wavecrate committed_file_mutations` — command passed; no matching binary tests
- `cargo check -p wavecrate --lib --bins`
- `git diff --check`
- Debug app bundle smoke with `WAVECRATE_INJECT_WAVEFORM_EDIT_FAILURE=replace`: Wavecrate reported that the crop failed and the original was restored; SHA-256 remained `814085a9b292f95851edfead6e51e470d4f9a13c0cdec22c64ec46ba7619153f`.

Blocked by unrelated existing infrastructure:

- `cargo check -p wavecrate --all-targets` reaches `tests/release_workflow_helpers.rs:936` and fails on the pre-existing unescaped `{` embedded-Python fixture.
- macOS-to-Windows cross-check cannot pass this host because the installed Rust target lacks the Windows C/MSVC sysroot required by `ring` and `libsqlite3-sys`. The Windows replacement code follows the repository's existing `MoveFileExW` implementation pattern.

## Known limitations

- The debug failure injection seam is available only in debug builds through `WAVECRATE_INJECT_WAVEFORM_EDIT_FAILURE`; release builds contain no injection behavior.
- Full all-target and Windows cross-target validation remain blocked as described above.

### DIFF
--- a/src/native_app/waveform_edits/atomic_write.rs
+++ b/src/native_app/waveform_edits/atomic_write.rs
@@ -244,6 +244,7 @@ fn restore_recovery_copy(
         .tempfile_in(parent)?;
     let mut recovery = fs::File::open(recovery_copy)?;
     io::copy(&mut recovery, staged.as_file_mut())?;
+    fs::set_permissions(staged.path(), fs::metadata(recovery_copy)?.permissions())?;
     staged.as_file().sync_all()?;
     let staged_path = staged.into_temp_path();
     edit_io.replace(&staged_path, target, ReplacePhase::Restore)?;
--- a/src/native_app/waveform_edits/atomic_write.rs
+++ b/src/native_app/waveform_edits/atomic_write.rs
@@ -1,0 +1,297 @@
+use std::{fs, io, path::Path};
+
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+enum ReplacePhase {
+    Commit,
+    Restore,
+}
+
+#[derive(Debug)]
+pub(super) struct AtomicWriteFailure {
+    message: String,
+    recovery_copy_required: bool,
+}
+
+impl AtomicWriteFailure {
+    fn ordinary(message: String) -> Self {
+        Self {
+            message,
+            recovery_copy_required: false,
+        }
+    }
+
+    fn recovery_required(message: String) -> Self {
+        Self {
+            message,
+            recovery_copy_required: true,
+        }
+    }
+
+    pub(super) fn recovery_copy_required(&self) -> bool {
+        self.recovery_copy_required
+    }
+}
+
+impl std::fmt::Display for AtomicWriteFailure {
+    fn fmt(&self, formatter: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        formatter.write_str(&self.message)
+    }
+}
+
+trait AtomicEditIo {
+    fn before_sample_write(&self, _sample_index: usize) -> Result<(), String> {
+        Ok(())
+    }
+
+    fn before_finalize(&self) -> Result<(), String> {
+        Ok(())
+    }
+
+    fn replace(&self, staged: &Path, target: &Path, _phase: ReplacePhase) -> io::Result<()> {
+        replace_file(staged, target)
+    }
+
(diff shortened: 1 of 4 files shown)


src/native_app/waveform_edits/atomic_write.rs
```diff
@@ -0,0 +1,297 @@
...
+    recovery_copy: &Path,
+    parent: &Path,
+    edit_io: &impl AtomicEditIo,
+) -> io::Result<()> {
+    let mut staged = tempfile::Builder::new()
+        .prefix(".wavecrate-restore-")
+        .suffix(".wav")
+        .tempfile_in(parent)?;
+    let mut recovery = fs::File::open(recovery_copy)?;
+    io::copy(&mut recovery, staged.as_file_mut())?;
```
PORTALSURFER, line 246: Required fix — `tempfile_in` creates a new restrictive file and `io::copy` transfers only bytes, so a restore after a failed replace or directory sync publishes the temp file’s permissions (for example, 0600) rather than the original WAV’s. This violates the stated requirement to preserve original permissions even though the error reports that the original was restored. Before syncing/replacing the restore staging file, apply `fs::metadata(recovery_copy)?.permissions()` to it, and add a failure fixture that asserts the restored target’s permissions match the original.

PORTALSURFER (commented):
Required fix identified. I cannot submit a change request because this GitHub identity authored the PR.

src/native_app/waveform_edits/atomic_write.rs
PORTALSURFER: Fixed in d080d44fa. The rollback staging file now receives the recovery copy's permissions before sync and replacement, and the replace-failure regression asserts restored target permissions match the recovery/original. Validation: `cargo test -p wavecrate --lib native_app::waveform_edits`, undo/redo round trip, `cargo check -p wavecrate --lib --bins`, and `git diff --check` all pass.

PORTALSURFER (commented):
Clean review — the current diff fulfills the PR goal and Definition of Done: it stages and syncs the rendered WAV and redo snapshot before publication, preserves permissions, restores from the retained original copy after replace or directory-sync failures, and retains/reports that copy if restoration fails. The deterministic fixtures cover the declared failure and success paths. Reviewed head: d080d44facffbb71793e5e1d398f000f4d45a73a.

PR #870, opened 2026-07-21
OPT-1221 Prevent late destination collisions in native transfers

PORTALSURFER:
## Goal
Prevent files created after transfer planning from being overwritten by native move, copy, conflict-resolution, and external-drop execution.

## Scope
- add a shared exclusive file-transfer primitive with staged no-clobber copy publication and platform no-replace rename
- apply it to folder-browser transfers, unique-name moves, overwrite backups, rollback cleanup, and external waveform drops
- preserve predictable `_copyNNN` allocation while retrying late copy/import collisions
- document the durable no-replace transfer contract and add deterministic race coverage

## Non-goals
- folder-browser redesign
- source-database transaction changes
- unrelated filesystem-operation cleanup

## Definition of Done
- [x] same-filesystem moves cannot replace a late destination
- [x] copy and external-drop commits publish only into an absent destination and retry predictable unique names
- [x] cross-device fallback commits a complete staged copy before source removal and rejects late collisions
- [x] rollback and cleanup act only on transaction-owned destination identities
- [x] deterministic tests inject late destinations for move, copy, fallback, and external-drop paths
- [x] existing conflict and cancellation routing remains intact

## Validation
Passed:
- `cargo test -p wavecrate --lib exclusive_file_transfer -- --nocapture` (10 passed)
- `cargo test -p wavecrate --lib file_move_transaction -- --nocapture` (7 passed)
- `cargo test -p wavecrate --lib native_file_drop_actions -- --nocapture` (2 passed)
- `cargo test -p wavecrate --lib committed_file_mutations -- --nocapture` (14 passed)
- `bash scripts/build.sh --release`
- signed sandbox app launch and isolated source-add smoke

Baseline-blocked on current `main` (unchanged files):
- `bash scripts/ci.sh agent` / `cargo check -p wavecrate --all-targets`: `tests/release_workflow_helpers.rs:936` has an invalid unescaped format-string brace
- agent request preflight: `tests/gui_boundary.rs` reports existing filesystem calls in `src/native_app/waveform_edits/atomic_write.rs`
- strict clippy: existing type-complexity/too-many-arguments findings in wavecrate-library and Radiant
- file-size guard: existing over-budget Radiant `native_file_open.rs` and `runner.rs`

Manual move/copy/rename/drop completion was limited by Radiant macOS activation/window availability after the successful app launch and source add; deterministic filesystem tests above cover the collision states.

## Known limitations
- Windows and Linux platform implementations are source-validated here; this local run exercised macOS.
- The unrelated current-main gate failures listed above prevent a fully green repository-wide lane.

Linear: OPT-1221

### DIFF
--- a/src/native_app/sample_library/exclusive_file_transfer.rs
+++ b/src/native_app/sample_library/exclusive_file_transfer.rs
@@ -5,6 +5,8 @@ use std::{
     sync::Arc,
 };
 
+use platform::{rename_no_replace, same_file_handles};
+
 #[derive(Clone, Debug)]
 pub(super) struct CommittedFile {
     path: PathBuf,
@@ -17,10 +19,51 @@ impl CommittedFile {
     }
 
     pub(super) fn remove_if_owned(&self) -> io::Result<bool> {
+        self.remove_if_owned_with(|| {})
+    }
+
+    fn remove_if_owned_with(&self, before_quarantine: impl FnOnce()) -> io::Result<bool> {
         if !self.still_owned()? {
             return Ok(false);
         }
-        fs::remove_file(&self.path)?;
+        before_quarantine();
+        let parent = self.path.parent().unwrap_or_else(|| Path::new(""));
+        let quarantine = tempfile::Builder::new()
+            .prefix(".wavecrate-cleanup-")
+            .tempdir_in(parent)?;
+        let quarantined = quarantine.path().join("owned-file");
+        rename_no_replace(&self.path, &quarantined)?;
+        // From this point onward, keep the private directory on every error path. Its
+        // destructor must never recursively remove an object that was moved from the
+        // user-visible destination before its identity was verified.
+        let quarantine = quarantine.keep();
+        let moved = fs::metadata(&quarantined)
+            .and_then(|metadata| {
+                metadata
+                    .is_file()
+                    .then(|| File::open(&quarantined))
+                    .transpose()
+            })
+            .map_err(|error| preserved_cleanup_error(error, &quarantined))?;
+        let owned = match moved.as_ref() {
+            Some(moved) => same_file_handles(&self.owned_file, moved)
+                .map_err(|error| preserved_cleanup_error(error, &quarantined))?,
+            None => false,
+        };
+        if !owned {
+            return restore_quarantined_destination(&quarantined, &self.path, &quarantine);
+        }
+        drop(moved);
+        if let Err(remove_error) = fs::remove_file(&quarantined) {
+            return Err(io::Error::new(
+                remove_error.kind(),
+                format!(
+                    "owned destination cleanup failed and was preserved at {}: {remove_error}",
+                    quarantined.display()
+                ),
+            ));
+        }
+        fs::remove_dir(quarantine)?;
         Ok(true)
     }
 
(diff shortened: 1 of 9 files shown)


src/native_app/sample_library/exclusive_file_transfer.rs
```diff
@@ -0,0 +1,349 @@
...
+impl CommittedFile {
+    pub(super) fn path(&self) -> &Path {
+        &self.path
+    }
+
+    pub(super) fn remove_if_owned(&self) -> io::Result<bool> {
+        if !self.still_owned()? {
+            return Ok(false);
+        }
+        fs::remove_file(&self.path)?;
```
PORTALSURFER: Required fix — `still_owned()` verifies the pathname before this call, but the unlink happens afterward. A file can be replaced or created at `self.path` between those operations, and `remove_file` will then delete that later owner. Both rollback (`move_back_if_owned`) and overwrite-backup cleanup route through this path, so this violates the Definition of Done that rollback and cleanup affect only transaction-owned destination identities. Use an ownership-preserving cleanup protocol whose final removal cannot target a rebound pathname (a second stat check has the same race), and add deterministic interleaving coverage for replacement after the identity check.

PORTALSURFER (commented):
Required fix recorded inline: rollback/cleanup ownership is not safe against pathname replacement at the current head.

src/native_app/sample_library/exclusive_file_transfer.rs
PORTALSURFER: Fixed in fc09c067f. Cleanup now atomically removes the public destination into a transaction-private quarantine, verifies identity there, restores any replacement without clobbering, and deletes only the verified owned object inside the private namespace. Added deterministic regressions for replacements injected after the initial check, including a non-empty directory. Exact-head focused validation: exclusive transfer 8/8, transaction 7/7, external drop 2/2, committed mutations 14/14.

src/native_app/sample_library/exclusive_file_transfer.rs
```diff
@@ -0,0 +1,287 @@
...
+            &mut input,
+            self.owned_file.metadata()?.permissions(),
+            source,
+        )?;
+        let _ = self.remove_if_owned();
+        Ok(true)
+    }
+
+    fn still_owned(&self) -> io::Result<bool> {
+        let actual = match File::open(&self.path) {
```
PORTALSURFER, line 83: **Required fix** — `File::open` resolves symlinks before `same_file_handles` compares identities. If a transaction-owned destination is replaced by a new symlink that targets the same file as `owned_file` (including a moved symlink source), `still_owned()` returns true and `remove_if_owned()` renames/removes that foreign directory entry. That violates the Definition of Done requirement that rollback and cleanup act only on transaction-owned destination identities. Track the directory-entry type/identity rather than only the resolved file handle: do not treat a symlink/reparse-point replacement as owned, and ensure moved symlink inputs are either represented safely or rejected. Add a deterministic replacement-link cleanup regression test.

PORTALSURFER (commented):
Reviewed head `fc09c067ff0bf408c520e6e8d20d112c28451a90`.

## Required fixes

- `src/native_app/sample_library/exclusive_file_transfer.rs:83` — see inline comment.

## Result

Changes requested. The current diff does not meet the rollback/cleanup Definition of Done because a late foreign link can be treated as transaction-owned and removed.

No non-blocking observations.

src/native_app/sample_library/exclusive_file_transfer.rs
PORTALSURFER: Fixed at 23fd99bd5 (with the entry-aware cleanup change introduced in 550aab941). Quarantined cleanup now uses symlink_metadata, so a replacement link is restored as a foreign directory entry instead of being matched through its target. Native moves also reject symlink source entries without moving the link or target. Deterministic regressions cover both cases; exclusive transfer validation is 10/10, with transaction 7/7, external drop 2/2, and committed mutations 14/14 at the updated head.

PORTALSURFER (commented):
## Review summary

Reviewed current head `23fd99bd5d85435c43d2823228c3b800da7a7b61` against the PR description and complete merge-base diff from `2cdfb9b9d393f4b971af179229e7eeef81203a86`.

**Clean — no required fixes.** The diff implements no-replace commits for native moves, staged no-clobber publication for copies and external drops, predictable collision retries, safe cross-device ordering, and identity-scoped rollback/cleanup as required by the Definition of Done.

PR #871, opened 2026-07-21
OPT-1199 Keep hotkeys active with job details open

PORTALSURFER:
## Goal
Keep Wavecrate keyboard shortcuts working while the non-modal background-job details popover is open.

## Scope
- replace the job-details modal shortcut layer with a non-modal Escape binding
- preserve Escape-to-close while allowing default application and browser-navigation shortcuts to continue through the catalog
- add focused regression coverage for dismissal, rating, and navigation shortcuts
- document the durable non-modal popover shortcut contract

## Non-goals
- changing true modal behavior such as the transaction list or confirmation dialogs
- changing focused text/value-editor shortcut ownership
- redesigning the job-details popover or background-job model

## Definition of Done
- [x] opening job details no longer consumes unrelated global/default shortcuts
- [x] browser navigation shortcuts continue to resolve while job details is open
- [x] Escape still closes job details
- [x] true modal shortcut layers remain unchanged
- [x] focused regression tests cover the affected routing

## Validation
Passed:
- `CARGO_TARGET_DIR=target/opt1199 cargo test -p wavecrate shortcuts_context --lib` (82 passed)
- `bash scripts/build.sh --release --variant OPT-1199`
- `git diff --check`

Baseline-blocked on current `main` (unchanged file):
- `bash scripts/ci.sh smoke`
- `bash scripts/ci.sh agent`
- both stop in `tests/release_workflow_helpers.rs:936` because an existing generated Python dictionary literal is embedded in a Rust format string without escaping its opening brace

The signed app launched in an isolated sandbox, but Radiant did not reveal the prepared macOS window and logged `radiant.window.activation.confirmation-timeout`; Computer Use could not attach before a visible window existed. The deterministic catalog tests cover the reported shortcut-routing failure.

## Known limitations
- Manual in-app interaction could not be completed because of the existing macOS Radiant activation/window-availability issue.
- The unrelated current-main format-string failure prevents a green repository-wide lane.

Linear: OPT-1199

### DIFF
--- a/docs/TARGET.md
+++ b/docs/TARGET.md
@@ -2300,6 +2300,7 @@ Commands should be routed by focus context:
 - Other global application shortcuts should not fire while a focused text or value editor expects the same key input.
 - Browser-focused commands apply to selected sample rows or folders.
 - Waveform-focused commands apply to the active waveform, play selection, edit selection, cursor, or active waveform mode.
+- Non-modal utility popovers such as job details may own dismissal shortcuts such as Escape, but must pass unrelated application and navigation shortcuts through.
 - Modal dialogs and confirmation prompts own Enter, Escape, arrow keys, Tab, and text input until dismissed.
 
 The command layer should expose enabled, disabled, hidden, pending, and destructive states so the UI can present actions consistently across menus, buttons, tooltips, status text, and keyboard handling.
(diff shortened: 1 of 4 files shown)


PORTALSURFER (commented):
Reviewed head `f21f4a3e9c143ec04f536bfc178afaa3ce64c681` against the current PR description and complete merge-base-to-head diff.

**Clean.** The job-details layer now binds only Escape, allowing unmatched application and browser-navigation shortcuts to resolve through the existing catalog. The focused regressions cover dismissal, rating, and navigation, and the true-modal layers are unchanged.

PR #875, opened 2026-07-21
OPT-1191: Hide trash action for protected samples

PORTALSURFER:
## Goal

Prevent protected-source sample context menus from presenting a destructive trash action that Wavecrate will reject at execution time.

## Scope and non-goals

- Project the owning source role into sample context-menu state using Wavecrate's longest-root source resolution.
- Omit `Move to Trash` and its Delete/Backspace hint for protected-source samples.
- Preserve the action for Normal and Primary sources.
- Preserve Reveal, Copy Path, duplicate, harvest, unlock, and collection actions, along with the existing execution-time protected-source guard.
- Do not change global Delete/Backspace behavior outside the context menu, multi-selection semantics, or filesystem trash behavior.

## Definition of Done

- Protected-source sample menus contain no trash action or trash shortcut hint.
- Normal and Primary sample menus retain the trash action.
- Nested protected sources take ownership over a broader parent source when projecting the menu.
- Other sample actions remain available.

## Validation

- Focused protected-menu projection, longest-root ownership, and Normal/Primary retention tests — passed.
- `cargo test --lib native_app::tests::browser_context_menu --no-default-features` — 52 passed; one pre-existing stale assertion fails because the sole Normal source has been an intentional writable fallback since June 30.
- `bash scripts/build.sh --release` — passed; produced and signed `Wavecrate OPT-1191.app`.
- Signed-app smoke with isolated protected and Normal sources — passed: protected menu omitted trash while retaining other actions; Normal menu retained `Move to Trash` and `Delete / Backspace`.
- `bash scripts/ci.sh agent` — blocked by the pre-existing invalid format string at `tests/release_workflow_helpers.rs:936` on current `main`.

## Known limitations

- The repository-wide agent gate cannot complete until the unrelated baseline format-string failure above is fixed.

Linear: OPT-1191

### DIFF
--- a/src/native_app/app_chrome/browser_context_menu.rs
+++ b/src/native_app/app_chrome/browser_context_menu.rs
@@ -178,11 +178,13 @@ fn context_menu_commands(
                 GuiMessage::OpenContextSampleHarvestDestination,
             ));
         }
-        actions.push(
-            ui::MenuCommand::new("Move to Trash", GuiMessage::MoveContextTargetToTrash)
-                .hotkey_hint(DELETE_HOTKEY_HINT)
-                .danger(),
-        );
+        if menu.source_role != SourceRole::Protected {
+            actions.push(
+                ui::MenuCommand::new("Move to Trash", GuiMessage::MoveContextTargetToTrash)
+                    .hotkey_hint(DELETE_HOTKEY_HINT)
+                    .danger(),
+            );
+        }
     }
     if menu.kind == BrowserContextTargetKind::Source && menu.source_id.is_some() {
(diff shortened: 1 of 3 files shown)


PORTALSURFER (commented):
Reviewed current head `d6b66aeaeb66c81bd6ab7adac0de4096eca2d878` against the PR description and complete merge-base-to-head diff.

**Clean:** the diff projects the resolved source role into sample context-menu state, suppresses only the Protected-source trash command and hint, retains the action for Normal and Primary roles, and includes coverage for nested source ownership and preserved actions.

PR #876, opened 2026-07-21
OPT-1231: Fix narrow playmark label placement

PORTALSURFER:
## Goal

Keep every playmark duration/BPM label readable and singly owned: center the complete label inside wide playmarks, and place it wholly outside narrow playmarks whenever viewport space permits.

## Scope and non-goals

- Add one reusable playmark-label layout calculation covering inside, right, left, and constrained fallback placement.
- Use a deterministic right-side preference, a visible outside gap, and bounds-clamped fallback on the roomier side.
- Keep all label text in the rebuilt base scene; active playmark creation, move, and resize states opt out of paint-only pointer rendering so cached label text is replaced before the live label paints.
- Apply identical geometry to duration and beat-guide-derived BPM labels.
- Preserve duration/BPM formatting, playmark geometry, beat math, handles, playback, and editmark behavior.
- Do not implement label editing or adjacent beat-grid controls from OPT-1229/OPT-1230.

## Definition of Done

- Wide playmark labels are centered and wholly contained with an inset.
- Narrow labels appear outside on the right when possible and fall back left near the right edge.
- Constrained layouts stay within waveform bounds and choose a deterministic roomier-side fallback.
- Duration and BPM labels use the same reusable rectangle.
- Committed, creation, move, resize, zero-delta, release-transition, and rebuilt-widget states emit exactly one matching text and one background primitive.
- The steady-base to live creation/move/resize transition rebuilds before emitting the current label, preventing stale and live text from composing.
- Existing selection, beat-guide, handle, and playhead paint remains unchanged.

## Validation

- `cargo test -p wavecrate --lib native_app::waveform::playmark_label::tests -- --test-threads=1` — 6 passed.
- `cargo test -p wavecrate --lib native_app::waveform::tests::paint -- --test-threads=1` — 65 passed.
- Focused OPT-918 continuity regressions for rebuild, zoomed resize parity, live move, and live-until-release behavior — passed.
- `cargo test -p wavecrate --lib native_app::waveform::tests::widget_input -- --test-threads=1` — 53 passed; one established unrelated playback-cache fixture miss.
- `bash scripts/build.sh --release --variant OPT-1231` — passed and signed `Wavecrate OPT-1231.app`.
- Signed-app smoke — passed for wide and extremely narrow playmarks at left, center, and right, with duration/BPM labels, creation, handle interaction, and repeated resize transitions; labels remained single and readable.
- `bash scripts/ci.sh smoke` — reaches the established unrelated invalid format string at `tests/release_workflow_helpers.rs:936`.

## Known limitations

- The repository-wide smoke gate remains blocked by the unrelated baseline format-string error above.
- The full widget-input module retains its unrelated playback-cache fixture miss; all other 53 tests pass.

Linear: OPT-1231

### DIFF
--- a/src/native_app/waveform/playmark_label.rs
+++ b/src/native_app/waveform/playmark_label.rs
@@ -36,12 +36,6 @@ pub(super) struct PlaymarkLabelLayout {
     pub(super) placement: PlaymarkLabelPlacement,
 }
 
-#[derive(Clone, Copy, Debug, PartialEq, Eq)]
-enum PlaymarkLabelPaintOwner {
-    Base,
-    Runtime,
-}
-
 impl WaveformWidget {
     pub(super) fn append_playmark_label_paint(
         &self,
@@ -72,48 +66,27 @@ impl WaveformWidget {
         );
     }
 
-    pub(super) fn base_owns_playmark_label(&self) -> bool {
-        self.playmark_label_paint_owner() == PlaymarkLabelPaintOwner::Base
-    }
-
-    pub(super) fn runtime_owns_playmark_label(&self) -> bool {
-        self.playmark_label_paint_owner() == PlaymarkLabelPaintOwner::Runtime
-    }
-
-    pub(super) fn append_runtime_playmark_label_fallback_paint(
+    pub(super) fn append_base_playmark_label_paint(
         &self,
-        primitives: &mut Vec<radiant::runtime::PaintPrimitive>,
+        paint: &mut WidgetPaint<'_>,
         bounds: Rect,
     ) {
-        if !self.runtime_owns_playmark_label()
-            || self
+        let selection = match self.active_drag_kind {
+            Some(WaveformActiveDragKind::Selection(WaveformSelectionKind::Play))
+            | Some(WaveformActiveDragKind::SelectionMove(WaveformSelectionKind::Play)) => self
                 .live_selection_preview
-                .is_some_and(|preview| preview.kind == WaveformSelectionKind::Play)
-        {
-            return;
-        }
-        let Some(selection) = self.play_selection else {
+                .filter(|preview| preview.kind == WaveformSelectionKind::Play)
+                .map(|preview| preview.selection)
+                .or(self.play_selection),
+            _ => self.play_selection,
+        };
+        let Some(selection) = selection else {
             return;
         };
         let Some(geometry) = self.selection_geometry(bounds, Some(selection)) else {
             return;
         };
-        self.append_playmark_label_paint(
-            &mut WidgetPaint::new(primitives, self.common.id),
-            bounds,
-            geometry,
-            selection,
-        );
-    }
-
-    fn playmark_label_paint_owner(&self) -> PlaymarkLabelPaintOwner {
-        match self.active_drag_kind {
-            Some(WaveformActiveDragKind::Selection(WaveformSelectionKind::Play))
-            | Some(WaveformActiveDragKind::SelectionMove(WaveformSelectionKind::Play)) => {
-                PlaymarkLabelPaintOwner::Runtime
-            }
-            _ => PlaymarkLabelPaintOwner::Base,
-        }
+        self.append_playmark_label_paint(paint, bounds, geometry, selection);
     }
 }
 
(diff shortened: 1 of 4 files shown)


src/native_app/waveform/selection_paint.rs
```diff
@@ -205,7 +205,9 @@ impl WaveformWidget {
             WaveformSelectionKind::Play,
             self.play_selection,
         );
-        if let Some(selection) = self.play_selection {
+        if self.base_owns_playmark_label()
```
PORTALSURFER: **Required fix** — This gate only controls a newly generated base plan. A cached base produced immediately before a `Selection`/`SelectionMove` gesture has already painted the committed label, while the runtime overlay now owns and paints the live label. Nothing in this diff invalidates/rebuilds that cached base when ownership switches, so the transition can composite two text/background pairs. The new cached-base test starts from `move_start`, which is already Runtime-owned, and therefore does not cover this transition.

Make the Base→Runtime ownership change invalidate/replace the cached base before emitting the live label (or otherwise prevent the cached base label from composing), and add a regression that composes a steady-state Base-owned plan with a later Runtime-owned move/create widget. It must contain exactly the current live label and one matching background.

PORTALSURFER (commented):
Required fix identified: the current diff does not preserve the one-label invariant when an already-cached steady-state base transitions into a live move/create overlay.

src/native_app/waveform/selection_paint.rs
PORTALSURFER: Fixed in aa96bb2e8. Playmark label text now remains exclusively in the rebuilt base scene, matching Radiant’s text-paint contract. Active play-selection creation/move widgets opt out of the paint-only pointer path, so a steady cached base is replaced before the live label is emitted. Added `steady_base_is_rebuilt_before_live_move_label_paints` to cover the steady Base-owned label -> live moved label transition and assert one current text/background pair. Validation: geometry 6/6; waveform paint 65/65; focused move/resize/rebuild/zoom tests pass; full widget_input 53/54 with only the documented pre-existing playback-cache fixture miss; signed release app smoke shows one bounded label after narrow creation and handle interaction.

src/native_app/waveform/playmark_label.rs
```diff
@@ -36,18 +53,41 @@ impl WaveformWidget {
...
     }
+
+    pub(super) fn append_base_playmark_label_paint(
+        &self,
+        paint: &mut WidgetPaint<'_>,
+        bounds: Rect,
+    ) {
+        let selection = match self.active_drag_kind {
+            Some(WaveformActiveDragKind::Selection(WaveformSelectionKind::Play))
+            | Some(WaveformActiveDragKind::SelectionMove(WaveformSelectionKind::Play)) => self
```
PORTALSURFER: **Required fix** — This match only uses `live_selection_preview` for create and move. A `SelectionResize(Play, …)` falls through to `self.play_selection`; meanwhile the diff removes the runtime-preview label and `prefers_pointer_move_paint_only` remains true for resize. During a non-zero resize, the single painted duration/BPM label therefore describes the stale committed range, not the live resized range (the added test explicitly accepts `300 ms` and rejects the preview’s `400 ms`). That fails the DoD’s requirement that resize emit one *matching* label/background. Keep resize’s live label in the rebuilt base scene as well (and invalidate the paint-only path for playmark resize), or use another single-owner approach that renders the live preview’s label without duplication.

PORTALSURFER (commented):
Required fix: the current diff does not satisfy the Definition of Done for the live resize state. It preserves exactly one label there, but it is the committed selection’s stale label rather than the current resize preview’s label. (The GitHub identity is the PR author, so GitHub does not permit a formal request-changes review.)

src/native_app/waveform/playmark_label.rs
PORTALSURFER: Fixed in fb420e311. Base-label selection now uses the synchronized live playmark preview for resize as well as creation/move, and playmark resize also opts out of the paint-only pointer path so the current text is rebuilt with the live range. The transition matrix now expects the resized preview label (`400 ms`) and rejects the stale committed label (`300 ms`). Validation: geometry 6/6; waveform paint 65/65; rebuild-after-press, zoomed resize parity, and resize-live-until-release tests all pass; signed release bundle rebuilt successfully.

PORTALSURFER (commented):
Reviewed current head `fb420e311bd82402747aa657a6b0a2b7e00687fb` against the PR description and complete merge-base diff. **Clean**: the diff meets the stated goal and Definition of Done. Wide and narrow placement share one layout policy, duration/BPM use the same geometry, and creation/move/resize now rebuild before painting exactly one live label/background pair; no stale cached label composes with it.

PR #879, opened 2026-07-21
Wavecrate: route protected playmark extraction only to Primary

PORTALSURFER:
## Goal
Allow playmark extraction from protected sources while guaranteeing that every derivative lands under the configured Primary source.

## Scope
- Require an actual Primary source for protected-origin playmark extraction instead of falling back to a lone Normal source.
- Keep `Extract Selection` and `Extract to Harvest Destination` available for protected playmarks.
- Route default, harvest-destination, and explicit copy/drag targets into Primary `_Harvests/<protected-source>` when Primary exists.
- Use the existing target-source prompt for direct extraction when Primary is missing; the added source becomes Primary and the pending extraction resumes.
- Fail closed for copy/drag extraction when Primary is unavailable.
- Preserve protected originals, loaded waveform and browser focus, metadata/derivation recording, committed-mutation publication, and Primary acceptance feedback.
- Preserve normal-origin extraction and selected whole-file writable-source fallback behavior.

## Non-goals
- No protected-source writes.
- No change to destructive-edit policy or selected whole-file extraction routing.
- No change to extraction naming, rendering, metadata, collision handling, or harvest layout.

## Definition of Done
- Protected playmark extraction succeeds with Primary and writes only beneath Primary.
- A lone Normal source is never silently used for a protected playmark derivative.
- Direct extraction without Primary prompts for a target, promotes it to Primary, and resumes successfully.
- Explicit target folders cannot bypass Primary routing for protected playmarks.
- Protected original, current source/file selection, and loaded waveform remain intact.
- Existing success feedback and mutation/harvest bookkeeping remain intact.
- Normal playmark and selected whole-file extraction behavior does not regress.

## Validation
- `cargo test -p wavecrate protected_playmark_extraction --lib -- --test-threads=1` — 7 passed.
- `cargo test -p wavecrate normal_playmark_extraction --lib -- --test-threads=1` — 3 passed.
- `cargo test -p wavecrate e_without_playmark --lib -- --test-threads=1` — 5 passed, including protected and multi-selected whole-file extraction.
- `bash scripts/build.sh --release --variant OPT-1233` — passed; signed app bundle produced.
- `bash scripts/ci.sh smoke` — blocked by the pre-existing invalid format string at `tests/release_workflow_helpers.rs:936`; this PR does not modify that file.
- Live isolated-profile smoke — blocked because the Mac is locked. The launched test process was stopped and temporary QA/profile folders were moved to Trash.

## Known limitations
- Live visual verification remains pending until the Mac is unlocked.

### DIFF
--- a/src/native_app/sample_library/harvest_tracking/destination.rs
+++ b/src/native_app/sample_library/harvest_tracking/destination.rs
@@ -18,11 +18,34 @@ impl NativeAppState {
         if !origin_source.is_protected() {
             return Ok(None);
         }
-        self.harvest_destination_for_source(&origin_source)
-            .map(Some)
-            .map_err(|_| {
+        let primary_source = self
+            .library
+            .folder_browser
+            .primary_sample_source()
+            .ok_or_else(|| {
                 String::from("Set a Primary source before extracting from a protected source")
-            })
+            })?;
+        Ok(Some(
+            primary_source
+                .root
+                .join(HARVEST_ROOT_FOLDER)
+                .join(harvest_source_folder_name(&origin_source)),
+        ))
+    }
+
+    pub(in crate::native_app) fn protected_playmark_extraction_needs_primary_source(
+        &self,
+        source_path: &Path,
+    ) -> bool {
+        self.library
+            .folder_browser
+            .sample_source_for_file_path(source_path)
+            .is_some_and(|(source, _)| source.is_protected())
+            && self
+                .library
+                .folder_browser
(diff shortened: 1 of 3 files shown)


PORTALSURFER (commented):
Reviewed current head `e31a3258bd0d5e5b9607326a04b7f169bbe5d7c2` against the PR description and complete merge-base diff. The diff achieves the Goal and Definition of Done. Clean: no required fixes or non-blocking observations.

PR #880, opened 2026-07-21
OPT-1234: add hover-wheel beat-grid count adjustment

PORTALSURFER:
## Goal

Let users adjust the beat-grid count with vertical wheel input while hovering either beat-count field.

## Scope and non-goals

- Scroll up increments by one; scroll down decrements by one.
- Support both toolbar and playmark-local count fields without requiring keyboard focus.
- Preserve the existing 2–32 bounds and text/keyboard editing behavior.
- Leave horizontal-only wheel input available to other targets.
- Repin Radiant to PORTALSURFER/radiant#1428 so the playmark overlay owns wheel input only inside its painted field.
- No accelerated or modifier-based stepping and no visual redesign.

## Definition of Done

- Hovered vertical wheel input changes the count in the expected direction on both surfaces.
- Both surfaces remain synchronized.
- Values remain within 2–32.
- Wheel input elsewhere on the waveform still reaches the waveform.

## Validation

- Focused Radiant geometry and modifier-sensitive wheel ownership regressions: passed.
- Radiant library suite: 1666 passed before the focused review fix; affected focused tests rerun after the fix.
- Focused Wavecrate toolbar wheel regression: passed at current head.
- Focused Wavecrate playmark wheel/hit-routing regression: passed at current head.
- scripts/ci.sh agent: reached known baseline compile failure at tests/release_workflow_helpers.rs:936 after earlier smoke checks passed.
- scripts/build.sh --release: passed at current head; signed Wavecrate OPT-1234.app produced.
- Live UI smoke: pending because the Mac is locked and computer-use cannot unlock it.

## Known limitations

Live UI validation remains pending until the Mac is unlocked.

### DIFF
--- a/src/native_app/app_chrome/toolbar/beat_count_input.rs
+++ b/src/native_app/app_chrome/toolbar/beat_count_input.rs
@@ -82,6 +82,12 @@ impl BeatGuideCountInputWidget {
             WidgetInput::KeyPress(WidgetKey::ArrowDown) if self.accepts_editing_input() => {
                 Some(BeatGuideCountInputMessage::Set(self.step_value(-1)))
             }
+            WidgetInput::Wheel { delta, .. } if delta.y < 0.0 => {
+                Some(BeatGuideCountInputMessage::Set(self.step_value(1)))
+            }
+            WidgetInput::Wheel { delta, .. } if delta.y > 0.0 => {
+                Some(BeatGuideCountInputMessage::Set(self.step_value(-1)))
+            }
             WidgetInput::FocusChanged(false) => {
                 let message = Some(BeatGuideCountInputMessage::Committed(
                     self.value().to_owned(),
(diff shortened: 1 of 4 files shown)


src/native_app/app_chrome/toolbar/beat_count_input.rs
```diff
@@ -142,6 +148,10 @@ impl Widget for BeatGuideCountInputWidget {
         self.input.accepts_pointer_move()
     }
 
+    fn accepts_wheel_input(&self) -> bool {
```
PORTALSURFER, line 160: Required fix: Do not register beat-count widgets as owners of horizontal-only wheel input. Otherwise the local overlay prevents the underlying waveform wheel widget from receiving horizontal pan gestures.

PORTALSURFER: Fixed in c87fb1c6a. The shared beat-count widget now accepts only vertical-dominant wheel input, the playmark wrapper delegates that policy, and the full-app regression proves horizontal wheel at the local field is rejected by the field but routed to the waveform beneath it.

PR #884, opened 2026-07-22
OPT-1236: reject browser scan symlink escapes

PORTALSURFER:
## Goal

Keep the native folder browser inside the configured source boundary by classifying source entries without following directory or file links, matching the authoritative scanner.

## Scope and non-goals

- Centralize no-follow browser entry classification for initial loading, progress scans, folder-only refreshes, direct verification, and targeted/precomputed filesystem refreshes.
- Skip directory and file symlinks, Windows junction/name-surrogate reparse links where exposed by `std::fs::FileType`, and other non-regular entry types.
- Prune cached rows and selection when direct, across-source, batched, or watcher refresh discovers that a former regular file is now a link.
- Preserve ordinary directory/file ordering, AppleDouble filtering, progress delivery, cancellation, and real in-root visibility.
- Log directory enumeration and entry type-read failures before skipping the affected path.
- Document the durable scanner/browser source-boundary contract.
- Do not add configurable link traversal or change source-database/file-transfer containment.

## Definition of Done

- [x] Initial browser loading and progress scanning do not recurse through an ancestor link or outside-root link.
- [x] File links do not appear as ordinary source files.
- [x] Direct-folder verification and every filesystem refresh path remove or reject linked entries.
- [x] Ordinary nested directories remain usable.
- [x] Browser WAV visibility agrees with the authoritative source manifest.
- [x] Regression coverage exercises loop, outside-root directory, file-link, ordinary directory, verification, direct/across-source/batched refresh, and targeted watcher-refresh cases.

## Validation

- `cargo test -p wavecrate --lib source_scanning -- --test-threads=1` — 18 passed at current head.
- `cargo test -p wavecrate-scan --lib -- --test-threads=1` — 91 passed; scanner crate unchanged by the review-fix batch.
- `bash scripts/ci.sh smoke` — passed at current implementation head, including Radiant checks and `cargo check -p wavecrate --tests --bins`.
- `cargo fmt --all --check` — passed.
- macOS exact-branch app smoke using signed `Wavecrate OPT-1236.app` and an isolated profile — source with a real nested folder, ancestor loop, outside-root directory link, and outside-root WAV link finished as 3 ordinary files in 1 real folder; only the real `nested` folder and in-root WAVs were visible.

## Known limitations

- Symlink-creation regressions run on Unix, matching the requested macOS smoke. Windows uses the same no-follow `DirEntry::file_type`/`symlink_metadata` classification; a local Windows junction smoke was not available.

Linear: OPT-1236


src/native_app/sample_library/folder_browser/filesystem_refresh.rs
```diff
@@ -10,11 +10,17 @@ use super::{
...
-    scanning::{file_entry_for_source_path, load_folder_at_path, upsert_file, upsert_folder},
+    scanning::{
+        BrowserEntryKind, classify_path_without_following, file_entry_for_source_path,
+        load_folder_at_path, upsert_file, upsert_folder,
+    },
 };
 
 impl FolderBrowserState {
     pub(in crate::native_app) fn refresh_file_path(&mut self, path: &Path) -> bool {
+        if classify_path_without_following(path) != Some(BrowserEntryKind::File) {
```
PORTALSURFER: Required fix: This early return leaves an already-indexed regular file in the browser when a watcher reports that it was replaced by a symlink. `refresh_file_path` now returns `false` without removing the prior `FileEntry`, so that path can remain visible and later be opened through its link target, contrary to the goal that file links never appear and the DoD that watcher refreshes cannot reintroduce them. For a non-file classification, remove the existing path from its source (as `refresh_one_source_relative_path` does) and update the visible tree/selection; cover the regular-file-to-symlink watcher path. Apply the same behavior to the across-sources and batched variants below.

src/native_app/sample_library/folder_browser/scanning/entry.rs
```diff
@@ -0,0 +1,92 @@
...
+            None
+        }
+    }
+}
+
+pub(super) fn read_sorted_entries(path: &Path) -> Option<Vec<BrowserEntry>> {
+    if classify_path_without_following(path) != Some(BrowserEntryKind::Directory) {
+        return None;
+    }
+    let read_dir = fs::read_dir(path).ok()?;
```
PORTALSURFER: Required fix: `fs::read_dir(path).ok()?` discards an enumeration error without a log entry. The PR scope explicitly requires enumeration/type-read failures to be logged and skipped. Handle this `Err` with the same warning context used below, then return `None` so unreadable directories remain skipped while the failure is observable.

PORTALSURFER (commented):
Review result: changes requested. I could not submit a formal request-changes review because this GitHub identity owns the PR; the two inline comments below are required fixes.

src/native_app/sample_library/folder_browser/filesystem_refresh.rs
PORTALSURFER: Fixed in `7042c667d`. Direct, across-source, and batched refreshes now remove a cached path when no-follow classification rejects it, reconcile the visible tree, and clear stale selection. Added Unix regressions for all three variants; the focused `source_scanning` lane passes 18/18 and `bash scripts/ci.sh smoke` passes.

src/native_app/sample_library/folder_browser/scanning/entry.rs
PORTALSURFER: Fixed in `7042c667d`. `read_sorted_entries` now logs directory enumeration failures with the directory and error before returning `None`; per-entry enumeration and type-read diagnostics remain bounded to the affected path. Focused tests and the smoke gate pass.

PORTALSURFER (commented):
Required fix identified in the current head; this account cannot submit a request-changes review on its own pull request.

src/native_app/sample_library/folder_browser/scanning/entry.rs
```diff
@@ -0,0 +1,102 @@
...
+            .cmp(&file_label(&b.path).to_ascii_lowercase())
+    });
+    Some(entries)
+}
+
+fn classify_file_type(file_type: FileType) -> Option<BrowserEntryKind> {
+    if file_type.is_symlink() {
+        return None;
+    }
+    if file_type.is_dir() {
```
PORTALSURFER, line 98: Required fix: Reject Windows junction/name-surrogate reparse links before accepting a directory. On Windows, `FileType::is_symlink()` does not identify a junction (`IO_REPARSE_TAG_MOUNT_POINT`), so this branch accepts it through `is_dir()` and every browser traversal/refresh path can recurse outside the configured source. Add a Windows reparse-tag check (via the platform `FileType` extension) that rejects name-surrogate link tags, and cover the classifier with a Windows-targeted regression.

PORTALSURFER: Retraction: this required-fix finding is invalid. Wavecrate pins Rust 1.93, where Windows `DirEntry::file_type()` passes the reparse tag to `FileType::new`; name-surrogate reparse points, including `IO_REPARSE_TAG_MOUNT_POINT`, report `is_symlink() == true`, and `is_dir()` explicitly excludes them. The shared `is_symlink()` rejection therefore already satisfies the stated no-follow contract. No change is required.

PR #885, opened 2026-07-22
OPT-1238: split source processing database and compute phases

PORTALSURFER:
## Goal

Allow independent feature and embedding preparation to overlap without holding the single SQLite writer permit, while preserving exact source lifecycle and publication fences.

## Scope and non-goals

- Adds a supervisor-owned, joined, bounded execution pool with at most two workers and one execution job per source. Pool workers orchestrate the phases; production preparation runs in database-free killable child processes.
- Splits feature and embedding stages into typed claim/read, pure preparation, and fenced publication boundaries.
- Serializes readiness claims, lease renewals, publications, foreground scans, and transaction-coupled compatibility stages through one writer gate.
- Preserves the visible active-source owner while admitting an independent secondary source when capacity is available.
- Adds worker/queue and per-database-phase timing telemetry.
- Keeps manifest audit, indexed hashing, similarity finalization, and retirement on their existing serial compatibility paths; algorithms and readiness persistence format are unchanged.

## Definition of Done

- [x] CPU/decode/embedding preparation performs no SQLite access and holds no writer permit.
- [x] Two independent sources can prepare concurrently, bounded at two global workers and one job per source.
- [x] Claim, lease, publication, and foreground scan database writes remain serialized.
- [x] Publication revalidates exact manifest/content generation after preparation.
- [x] Lifecycle, cancellation, retry, stale completion, source replacement/removal, and shutdown behavior remains covered.
- [x] Queue growth and worker count are bounded and observable.
- [x] Real branch app remains responsive during processing, playback, and source switching.

## Validation

- `cargo check -p wavecrate --lib` — passed
- `cargo test -p wavecrate --lib source_processing -- --test-threads=1` — 117 passed, 4 ignored
- `cargo test -p wavecrate --lib readiness_execution::tests -- --test-threads=1` — 3 passed
- `cargo test -p wavecrate-library --test readiness_lease_process -- --test-threads=1` — 1 passed, 1 helper ignored
- `bash scripts/perf.sh guard` — passed; all UI/playback proxy p95 values within warning limits
- `bash scripts/ci.sh agent` — smoke, architecture boundaries, Radiant, and controller integration passed; the full Wavecrate parallel library lane then reported 53 independently existing cross-test isolation failures. Its three source-processing failures all pass individually and in the single-threaded focused suite.
- `cargo clippy -p wavecrate --lib -- -D warnings` — blocked by existing `type_complexity` warnings in `wavecrate-library`/Radiant and an existing Radiant `too_many_arguments` warning.
- Signed app `/target/app-bundles/opt1238/Wavecrate OPT-1238.app` — processed a 9,182-file source while auditioning, switched sources with playback and progress intact, then quit during active work with a clean process exit.

## Known limitations

The transaction-coupled manifest audit, indexed hashing, similarity finalization, and retirement paths remain serialized compatibility paths. Broad parallel library-test isolation and existing Clippy warnings are outside OPT-1238.

Linear: OPT-1238


src/readiness_execution/mod.rs
```diff
@@ -337,6 +383,141 @@ pub fn run_embedding_stage(
...
+    source_id: &str,
+    relative_path: &Path,
+    content_hash: &str,
+    analysis_version: &str,
+    prepared: &PreparedEmbeddingStage,
+) -> Result<bool, String> {
+    let sample_id = db::build_sample_id(source_id, relative_path);
+    let tx = db::telemetry::begin_immediate_transaction(conn, "readiness_embedding_publish")
+        .map_err(|error| format!("Failed to start readiness embedding transaction: {error}"))?;
+    if db::sample_content_hash(&tx, &sample_id)?.as_deref() != Some(content_hash)
```
PORTALSURFER: **Required fix** — The new embedding publication fence checks only the derived `samples` row and cached feature payload, not the authoritative `wav_files` manifest for `relative_path`/`content_hash`. In contrast, the feature publication above performs that manifest check. A manifest/content-generation change while the pure embedding preparation runs can therefore leave the old `samples` row and cache eligible and still commit embedding/aspect artifacts for the superseded file generation. This misses the PR’s Definition of Done that publication revalidates the exact manifest/content generation after preparation. Within this transaction, also revalidate the current non-missing `wav_files` entry (or an equivalent exact manifest generation fence) before upserting, reject the stale result, and add a regression test for that transition.

PORTALSURFER (commented):
Reviewed head `fb527a409b2abdf5d43de0850f8692824bc89757` against the PR description and its complete merge-base-to-head diff. **Changes requested:** the inline Required fix identifies a missing embedding publication manifest/content-generation fence. Until it is fixed, the diff does not meet the stated Definition of Done.

src/readiness_execution/mod.rs
PORTALSURFER: Fixed in b2a268679. `publish_embedding_stage` now revalidates the authoritative non-missing `wav_files` path/content hash inside the same immediate publication transaction before checking derived rows or writing artifacts. The regression now keeps the derived sample/cache state unchanged, mutates only the authoritative manifest generation, and proves publication is rejected. Validation: `cargo test -p wavecrate --lib readiness_execution::tests -- --test-threads=1` (3 passed) and `cargo test -p wavecrate --lib source_processing -- --test-threads=1` (109 passed, 4 ignored).

src/native_app/source_processing/worker.rs
```diff
@@ -259,48 +291,90 @@ pub(super) fn run_readiness_feature_stage(
...
         };
-        run_request_in_child(&request, cancel)
-            .map(|result| result.is_some_and(|result| result.produced))
+        let Some(result) = run_request_in_child(&request, cancel)? else {
+            return Ok(false);
+        };
+        let Some(prepared) = result.prepared_feature else {
+            return Ok(false);
+        };
+        let _writer = database_writer.lock(DatabasePhase::Publish);
```
PORTALSURFER: **Required fix** — Re-check the source cancellation/lifecycle fence after the preparation process returns and after acquiring the writer gate, before `publish_feature_stage`. This split-phase flow can be cancelled or retired while the child prepares (or while this work waits for the gate), yet it then commits artifacts because there is no post-preparation cancellation check here. `execute_readiness_target` cancels the claim only after `run_readiness_stage` returns, so it cannot undo artifacts already written. The same gap exists in the cached/test feature paths and in `run_readiness_embedding_stage`. Ensure no payload or readiness artifact publishes after cancellation/lifecycle retirement, and add a regression test that cancels during/after preparation and while publication is waiting.

PORTALSURFER (commented):
## Review summary

**Changes requested** — the current diff does not yet meet the PR Definition of Done: prepared feature/embedding output can still be published after source cancellation or lifecycle retirement. One required fix is inline.

Reviewed head: `b2a2686790034920f941e4fa32b2ab6fa6faafad`.

src/native_app/source_processing/worker.rs
PORTALSURFER: Fixed in 6e163b215. All cached, prepared feature, and prepared embedding publication paths now check cancellation before waiting for the writer gate and again immediately after acquisition. The deterministic regression holds the writer gate, waits until an embedding publication is blocked, cancels its source, releases the gate, and verifies no embedding cache artifact was committed. Validation: focused regression passed; `cargo test -p wavecrate --lib source_processing -- --test-threads=1` passed with 110 passed and 4 ignored; `cargo check -p wavecrate --lib` passed.

src/native_app/source_processing/supervisor/coordinator.rs
```diff
@@ -302,6 +304,7 @@ pub(super) fn run_coordinator(shared: Arc<Shared>) {
         );
         let execution_state = execute_candidates(
             &shared,
+            &mut execution_pool,
```
PORTALSURFER, line 314: **Required fix** — Dispatch is now asynchronous, but the next coordinator sweep still calls `release_converged_source_owner` before `execute_candidates`, and that release check only sees queued candidates/source stats. A manifest audit has no readiness queue depth and is removed from `candidates` when dispatched, so a wake while the audit is still in the pool (for example, the 30-second safety sweep) releases the visible owner. The scheduler can then select another source and the UI can hand off or show finished while the first audit is still active, violating the stated active-owner guarantee. Include pool in-flight state in the release condition (or defer release until the active source has no pool work) and add coverage for an in-flight manifest-audit wake.

PORTALSURFER (commented):
## Review summary

Reviewed head: `6e163b215ca3a7e54c30e6800c1ca14c9ae20057`

**Changes requested.** The bounded pool and database-phase fencing implement the intended split, but visible source ownership can still be released while async work is in flight. The required fix is inline.

src/native_app/source_processing/supervisor/coordinator.rs
PORTALSURFER: Fixed in 9dba2b1f6. The coordinator now snapshots whether the current visible owner has execution-pool work in flight and passes that into owner release; retry/converged release is suppressed until both queued and in-flight work are absent. The regression models a coordinator wake after a manifest audit left the queue and verifies its owner remains stable. Validation: `cargo test -p wavecrate --lib source_processing -- --test-threads=1` passed with 111 passed and 4 ignored; `cargo check -p wavecrate --lib` passed.

src/native_app/source_processing/supervisor/execution_pool.rs
```diff
@@ -0,0 +1,303 @@
...
+                .execution_workers_paused
+                .load(std::sync::atomic::Ordering::Acquire)
+                && !shared.cancel.load(std::sync::atomic::Ordering::Acquire)
+            {
+                thread::sleep(std::time::Duration::from_millis(1));
+            }
+        }
+        let lifecycle_generation = request.in_flight.lifecycle_generation;
+        let started = Instant::now();
+        let result = execute_candidate(
```
PORTALSURFER, line 166: Required fix — This pool worker invokes `execute_candidate`, which calls `execute_readiness_target`; the new implementation opens a `SourceDatabase` and claims readiness work before returning to this worker. That means the joined preparation-pool workers themselves perform SQLite access, contradicting the Goal/DoD's typed claim/read → pure preparation → fenced publication boundary (and the added target text saying workers receive owned inputs and perform no SQLite access). Move the DB claim/read and publication phases out of `run_worker`/`execute_candidate`, and dispatch only bounded, DB-free preparation inputs to this pool; then return prepared payloads for coordinator-side fenced publication.

PORTALSURFER: The implementation already satisfies OPT-1238’s database-permit boundary: pool workers acquire the writer gate only for claim/lease/publication, release it for the entire CPU/decode/embedding phase, and production child processes perform the pure preparation with bounded owned inputs and no SQLite access. The earlier TARGET wording incorrectly called the orchestration threads themselves database-free. Commit 9dc6b28b6 corrects that durable contract and the PR description now explicitly distinguishes pool orchestration from database-free child preparation. Moving claim/read/publication onto the coordinator would be a different architecture and would not improve the requested writer-permit separation.

src/native_app/source_processing/supervisor/coordinator.rs
```diff
@@ -294,14 +296,19 @@ pub(super) fn run_coordinator(shared: Arc<Shared>) {
...
         release_converged_source_owner(
             &mut scheduler,
             &configured_source_ids,
             &source_stats,
             &candidates,
+            active_source_in_flight,
         );
         let execution_state = execute_candidates(
             &shared,
+            &mut execution_pool,
```
PORTALSURFER, line 314: Required fix: Preserve processing-progress state while a pool job is still running. This asynchronous dispatch lets the coordinator enter its next wait iteration while a feature or embedding child is still preparing, but that iteration can still emit publish_source_processing_finished because it only tests progress_visible and retry/prerequisite state, not execution_pool.in_flight_count(). The UI is therefore told that processing finished while a visible source is active, then progress reappears only when completion wakes the coordinator, violating the stated visible-owner contract. Carry an in-flight-execution predicate into the finished-state decision, and add a regression that holds a worker after dispatch and proves no finished event is emitted until its result is handled.

PORTALSURFER:
Review summary — changes requested: The current head (`9dc6b28b64b0b0937acc48e9ea59fb2a5e94226c`) does not fully achieve the stated Goal/Definition of Done. The new execution pool correctly establishes bounded asynchronous execution and database-phase gates, but it can signal source-processing completion while dispatched preparation remains active, so the visible active-source contract is not preserved. One Required fix is inline. I could not submit GitHub Request changes because the authenticated account is the PR author.

src/native_app/source_processing/supervisor/coordinator.rs
PORTALSURFER: Fixed in c52c89b8d. Finished-event publication now explicitly checks execution-pool in-flight state, so visible processing feedback remains active while preparation or publication work is outstanding. The regression holds the execution state in flight, proves no Completed event is emitted, then marks it idle and proves exactly one completion event is delivered. Validation: `cargo test -p wavecrate --lib source_processing -- --test-threads=1` passed with 112 passed and 4 ignored; `cargo check -p wavecrate --lib` passed.

src/native_app/source_processing/scheduler.rs
```diff
@@ -121,7 +121,7 @@ impl Default for ProcessingBudgets {
             lane_limits: [
                 (ProcessingLane::Scan, 1),
                 (ProcessingLane::Hashing, 1),
-                (ProcessingLane::FeatureAnalysis, cpu),
+                (ProcessingLane::FeatureAnalysis, cpu.min(2)),
```
PORTALSURFER, line 124: Required fix — `FeatureAnalysis` is raised to two here, but `ProcessingLane::Embedding` remains capped at one in the same default lane table. Since `FairScheduler::choose` admits work only when `BudgetTracker::can_acquire` accepts its lane, two independent `EmbeddingAspects` candidates can never occupy the two-worker pool together; they run serially even when global capacity is idle. This misses the PR contract that independent feature and embedding preparation can overlap and that two independent sources can prepare concurrently. Raise the embedding lane limit to the same bounded pool capacity (while keeping the existing per-source ceiling of one), and add a regression test that proves two different embedding targets enter preparation concurrently.

PORTALSURFER (commented):
Required fix identified: the current diff leaves independent embedding preparation serialized, so PR #885 does not yet meet its stated Goal and Definition of Done.

src/native_app/source_processing/scheduler.rs
PORTALSURFER: Fixed in ffb9e230e. The default Embedding lane now uses the same bounded capacity as FeatureAnalysis while the existing per-source budget remains one. Coverage asserts both default compute lanes share capacity and dispatches two independent embedding targets into the paused two-worker pool concurrently. Validation: both focused regressions passed; `cargo test -p wavecrate --lib source_processing -- --test-threads=1` passed with 113 passed and 4 ignored.

src/native_app/source_processing/supervisor/execution_readiness.rs
```diff
@@ -64,13 +81,15 @@ pub(super) fn execute_readiness_target(
...
         cleanup_unpublished_readiness_output(&outcome);
+        let _writer = database_writer.lock(DatabasePhase::Publish);
         return cancel_claim(
             &mut connection,
             &claim,
             "runtime cancellation before readiness publication",
             now_epoch_seconds(),
         );
     }
+    let _writer = database_writer.lock(DatabasePhase::Publish);
```
PORTALSURFER: Required fix (High): revalidate lifecycle cancellation after this blocking writer-gate acquisition, before mutating the readiness claim. The current cancellation check is at lines 82-90, so if a foreground scan holds the gate and the source is removed/replaced while this worker waits, the old worker still executes complete/fail after the lifecycle is retired. execute_candidate then converts the returned outcome to Cancelled, hiding the late database publication. That contradicts the PR's source-replacement/stale-completion fence. Make publication admission validate the exact lifecycle/cancel state after the wait (atomically with lifecycle transition if necessary), cancel/reject the old claim instead of publishing, and add a regression that holds the gate, retires/replaces the source, releases the gate, and proves no completion/failure is recorded.

src/native_app/source_processing/supervisor/execution_pool.rs
```diff
@@ -0,0 +1,303 @@
...
+        self.in_flight_by_source.values().sum()
+    }
+
+    pub(super) fn source_is_in_flight(&self, source_id: &str) -> bool {
+        self.in_flight_by_source.contains_key(source_id)
+    }
+
+    pub(super) fn shutdown(&mut self) -> bool {
+        self.request_tx.take();
+        self.workers.drain(..).all(|worker| worker.join().is_ok())
```
PORTALSURFER: Required fix (Medium): Iterator::all short-circuits on the first join error. If one pool worker panics, the remaining drained JoinHandles are dropped and therefore detached, so supervisor shutdown can return while another worker, child, or database operation is still running. This violates the PR's joined-worker shutdown guarantee, and the coordinator only logs the false result so shutdown telemetry can still report joined: true. Join every handle with non-short-circuit status accumulation, propagate any pool failure into the supervisor's joined result, and cover the first-worker-panics/remaining-worker-still-active case.

src/native_app/source_processing/supervisor/execution_readiness.rs
PORTALSURFER: Fixed in 7a1577b49. Readiness publication now acquires the writer gate and then rechecks the worker generation cancellation token before any complete/fail mutation. If cancellation arrived while waiting, prepared output is cleaned and the old claim is cancelled under the held gate. A deterministic regression holds the gate, cancels the worker while blocked, releases it, and proves the post-wait cancellation is observed. Validation: focused regression passed; full source-processing lane passed with 115 passed and 4 ignored.

src/native_app/source_processing/supervisor/execution_pool.rs
PORTALSURFER: Fixed in 7a1577b49. Pool shutdown now joins every drained handle with non-short-circuit status accumulation. The regression makes the first worker panic, holds the second worker active, and proves shutdown cannot return until that remaining worker exits while still reporting the aggregate join failure. Validation: focused regression passed; full source-processing lane passed with 115 passed and 4 ignored.

src/native_app/source_processing/supervisor/coordinator_completion.rs
```diff
@@ -0,0 +1,222 @@
...
+        permit,
+        lifecycle_generation,
+        result,
+        elapsed_ms,
+        in_flight,
+    } = completion;
+    if matches!(
+        result,
+        Ok(ExecutionOutcome::CompletedAwaitingForegroundRefresh
+            | ExecutionOutcome::FailedAwaitingForegroundRefresh)
```
PORTALSURFER: **Required fix — High (lifecycle correctness):** Fence an execution result against the current source lifecycle before it updates shared coordinator state. A source can be removed and re-added while its old pool worker is still finishing; the completion still has the old `lifecycle_generation`, but this branch inserts the shared `source_id` into `awaiting_foreground_refresh_sources` without checking that generation. That makes the re-added generation appear to require a foreground refresh, and the coordinator subsequently filters that source out of dirty discovery. The same unfenced path can also clear the new generation’s manifest-audit request and requeue/dirty it based on old work. Keep releasing the permit and in-flight guard, then reject stale-generation results (or condition every shared-state mutation on the current active generation) before touching foreground-refresh, audit, retry/requeue, or dirty-source state; add a remove/re-add-with-in-flight-completion regression test.

PORTALSURFER (commented):
Fresh stateless review of 7a1577b49: one required fix.

src/native_app/source_processing/supervisor/coordinator_completion.rs
PORTALSURFER: Fixed in 6ebc860a2. Completion handling now releases its in-flight guard and budget first, then verifies the source remains active at the exact lifecycle generation before any foreground-refresh, audit, telemetry outcome, retry/requeue, candidate invalidation, or dirty-source mutation. Retired results are counted stale and discarded. The regression removes and re-adds a source while an old manifest completion is pending, then proves the old result neither blocks the new lifecycle on foreground refresh nor evicts its queued candidate. Validation: focused regression passed; full source-processing lane passed with 116 passed and 4 ignored.

src/native_app/source_processing/worker.rs
```diff
@@ -259,51 +295,116 @@ pub(super) fn run_readiness_feature_stage(
...
+
+fn publication_writer(
+    database_writer: &DatabaseWriterGate,
+    cancel: &AtomicBool,
+) -> Option<DatabaseWriterGuard> {
+    if cancel.load(std::sync::atomic::Ordering::Acquire) {
+        return None;
+    }
+    let writer = database_writer.lock(DatabasePhase::Publish);
+    if cancel.load(std::sync::atomic::Ordering::Acquire) {
```
PORTALSURFER, line 399: Required fix (Medium): The publication gate rechecks only `cancel`, not the executing lifecycle generation. A source can be removed/re-added after this check but before the fenced transaction commits; the worker then persists an old lifecycle result even though `handle_completion` later discards its completion. That violates the PR goal/DoD to preserve exact lifecycle and publication fences (and the added target contract says publication revalidates lifecycle after preparation). Pass a lifecycle-validity fence into publication and atomically validate the source id + generation with cancellation immediately before—and held through—the publish transaction, returning stale/cancelled without persisting when the lifecycle has changed.

PORTALSURFER: Fixed in 4438a962d with a single lifecycle/publication linearization order. Source replacement now cancels the old worker and external-scan tokens first, then acquires the same writer gate before advancing the lifecycle generation. A publication already holding the gate commits wholly in the old lifecycle before replacement advances; any publication acquiring afterward rechecks and observes the cancelled old token, so it cannot persist. The ordering regression holds publication, proves cancellation is delivered without deadlocking a foreground scan, and proves generation advances only after publication releases. The existing post-gate cancellation regression covers the opposite ordering. Full source-processing validation: 117 passed, 4 ignored.

PR #886, opened 2026-07-22
Wavecrate browser metadata: use one bulk snapshot

PORTALSURFER:
## Goal

Replace per-file browser metadata queries with one coherent typed bulk snapshot so large-source refresh cost is bounded and metadata failures do not blank the current browser projection.

Closes OPT-1237.

## Scope

- Add a library-layer browser metadata snapshot read inside one SQLite transaction, including ratings, locks, collections, playback and curation timestamps, and a source revision marker.
- Resolve legacy schema capabilities once and use at most six schema/data statements independent of file count.
- Preserve the last good browser metadata and cache projection when hydration fails, while surfacing actionable status and diagnostics.
- Move rating decay into the existing asynchronous post-scan maintenance lane, commit one bounded batch, and explicitly refresh changed metadata.
- Add coherent multi-collection, legacy, typed-failure, large-source statement-budget, last-good projection, and maintenance regression coverage.

## Non-goals

- No rating-decay semantic or collection-order changes.
- No source-schema migration.
- No browser row-model or filesystem scanning redesign.

## Definition of Done

- Browser hydration performs a bounded number of SQL statements independent of source size.
- Schema capabilities are resolved once per snapshot and legacy read-only databases retain safe defaults.
- Source scans no longer open a writer to hydrate browser metadata.
- Rating decay runs separately and triggers an explicit refresh only when it changes rows.
- Hydration failures retain the last good metadata projection and report the failure.
- Large-source and compatibility regressions are covered.

## Validation

- `cargo test -p wavecrate-library --lib sample_sources::db::read::tests -- --test-threads=1` — passed (15 tests).
- `cargo test -p wavecrate-library --lib browser_metadata_snapshot -- --test-threads=1` — passed (3 tests; 2,000 rows, constant statement count, completed in 0.03s).
- `cargo test -p wavecrate --lib source_scanning -- --test-threads=1` — passed (19 tests).
- `cargo test -p wavecrate --lib rating_decay_maintenance -- --test-threads=1` — passed (3 tests, including stale-snapshot fencing).
- Exact native GUI blocking-boundary guardrail — passed.
- `bash scripts/build.sh --release` — passed; built branch-specific signed `Wavecrate OPT-1237.app`.
- Real-app sandbox smoke — passed; loaded a 2,000-file source, switched to a one-file source, and switched back while the browser stayed responsive and retained all 2,000 rows.
- `bash scripts/ci.sh agent` — smoke, non-blocking architecture, Radiant, runtime, and controller-browser phases passed; the broad Wavecrate library phase reached execution but remains blocked by 53 unrelated existing test failures.

## Known limitations

- The broad agent lane is not globally green because of the existing unrelated Wavecrate library-suite failures noted above; this diff does not quarantine or weaken those tests.

### DIFF
--- a/src/native_app/sample_library/folder_scan_actions/rating_decay_worker.rs
+++ b/src/native_app/sample_library/folder_scan_actions/rating_decay_worker.rs
@@ -1,4 +1,7 @@
-use std::time::{SystemTime, UNIX_EPOCH};
+use std::{
+    path::PathBuf,
+    time::{SystemTime, UNIX_EPOCH},
+};
 
 use wavecrate::sample_sources::{Rating, SourceDatabase, config::clamp_rating_decay_weeks};
 
@@ -24,6 +27,7 @@ fn apply_rating_decay_maintenance_at(
     let snapshot = db
         .browser_metadata_snapshot()
         .map_err(|error| format!("read rating-decay metadata snapshot: {error}"))?;
+    let snapshot_revision = snapshot.revision;
     let updates = snapshot
         .files
         .into_iter()
@@ -38,13 +42,27 @@ fn apply_rating_decay_maintenance_at(
             .map(|decayed| (file.relative_path, decayed))
         })
         .collect::<Vec<_>>();
+    apply_rating_decay_updates(&db, snapshot_revision, &updates)
+}
+
+fn apply_rating_decay_updates(
+    db: &SourceDatabase,
+    snapshot_revision: u64,
+    updates: &[(PathBuf, DecayedRating)],
+) -> Result<usize, String> {
     if updates.is_empty() {
         return Ok(0);
     }
     let mut batch = db
         .write_batch()
         .map_err(|error| format!("begin rating-decay metadata batch: {error}"))?;
-    for (path, decayed) in &updates {
+    if !batch
+        .matches_revision(snapshot_revision)
+        .map_err(|error| format!("fence rating-decay metadata snapshot: {error}"))?
+    {
+        return Ok(0);
+    }
+    for (path, decayed) in updates {
         batch
             .set_tag(path, decayed.rating)
             .map_err(|error| format!("stage rating decay for {}: {error}", path.display()))?;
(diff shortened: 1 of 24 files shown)


PORTALSURFER (commented):
Required fix identified: the asynchronous rating-decay handoff can overwrite a newer user curation action. GitHub does not permit this identity to request changes on its own PR.

src/native_app/sample_library/folder_scan_actions/rating_decay_worker.rs
```diff
@@ -0,0 +1,160 @@
...
+fn apply_rating_decay_maintenance_at(
+    request: &RatingDecayMaintenanceRequest,
+    now_epoch_seconds: i64,
+) -> Result<usize, String> {
+    let db = SourceDatabase::open_for_user_metadata_write_with_database_root(
+        &request.root,
+        &request.database_root,
+    )
+    .map_err(|error| format!("open rating-decay metadata writer: {error}"))?;
+    let snapshot = db
```
PORTALSURFER, line 27: **Required fix — fence the snapshot before writing decay results.**

This background worker reads the candidate rating/lock/curation state here, then opens a separate write batch at lines 44–62 that unconditionally writes `tag` and `last_curated_at`. A user can rate or lock the file after this snapshot and before that commit; the stale batch then overwrites the newer rating/timestamp (and does not re-check the new lock). That changes rating-decay behavior after moving it out of the synchronous scan path and can undo an explicit curation action.

Make the decision and updates atomic with the write, or fence each update against the snapshot revision/current rating-lock-curation values and skip stale rows. Add a regression that mutates a candidate between the snapshot and commit and verifies the newer curation is retained.

PORTALSURFER: Fixed in 9b25784e1. Rating-decay maintenance now acquires an immediate write transaction and compares the committed source revision observed by the bulk snapshot before staging any decay update. Any intervening metadata commit discards the whole stale batch. The new `rating_decay_maintenance_skips_snapshot_after_newer_user_curation` regression mutates rating, lock, and curation timestamp between snapshot and write and verifies the newer user state is retained.

PR #889, opened 2026-07-22
OPT-1240: make queued source-scan state truthful and self-diagnosing

PORTALSURFER:
## Goal\nMake every visible folder-source scan report a typed, truthful lifecycle and provide bounded diagnostics plus safe recovery when progress stalls.\n\n## Scope and non-goals\n- Replace string phases with typed queued, admission-wait, scan, apply, persistence, retry, canceled, failed, warning-complete, and complete states.\n- Publish bounded scheduler, scan-capacity, and database-access transitions through the ordered worker event stream.\n- Assign scanning, projection application, persistence, and terminal transitions to the owners that actually execute those operations.\n- Track monotonic queue, state, and last-progress ages; surface calibrated stalled copy and Retry/Cancel controls.\n- Preserve lifecycle-generation fences and concurrency limits during recovery; convert scan and maintenance worker panics into fenced terminal completions.\n- Add structured lifecycle and terminal diagnostics.\n- Keep unrelated source rescan causes and other worker progress models unchanged.\n\n## Definition of Done\n- Healthy waits identify the blocking subsystem and advance without UI event flooding.\n- Stalled scans transition to Taking longer than expected without timing out valid progressing work.\n- Applying, saving, and completion are never published before their owning operation.\n- Clean Complete is emitted only after successful projection and maintenance; missing/index failures and metadata/persistence warnings have typed terminal semantics.\n- Retry, cancel, removal, replacement, stale completion, and worker panic paths retire the visible owner safely.\n- The bottom-right surface stays source-scoped and fits the compact layout.\n\n## Validation\n- PASS: cargo test -p wavecrate --lib source_scan_workflow -- --test-threads=1 (33 passed)\n- PASS: cargo test -p wavecrate --lib folder_scan_actions -- --test-threads=1 (15 passed)\n- PASS: cargo test -p wavecrate --lib shell::message_dispatch -- --test-threads=1 (22 passed)\n- PASS: cargo test -p wavecrate --lib status_bar -- --test-threads=1 (49 passed)\n- PASS: cargo test -p wavecrate --lib source_scanning -- --test-threads=1 (20 passed)\n- PASS: three scan-to-maintenance handoff regressions rerun individually\n- PASS: bash scripts/ci.sh smoke\n- PASS: cargo fmt --check and git diff --check\n- PASS: fresh signed release bundle b6699 at target/app-bundles/opt1240/Wavecrate OPT-1240.app; normal and compact visual smoke plus live source switching reached idle without a stranded progress owner. The final commit changes only cfg(test) maintenance execution.\n- PASS: fresh stateless complete-diff review at exact head 13cdd4f7c found no required fixes or observations.\n- PARTIAL: bash scripts/ci.sh agent passes smoke and architecture lanes, then remains red in pre-existing flaky/legacy coverage. The full Wavecrate suite repeated at 3864 passed / 53 failed; a prior run observed one additional intermittent failure, and controller_browser_integration failed once then passed 5/5 serially. None are in the focused OPT-1240 suites.\n\n## Known limitations\n- The repository broad agent lane remains red on independently tracked existing flaky/legacy tests. This change does not quarantine or alter those tests.\n- Deliberate contention and dropped/canceled lifecycle behavior are exercised deterministically rather than locking the users live source database.\n\nUser approval is required before merge.\n\nLinear: OPT-1240


PORTALSURFER (commented):
Review outcome: changes requested (self-review identity cannot submit GitHub’s REQUEST_CHANGES event). One required lifecycle-fencing fix is inline.

src/native_app/sample_library/folder_scan_actions/worker.rs
```diff
@@ -111,36 +114,142 @@ impl NativeAppState {
...
+                        );
+                        emit_lifecycle(
+                            &recovery_events,
+                            FolderScanLifecycle::Failed,
+                            "Source scan stopped unexpectedly",
+                            admission_generation,
                         );
                         let cancel = std::sync::Arc::new(std::sync::atomic::AtomicBool::new(true));
-                        return run_folder_scan_worker(request, events, cancel);
+                        run_folder_scan_worker(recovery_request, recovery_events, cancel)
```
PORTALSURFER: **Required fix (High) — retain the admitted generation and make this a fenced failed terminal completion.** The panic fallback creates a fresh canceled worker result here but never assigns the generation that was admitted before the panic. Consequently `finish_folder_scan` receives `lifecycle_generation = None`, skips its current-generation check, and can treat the retired worker as current if replacement races the panic. The prior `Failed` progress event is also immediately converted into the normal canceled/retry path instead of a terminal failure. Capture the actual admission generation before entering fallible work, attach it to the recovery completion, and route it through a failed terminal outcome (retiring the visible owner); only an explicit user retry should start a replacement.

PORTALSURFER: Fixed in 89b4a0a90. Panic recovery now records the admitted lifecycle generation before fallible scan work, attaches it to a typed terminal-failure completion, rechecks the source generation in the UI owner, retires the visible owner as Failed, and deliberately avoids scheduling an automatic retry. Added regression coverage proving worker failure releases ownership without a pending refresh. Validation: source_scan_workflow 32 passed; folder_scan_actions 15 passed; status_bar 49 passed; smoke passed.

src/native_app/app/state/source_scan_workflow.rs
```diff
@@ -332,6 +400,42 @@ impl SourceScanWorkflow {
         active
     }
 
+    pub(in crate::native_app) fn cancel_active_scan_by_user(
+        &mut self,
+        browser: &mut FolderBrowserState,
+    ) -> Option<(String, String)> {
+        let progress = self.progress.take()?;
+        if !browser.cancel_scan(&progress.source_id, progress.task_id) {
```
PORTALSURFER: **Required fix (High) — make recovery work after projection, or do not offer it there.** `PersistingResults` is restored as the visible owner after `finish_scan_with_lifecycle` has applied the browser result, and the stalled UI offers Retry/Cancel for every nonterminal lifecycle. Both actions enter this method, whose only success condition is `browser.cancel_scan(...)`; the already-applied browser scan is no longer the owner being displayed, so this restores `progress` and returns `None`, leaving the stalled persistence owner and buttons unchanged. Retire/fence the workflow's persistence owner by task/generation independently of the browser scan (and ignore its late maintenance completion), or restrict the recovery controls to pre-projection states where this cancellation is valid.

PORTALSURFER (commented):
Review outcome: changes requested (the authenticated PR-owner identity cannot submit GitHub's REQUEST_CHANGES event). One required recovery-path fix is inline.

src/native_app/app/state/source_scan_workflow.rs
PORTALSURFER: Fixed in d81b983c9. User recovery now distinguishes the post-projection PersistingResults workflow owner from the retired browser scan owner, retires it directly, clears pending equivalent work, and leaves the task/generation fence in place so late maintenance completion is ignored. Added a regression test covering post-projection cancellation and late completion. Validation: source_scan_workflow 33 passed; message_dispatch 22 passed; status_bar 49 passed; smoke passed.

PORTALSURFER (commented):
Clean review: the current merge-base-to-head diff satisfies the stated source-scan lifecycle and recovery Definition of Done. The persistence-phase cancellation path now retires the visible owner and rejects its late terminal completion; lifecycle generation fencing and panic recovery remain consistent with the PR contract.

src/native_app/sample_library/folder_scan_actions/worker.rs
```diff
@@ -267,22 +446,14 @@ impl NativeAppState {
...
-            if let Some(error) = result.scan_cache_error {
-                tracing::warn!("failed to persist source scan cache after scan: {error}");
-            }
-            for error in result.harvest_errors {
-                tracing::warn!("{error}");
-            }
+            let result = persist_folder_scan_maintenance_recovering(request.clone());
+            self.finish_folder_scan_maintenance(result, context);
         }
         context
```
PORTALSURFER, line 453: **Required fix (High) — test-build completion starts a duplicate maintenance owner.** The `#[cfg(test)]` block above executes `persist_folder_scan_maintenance_recovering(request.clone())` and immediately calls `finish_folder_scan_maintenance`, which emits the terminal lifecycle and releases the visible owner. This unconditional background submission then executes the same cache/config/harvest/rating-decay maintenance after that terminal completion. Consequently, test builds can report `Complete` before all maintenance for the scan has run, and the second result is discarded as stale (while its side effects and failures are no longer represented by the lifecycle). Make the test path choose exactly one maintenance execution—e.g. return after the synchronous completion and compile the background submission only outside `cfg(test)`—so terminal state follows the sole owner’s actual work.

PORTALSURFER (commented):
Changes requested: the test-build maintenance path still starts a second copy of the same maintenance request after synchronously completing and retiring its visible lifecycle owner. This violates the PR's owner/terminal ordering contract in the test configuration and can hide the actual asynchronous maintenance outcome.

src/native_app/sample_library/folder_scan_actions/worker.rs
PORTALSURFER: Fixed in 13cdd4f7c. Test builds now execute exactly one synchronous maintenance owner; the background submission is compiled only for non-test builds. The production-only completion message remains explicitly permitted as dead code in the test build. Focused lifecycle, folder-scan action, and message-dispatch suites pass (33 + 15 + 22 tests).

PORTALSURFER (commented):
Clean review: the current complete merge-base-to-head diff satisfies the typed source-scan lifecycle and recovery Definition of Done. Lifecycle/owner transitions, bounded admission diagnostics, terminal fencing, and the test-build maintenance path are consistent with the stated contract.

PR #893, opened 2026-07-23
OPT-1179 Make source discovery revisioned, ordered, and self-healing

PORTALSURFER:
## Goal

Make source discovery publish authoritative committed identity deltas and coordinate full scans with targeted watcher synchronization so readiness work starts only from committed, ordered source state.

## Scope and non-goals

- Carry the exact committed created, changed, moved, and deleted identity delta through foreground scan projection into readiness reconciliation.
- Keep unchanged authoritative scans as no-ops and fall back to complete reconciliation when projection or delta authority is incomplete.
- Coordinate source-scoped full scans and targeted watcher syncs: deduplicate paths, serialize commits, let full work subsume queued paths, and lifecycle-fence follow-up work.
- Recover targeted worker panics as terminal failures so the source lane cannot remain wedged.
- Document the causal discovery contract in docs/TARGET.md.
- Non-goal: redesign periodic lifecycle audit admission, which remains tracked by OPT-1245.

## Definition of Done

- Successful foreground scans publish their exact committed identity delta after the source database commit.
- Unchanged scans do not request generic rediscovery.
- Concurrent watcher paths coalesce into one ordered follow-up, while authoritative full refreshes supersede compatible queued paths.
- Source retirement and lifecycle changes purge or suppress stale causal work.
- Targeted sync failure or panic releases ownership and escalates through the bounded full-reconciliation path.

## Validation

- cargo fmt --all -- --check — passed
- cargo test -p wavecrate --lib foreground_scan_terminal_release_admits_coalesced_watcher_paths -- --test-threads=1 — passed
- cargo test -p wavecrate --lib source_scan_workflow -- --test-threads=1 — 37 passed
- cargo test -p wavecrate --lib filesystem_refresh -- --test-threads=1 — 26 passed
- cargo test -p wavecrate --lib native_app::source_processing::supervisor::tests -- --test-threads=1 — 92 passed, 1 ignored
- cargo test -p wavecrate-scan --lib -- --test-threads=1 — 95 passed
- bash scripts/ci.sh smoke — passed at c4dcfcd9d
- bash scripts/ci.sh agent — 3925 passed, 47 failed, 24 ignored; failures are the existing parallel shared-state/test-isolation family. Representative facade failure reproduces on origin/main, while affected supervisor cases pass serially.
- bash scripts/build.sh --release --variant OPT-1179-FIX — passed at c4dcfcd9d
- Signed single-instance app smoke: target/app-bundles/opt1179fix/Wavecrate OPT-1179-FIX.app launched with two sources and a usable browser, with zero Database is busy, disk-I/O, or malformed-database errors in the controlled run. Source database immutable quick_check returned ok afterward.

## Known limitations

- Periodic focus/lifecycle audit admission remains owned by OPT-1245.
- The repository-wide agent gate retains unrelated parallel test-isolation failures described above.

Closes OPT-1179


src/native_app/app/state/source_scan_workflow.rs
```diff
@@ -313,19 +347,44 @@ impl SourceScanWorkflow {
...
+                    target: "wavecrate::source_processing",
+                    source_id,
+                    path_count = paths.len(),
+                    outcome = "subsumed_by_pending_full_scan",
+                    "Source discovery causal plan merged watcher paths"
+                );
+                return SourceFilesystemChangePlan::DeferredAlreadyRunning { source_id };
+            }
+            if full_scan_active_for_source || self.active_targeted_syncs.contains_key(&source_id) {
+                self.queue_targeted_sync(source_id.clone(), paths, lifecycle_generation);
```
PORTALSURFER, line 378: **Required fix — queued watcher paths are never admitted after a normal foreground scan completes.** This branch stores paths in `pending_targeted_syncs` while the full scan is active, but the new drain (`maybe_run_pending_source_refresh`) is only called by targeted-sync completion. The normal foreground-scan terminal path clears `progress` in `finish_folder_scan_maintenance` without invoking that drain, so if no targeted sync was already running and no later event arrives, these paths remain queued indefinitely. That violates the DoD requirement that concurrent watcher paths coalesce into one ordered follow-up. After releasing the foreground scan owner, trigger the pending causal-plan drain (and add a runtime-level regression test that drives the normal completion path, rather than dequeuing directly from the workflow test).

PORTALSURFER (commented):
Required fix identified. GitHub does not permit this identity to request changes on its own pull request.

src/native_app/app/state/source_scan_workflow.rs
PORTALSURFER: Fixed in c4dcfcd9d. The foreground-scan terminal and worker-failure release paths now immediately drain the source causal plan after releasing ownership. Added the runtime-level foreground_scan_terminal_release_admits_coalesced_watcher_paths regression, which drives NativeAppState through watcher arrival during an active scan and verifies targeted follow-up admission. Validation: focused regression passed; source_scan_workflow 37 passed; filesystem_refresh 26 passed; smoke gate passed; signed single-instance app launch was clean.

PORTALSURFER (commented):
Clean review — the current diff fulfills the PR goal and Definition of Done: committed identity deltas are handed from authoritative foreground scans to readiness, unchanged scans remain no-ops, watcher paths are coalesced and lifecycle-fenced around full work, and targeted-worker failures/panics release ownership into full reconciliation. No required fixes found.

PR #894, opened 2026-07-23
OPT-1096 Make scan rename recovery deterministic

PORTALSURFER:
## Goal
Prevent scan order from deciding which same-content destination inherits a removed sample identity and its metadata.

## Scope
- stage hash-matching destinations neutrally until the scan has reconciled its complete relevant destination set
- prefer a unique stable filesystem identity before considering content-hash recovery
- transfer by content hash only for exactly one retained source and one eligible live destination
- keep ambiguous destinations independent while retaining the original metadata and analysis identity for later reconciliation
- apply the same final reconciliation boundary to full quick scans, hard rescans, and targeted watcher sync
- document the durable rename-ownership contract

## Non-goals
- do not remove rename recovery
- do not merge duplicate files or assign one Sample ID to multiple paths
- do not rehash unchanged small files
- do not redesign duplicate management

## Definition of Done
- one removed file plus two same-content destinations cannot transfer metadata based on visit or batch order
- a unique filesystem-identity match receives the original metadata even when the copy is visited first
- without a unique identity match, neither arbitrary destination inherits ratings, tags, collections, curation history, or analysis identity, and the original metadata remains pending
- the genuine one-source/one-destination rename path remains intact for quick, hard, and targeted scans
- pending, restart, live-duplicate, and retained-destination safeguards remain green

## Validation
- `cargo test -p wavecrate-scan --lib -- --test-threads=1` — 97 passed
- `cargo test -p wavecrate-library sample_sources::db::schema::migrations::tests` — 10 passed
- `cargo test -p wavecrate-library source_db_migration_tests` — 11 passed
- `cargo clippy -p wavecrate-scan --all-targets --no-deps -- -D warnings -A clippy::too_many_arguments` — passed
- `git diff --check` — passed

The unmodified `deep_hash_scan_with_post_hash_hook` on `main` already triggers Clippy `too_many_arguments`; the focused Clippy validation allows only that existing lint.

## Known limitations
None.

### DIFF
--- a/crates/wavecrate-scan/src/sample_sources/scanner/scan/runner.rs
+++ b/crates/wavecrate-scan/src/sample_sources/scanner/scan/runner.rs
@@ -217,9 +217,6 @@ pub(crate) fn reconcile_scan_renames(
     cancel: Option<&AtomicBool>,
     writer: &impl ScanWriter,
 ) -> Result<(u64, Vec<SourceManifestEntry>), ScanError> {
-    if context.mode == ScanMode::Hard {
-        return Ok(committed_snapshot);
-    }
     let candidates = context
         .stats
         .rename_candidate_paths
@@ -232,7 +229,7 @@ pub(crate) fn reconcile_scan_renames(
         cancel,
         writer,
     )?;
-    if renamed.is_empty() {
+    if renamed.is_empty() && context.mode != ScanMode::Hard {
         return Ok(committed_snapshot);
     }
 
@@ -267,6 +264,13 @@ pub(crate) fn reconcile_scan_renames(
     context.stats.updated += renamed.len();
     context.stats.renames_reconciled += renamed.len();
     context.stats.renamed_samples.extend(renamed);
+    if context.mode == ScanMode::Hard {
+        let _writer = writer.lock(super::super::scan_writer::ScanWritePhase::Manifest);
+        let mut batch = db.write_batch()?;
+        batch.clear_all_pending_renames()?;
+        batch.clear_all_pending_rename_destinations()?;
+        batch.commit()?;
+    }
     Ok(db.manifest_snapshot_with_revision()?)
 }
 
--- a/crates/wavecrate-scan/src/sample_sources/scanner/scan/runner.rs
+++ b/crates/wavecrate-scan/src/sample_sources/scanner/scan/runner.rs
@@ -22,8 +22,8 @@ pub enum ScanMode {
     /// Update the database with new/modified files and mark missing entries.
     /// Full hashing is staged for large files to keep quick scans responsive.
     Quick,
-    /// Force a full rescan, pruning missing rows and unmatched pending renames
-    /// to rebuild state from disk.
+    /// Force a full rescan, pruning missing rows and pending renames without
+    /// matching current destinations to rebuild state from disk.
     Hard,
 }
 
@@ -34,7 +34,7 @@ pub fn scan_once(db: &SourceDatabase) -> Result<ScanStats, ScanError> {
 }
 
 /// Rescan the entire source, pruning rows for files that no longer exist and
-/// clearing any unmatched pending rename rows left over from prior quick scans.
+/// clearing pending rename rows that have no matching current destinations.
 pub fn hard_rescan(db: &SourceDatabase) -> Result<ScanStats, ScanError> {
     scan(db, ScanMode::Hard, None, None, None)
 }
@@ -265,9 +265,28 @@ pub(crate) fn reconcile_scan_renames(
     context.stats.renames_reconciled += renamed.len();
     context.stats.renamed_samples.extend(renamed);
     if context.mode == ScanMode::Hard {
+        let candidate_hashes = db
+            .list_manifest_entries()?
+            .into_iter()
+            .filter(|entry| candidates.contains(&entry.relative_path))
+            .filter_map(|entry| entry.content_hash)
+            .collect::<HashSet<_>>();
+        let pending_to_clear = db
+            .list_pending_renames()?
+            .into_iter()
+            .filter(|entry| {
+                entry
+                    .content_hash
+                    .as_ref()
+                    .is_none_or(|hash| !candidate_hashes.contains(hash))
+            })
+            .map(|entry| entry.relative_path)
+            .collect::<Vec<_>>();
         let _writer = writer.lock(super::super::scan_writer::ScanWritePhase::Manifest);
         let mut batch = db.write_batch()?;
-        batch.clear_all_pending_renames()?;
+        for path in pending_to_clear {
+            batch.clear_pending_rename(&path)?;
+        }
         batch.clear_all_pending_rename_destinations()?;
         batch.commit()?;
     }
--- a/crates/wavecrate-scan/src/sample_sources/scanner/scan/runner.rs
+++ b/crates/wavecrate-scan/src/sample_sources/scanner/scan/runner.rs
@@ -6,6 +6,7 @@ use std::sync::atomic::AtomicBool;
 use std::thread;
 
 use crate::sample_sources::SourceDatabase;
+use crate::sample_sources::db::SourceManifestEntry;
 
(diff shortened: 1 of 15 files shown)


crates/wavecrate-scan/src/sample_sources/scanner/scan/runner.rs
```diff
@@ -194,10 +195,81 @@ fn scan_with_writer(
...
 
+pub(crate) fn reconcile_scan_renames(
+    db: &SourceDatabase,
+    context: &mut ScanContext,
+    manifest_before: &[SourceManifestEntry],
+    committed_snapshot: (u64, Vec<SourceManifestEntry>),
+    cancel: Option<&AtomicBool>,
+    writer: &impl ScanWriter,
+) -> Result<(u64, Vec<SourceManifestEntry>), ScanError> {
+    if context.mode == ScanMode::Hard {
```
PORTALSURFER: Required fix: This early return makes `hard_rescan` skip the replacement reconciliation entirely. The diff also removes the previous in-loop hash-based transfer, and `mark_missing` only stages pending rows for `Targeted | Quick`, so a normal one-source/one-destination rename during a hard rescan now leaves the new row neutral and discards the old metadata/analysis identity. That violates the PR’s `do not remove rename recovery` non-goal and its Definition of Done that the genuine rename path remains intact. Preserve deterministic reconciliation for hard scans too (or retain an equivalent hard-scan rename path) and add a hard-rescan regression that verifies metadata and analysis-identity transfer.

PORTALSURFER: Fixed in 68d667141. Hard rescans now stage missing rows through the complete-set reconciliation, preserve a unique genuine rename, and clear unmatched pending rows before returning. Added `hard_rescan_preserves_genuine_rename_metadata_and_analysis_identity`, which verifies rating/user metadata, feature identity, and analysis-job path transfer. Exact-head validation: scanner lib 95/95 serial; source DB migrations 11/11; focused Clippy passed with only the documented pre-existing argument-count lint allowed; `bash scripts/ci.sh smoke`, formatting, and diff checks passed.

crates/wavecrate-scan/src/sample_sources/scanner/scan/runner.rs
```diff
@@ -194,10 +195,85 @@ fn scan_with_writer(
...
+            .iter()
+            .any(|rename| rename.new_relative_path == *candidate)
+    });
+    context.stats.updated += renamed.len();
+    context.stats.renames_reconciled += renamed.len();
+    context.stats.renamed_samples.extend(renamed);
+    if context.mode == ScanMode::Hard {
+        let _writer = writer.lock(super::super::scan_writer::ScanWritePhase::Manifest);
+        let mut batch = db.write_batch()?;
+        batch.clear_all_pending_renames()?;
```
PORTALSURFER: Required fix\n\nThis unconditional hard-rescan cleanup deletes every unresolved pending rename after reconciliation, including the one-source/two-same-content-destination case this PR must keep pending for later reconciliation. The new reconciler correctly leaves that case unmatched, but lines 270–271 then erase the original metadata and analysis identity instead of retaining it. Preserve unresolved pending rows (and add a hard-rescan ambiguity regression) while still clearing only state that was safely reconciled.

PORTALSURFER (commented):
Required fix: the hard-rescan reconciliation violates the pending-ambiguity requirement.

crates/wavecrate-scan/src/sample_sources/scanner/scan/runner.rs
PORTALSURFER: Fixed in 98359c20d. Hard cleanup now clears only pending rows with no matching current candidate hash; unresolved same-content candidates remain pending, while plain deletes continue to be pruned. Added `hard_rescan_retains_metadata_for_ambiguous_same_content_destinations`, which proves both destinations remain neutral and the original rating, user metadata, and analysis identity remain retained. Exact-head validation: scanner lib 96/96 serial; source DB migrations 11/11; focused Clippy passed with only the documented pre-existing argument-count lint allowed; smoke, formatting, and diff checks passed.

PORTALSURFER (commented):
Required fixes: the current reconciliation still misses two Definition-of-Done paths. GitHub does not allow this identity to request changes on its own PR.

crates/wavecrate-scan/src/sample_sources/scanner/scan_hash.rs
```diff
@@ -429,43 +524,29 @@ fn reconcile_missing_renames(
...
-                })
-                .collect::<Vec<_>>();
-            let [present_path] = matching_facts.as_slice() else {
-                continue;
-            };
-            if !rename_candidates.contains(*present_path) {
-                continue;
-            }
-            *present_path
+        let [present_path] = present_paths.as_slice() else {
```
PORTALSURFER: Required fix: this requires every live row with the hash to be a singleton, not merely the eligible destination set. With one removed source, one newly discovered `new.wav`, and an unchanged pre-existing duplicate that shares the hash, `present_paths` has two rows while `candidates` has exactly `new.wav`, so the genuine one-source/one-destination rename is refused. That contradicts the stated content-hash rule (one retained source and one eligible live destination) and the non-goal of redesigning duplicate management. Select the sole eligible candidate while retaining the multi-candidate rejection, and add a regression with an unchanged same-hash duplicate.

crates/wavecrate-scan/src/sample_sources/scanner/scan/runner.rs
```diff
@@ -194,10 +195,104 @@ fn scan_with_writer(
...
+                    .is_none_or(|hash| !candidate_hashes.contains(hash))
+            })
+            .map(|entry| entry.relative_path)
+            .collect::<Vec<_>>();
+        let _writer = writer.lock(super::super::scan_writer::ScanWritePhase::Manifest);
+        let mut batch = db.write_batch()?;
+        for path in pending_to_clear {
+            batch.clear_pending_rename(&path)?;
+        }
+        batch.clear_all_pending_rename_destinations()?;
```
PORTALSURFER: Required fix: do not discard all staged destinations while an ambiguous hard-rescan recovery remains pending. After the two-copy ambiguous case, this deletes both destinations' eligibility records. If one copy is removed before the next hard rescan, the survivor is an existing manifest row rather than a new candidate; `candidate_hashes` is then empty and lines 274-288 clear the original pending metadata instead of reconciling it. This violates the requirement to retain the original metadata and analysis identity for later reconciliation. Preserve/refresh the eligible-destination state until the ambiguity resolves, and cover ambiguous hard rescan followed by removal of one copy.

crates/wavecrate-scan/src/sample_sources/scanner/scan_hash.rs
PORTALSURFER: Fixed in 858b3c11e. Content-hash reconciliation now requires exactly one eligible destination, independent of unrelated live rows sharing the same hash. Added `sole_eligible_destination_reconciles_when_unchanged_duplicate_exists`, proving the candidate inherits metadata while the unchanged duplicate remains neutral. Validation: scanner 96/96 serial; structural migrations 10/10; source migration contract 11/11; focused Clippy, formatting, diff checks, and smoke all passed.

crates/wavecrate-scan/src/sample_sources/scanner/scan/runner.rs
PORTALSURFER: Fixed in 858b3c11e. Eligible destinations are now retained by verified content hash while recovery is unresolved, survive hard rescans, and are rehashed before later identity transfer. Removed retained candidates are not restaged as competing sources; invalid associations are pruned. The hard-rescan ambiguity regression now deletes one copy on a later scan and proves deterministic metadata and analysis-identity transfer to the survivor. Validation: scanner 96/96 serial; structural migrations 10/10; source migration contract 11/11; focused Clippy, formatting, diff checks, and smoke all passed.

PR #897, opened 2026-07-23
Add conditional overflow fades across the browser

PORTALSURFER:
## Goal

Make clipped browser surfaces communicate that more content is available, while keeping the waveform cue lighter and directional and preserving reliable sample-list scrolling.

## Scope

- Paint 48 px directional vignettes for overflowing folder tree, collections, filter controls, and metadata tags.
- Drive each cue from Radiant's live scroll affordance: bottom when content is hidden below, top when it is hidden above, and both between the endpoints.
- Crossfade each cue over 260 ms as clipping begins, ends, or moves between edges; only request paint-only animation frames during those short transitions.
- Make the collections panel's strength ramp visibly begin at its first clipped pixel while reaching full strength over roughly one row.
- Add subtle, similarly animated waveform edge fades only when zooming or panning leaves audio beyond the visible left or right edge.
- Give the metadata tags scroll viewport a stable registered id so its overlay can follow the live clip.
- Keep the virtualized sample-item list out of the transient overlay path and retain its already-materialized rows while wheel movement stays within overscan, updating safely at viewport boundaries.
- Arm otherwise-idle browser and waveform transitions for one visibility sample, then run fades on a dedicated capped 60 Hz paint-only overlay without changing playback's uncapped overlay cadence.
- Arm only from targeted accepted scrolls and effective panel-size changes; folder-tree movement stays paint-only inside its retained window and applies a normal window update only at materialization boundaries.

## Non-goals

- No changes to browser data, waveform navigation, audio processing, Radiant runtime APIs, or sample-list scroll routing.

## Definition of Done

- Every named vertical browser surface displays a lower cue at its top scroll position, an upper cue at its bottom scroll position, and both while content is hidden in both directions.
- Cues fade between states rather than popping, and disappear after content fits.
- After startup has settled, an accepted vertical scroll, effective sidebar resize, or waveform viewport change reliably arms the first fade frame, advances over 260 ms, and stops fade-only activity after settlement.
- Playback, worker, source, and Starmap primitives remain owned by the existing uncapped app overlay and are not replayed by the fade overlay.
- The waveform has no cue when fully zoomed out, a right cue at the start of zoomed audio, a left cue at its end, and both when panned between.
- The sample-item virtual list reaches its bottom reliably, preserves hover rendering, and avoids per-wheel host reprojection while overscan still covers the full visible viewport.
- Regression tests cover transition scheduling, vertical edge detection/direction, waveform one- and two-sided conditions, collection strength, widget-id registration, retained-window boundaries, and sample-list coverage.

## Validation

Current fixed head `acc7ffe78`:

- PASS: `cargo fmt --all -- --check`
- PASS: `git diff --check`
- PASS: `bash scripts/ci.sh smoke`
- PASS: production AppBridge idle vertical-scroll and idle waveform-zoom lifecycle coverage (6 focused bridge tests)
- PASS: overflow-fade tests (10)
- PASS: file-scroll tests (12)
- PASS: folder-tree retained-window/boundary tests (19)
- PASS: targeted dispatch/arming tests (15)
- PASS: idle scene frame-clock target remains 60 Hz; playback transient overlay remains uncapped.
- PASS: conflict resolution retains current `main` Radiant pin `c8398744`, whose history contains the PR's required retained-window support.
- BASELINE: full Wavecrate library run completed with 4,007 passed, 58 failed, and 27 ignored; failures are existing environment/shared-state baselines outside this diff.

Earlier dependency evidence:

- Radiant PR #1431: `cargo test -q --lib` (1,687 passed), including retained-window and partial-row native-runtime coverage.

## Known limitations

The native preview process launched successfully, but this desktop session could not attach its accessibility surface before timeout; compiled integration checks and production-bridge UI lifecycle tests are green.


src/native_app/app_chrome/scene.rs
```diff
@@ -29,7 +33,10 @@ fn frame_clock() -> ui::FrameClock<NativeAppState, GuiMessage> {
 fn app_transient_overlay() -> ui::TransientOverlay<NativeAppState> {
     ui::TransientOverlay::new(APP_TRANSIENT_OVERLAY_KEY)
         .paint_only()
-        .when(|state: &mut NativeAppState| state.should_paint_app_transient_overlay())
+        .when(|state: &mut NativeAppState| {
+            state.should_paint_app_transient_overlay()
+                || state.ui.chrome.overflow_fades.is_animating()
```
PORTALSURFER: Required fix — High: This activation predicate is false for a newly clipped idle surface, because `overflow_fades` starts empty and the only code that inserts a transition is the later overlay painter (`opacity` / `opacity_from`). The new folder/filter/tag and waveform cues use a zero entry opacity, so no cue is painted on that first pass. With no unrelated playback/worker overlay active, line 38 therefore requests no paint-only frame to run that painter and advance the 260 ms transition; the cues only start if another repaint happens incidentally. The focused tests call the painter directly and do not exercise this activity ordering. This violates the Definition of Done that every named surface shows and crossfades its directional cue. Arm a first paint-only frame when scroll/waveform clipping changes (or expose an equivalent pending-visibility condition in this predicate), then continue only while transitions are active. Add an integration-level idle scroll/zoom test that proves the first frame enters, the transition advances, and scheduling stops once settled.

PORTALSURFER: Fixed in `55d1274e768330dd4fc24dc32106df2ce7625c0e`. Overflow fades now have a one-shot visibility sample and a dedicated capped 60 Hz paint-only overlay; waveform viewport changes explicitly arm the sample, while real vertical scroll redraws let the painter discover live clipping. Production AppBridge regressions cover otherwise-idle vertical scroll and waveform zoom, prove the fade advances and settles, and preserve the existing idle 60 Hz and uncapped playback activity contracts. Focused overflow tests (10), file-scroll tests (12), both bridge regressions, and `bash scripts/ci.sh smoke` pass.

src/native_app/app_chrome/scene.rs
```diff
@@ -41,9 +55,106 @@ fn paint_app_transient_overlay(
     state.paint_waveform_transient_overlay(context, primitives);
     state.paint_worker_progress_indicator(context, primitives);
     state.paint_source_processing_source_pulse(context, primitives);
+    paint_library_overflow_fades(state, context, primitives);
```
PORTALSURFER: Required fix — High (UI/performance): Keep the new fade overlay independent of the existing playback overlay. `app_transient_overlay` is still registered separately and paints whenever `should_paint_app_transient_overlay()` is true. During an active fade plus playback, this branch calls that entire painter a second time, duplicating the waveform/playback/worker/Starmap primitives; it also advances and paints the fades on the playback overlays uncapped cadence rather than solely on the new 60 Hz overlay. Paint only `paint_library_overflow_fades` and `paint_waveform_scroll_fades` from the fade overlay, and remove those fade painters from `app_transient_overlay`.\n\nThe same separation must include browser arming: in the current diff, the only `overflow_fades.arm()` is the waveform interaction path. Once the default startup sample is consumed, an otherwise-idle folder-tree, collections, filter, or tag scroll (and a collections resize that changes clipping) cannot make this `.when(...)` predicate true, so it never establishes or advances the required fade. Arm/request one paint-only sample from accepted browser-scroll/clip changes and extend the bridge regression to consume the initial sample before scrolling.

PORTALSURFER (commented):
Required fix posted inline; GitHub does not permit this identity to request changes on its own PR.

src/native_app/app_chrome/scene.rs
PORTALSURFER: Fixed in `acc7ffe78367c6025cec911ecdb3c31ff8c6c85d`. The existing app overlay now exclusively paints playback/worker/source/Starmap primitives; the capped 60 Hz fade overlay exclusively consumes and paints browser/waveform fades, with no delegation or duplicate primitive path. Targeted accepted-scroll messages arm collections, filters, tags, and folder-tree fades; folder-tree scrolling remains paint-only inside the retained window and applies the existing window change only at materialization boundaries. Effective panel resizes also re-arm clipping while clamped/no-op resizes do not. The bridge regression now consumes and settles startup before the later scroll. Focused AppBridge (6), overflow (10), file-scroll (12), folder-tree (19), dispatch (15), formatting/diff checks, and `bash scripts/ci.sh smoke` pass.

PORTALSURFER (commented):
Performance review (PR description and current merge-base-to-head diff): clean at acc7ffe78367c6025cec911ecdb3c31ff8c6c85d.\n\nVerified the dedicated capped 60 Hz fade overlay owns only fade primitives; the existing app overlay retains playback, worker, source, and Starmap primitives without replay. Targeted accepted-scroll and waveform paths arm paint-only transitions, effective panel resizing samples the required reprojected clip, and folder-tree retained-window updates defer host materialization until its boundary while the sample list remains on its existing retained-materialized path. No Required fix.

PR #900, opened 2026-07-23
OPT-1236: contain targeted symlink sync

PORTALSURFER:
## Goal
Close OPT-1236's targeted watcher-sync symlink escape: a watcher-reported link, or a descendant below one, must not enumerate or index content outside the configured source.

## Scope
- Traverse targeted paths from a capability-scoped source-root handle, opening each component and file without following links.
- Carry the verified file descriptor through fact collection, hashing, and the apply boundary.
- Re-open and identity-check the manifest path immediately before commit; reject link, missing, non-file, and changed bindings safely.
- Reject parent-relative targets, preserve ordinary reconciliation, and retire rows when a tracked file becomes a link.
- Verify the native browser projection removes the retired file rather than exposing its link target.

## Non-goals
- No opt-in symlink traversal, watcher-debounce redesign, or source-root replacement changes.

## Definition of Done
- File links, directory links, ancestor loops, outside-root links, and link descendants are not indexed by targeted sync.
- A replaced regular WAV is removed from the source manifest and browser projection, including when replacement occurs during targeted apply.
- Existing nested-file reconciliation remains intact.

## Validation
- `cargo fmt --check`
- `cargo test -p wavecrate-scan --lib targeted_sync -- --test-threads=1` (11 passed)
- `cargo test -p wavecrate-scan --lib -- --test-threads=1` (102 passed)
- `cargo test -p wavecrate --lib filesystem_sync_ -- --test-threads=1` (6 passed)
- `bash scripts/ci.sh smoke` (passed)
- Fresh specialist filesystem-boundary review: clean at `142333217`.

## Known limitations
`bash scripts/ci.sh agent` reaches its broad Wavecrate library lane but remains red in unrelated pre-existing UI/status and destructive-edit tests. Focused scanner/native tests and the smoke gate are green.

### DIFF
--- a/crates/wavecrate-scan/src/sample_sources/scanner/scan_paths.rs
+++ b/crates/wavecrate-scan/src/sample_sources/scanner/scan_paths.rs
@@ -1,16 +1,19 @@
 use std::{
-    collections::{BTreeSet, HashMap},
-    fs,
+    collections::{BTreeMap, BTreeSet, HashMap},
+    io,
     path::{Component, Path, PathBuf},
     sync::atomic::{AtomicBool, Ordering},
 };
+
+use cap_fs_ext::{DirExt, FollowSymlinks, OpenOptionsFollowExt, ambient_authority};
+use cap_std::fs::{Dir, OpenOptions};
 
 use crate::sample_sources::{SourceDatabase, WavEntry, is_supported_audio};
 
 use super::{
     scan::{ScanContext, ScanError, ScanMode, ScanStats},
     scan_db_sync::db_sync_phase,
-    scan_diff_phase::prepare_diff,
+    scan_diff_phase::prepare_diff_from_facts,
     scan_fs::ensure_root_dir,
     scan_walk::apply_prepared_chunk,
     scan_writer::{ScanWriter, UncoordinatedScanWriter},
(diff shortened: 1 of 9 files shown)


crates/wavecrate-scan/src/sample_sources/scanner/scan_paths.rs
```diff
@@ -168,14 +173,56 @@ fn collect_current_files(
     cancel: Option<&AtomicBool>,
     current_files: &mut BTreeSet<PathBuf>,
 ) -> Result<(), ScanError> {
-    if absolute_path.is_dir() {
+    if has_symlinked_target_ancestor(root, absolute_path)? {
```
PORTALSURFER: Required fix: this metadata check is separated from later path-based enumeration/preparation. A source can replace the checked entry with a symlink in between. Bind the no-follow classification to the exact object consumed, rather than narrowing the window with another check.

PORTALSURFER (commented):
Required fix: source-boundary review found a check/use race. The current no-follow classification must be bound to the object later enumerated or indexed before this PR can be considered ready.

crates/wavecrate-scan/src/sample_sources/scanner/scan_paths.rs
PORTALSURFER: Resolved in 53d0b4b0d. Targeted sync now traverses from a capability-scoped source-root directory, opens every component and file with no-follow semantics, and carries the resulting file descriptor through fact collection and hashing. Apply uses that verified descriptor snapshot rather than re-resolving the watcher path.

PORTALSURFER: Follow-up in 412648fee: the verified descriptor is retained through apply. Immediately before the path-keyed manifest/projection write, targeted sync reopens every component and the leaf with no-follow semantics and compares the resulting handle identity; a replaced file or directory link is rejected and the stale row is retired.

PORTALSURFER: Follow-up in 142333217: the apply fence now distinguishes a definite missing/link/non-file result from ordinary regular-file churn. Definite link or missing replacements remain in `context.existing`, so the existing missing-row phase retires the manifest and browser-projection row. Added a regression that replaces the verified WAV with an outside symlink immediately before apply.

PORTALSURFER (commented):
Clean specialist review at 142333217: targeted sync now keeps discovery, hashing, and apply bound to no-follow source handles, verifies the manifest path binding at apply time, and preserves link/missing replacements for normal row retirement. No required fixes remain.

PR #901, opened 2026-07-23
Preserve uncertain scan subtrees

PORTALSURFER:
## Goal
Prevent uncertain nested filesystem traversal from being reconciled as deletion or a move.

## Scope
- Carry unreadable and unenumerated relative prefixes from full and targeted traversal into missing-row reconciliation.
- Canonicalize targeted paths before prefix matching so `./subtree` shares the same safety boundary.
- Preserve matching manifest rows and pending rename metadata, return a retryable incomplete result, and leave completion metadata unchanged.
- Skip rename reconciliation during a partial traversal; a later authoritative pass converges normally.

Non-goals: changing confirmed-missing root handling or source availability semantics.

## Definition of Done
- Full scans, targeted sync, manifest audits, and hard rescans preserve descendants below uncertain prefixes.
- Iterator and entry-type faults cannot publish deleted or moved identities for unobserved descendants.
- Recovery reconciles real deletion after readability returns.

## Validation
- `cargo test -p wavecrate-scan --lib -- --test-threads=1` (109 passed)
- Focused native manifest-audit and filesystem-refresh test lanes
- `bash scripts/ci.sh agent`
- Independent review is being rerun at the current head

## Known limitations
None.

Fixes OPT-1241.

### DIFF
--- a/crates/wavecrate-scan/src/sample_sources/scanner/scan/context.rs
+++ b/crates/wavecrate-scan/src/sample_sources/scanner/scan/context.rs
@@ -1,4 +1,4 @@
-use std::collections::{BTreeMap, HashMap, HashSet};
+use std::collections::{BTreeMap, BTreeSet, HashMap, HashSet};
 use std::path::PathBuf;
 
 use crate::sample_sources::SourceDatabase;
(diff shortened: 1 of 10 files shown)


PORTALSURFER (commented):
Clean review: the published diff meets the PR goal and Definition of Done. Uncertain full and targeted traversal prefixes preserve indexed rows and pending rename state, produce a retryable incomplete result, and recover on the next authoritative pass. Validated at f83cf922bc32860c6e843b389ad5f8751073ea85.

PR #902, opened 2026-07-23
Share Wavecrate source entry traversal policy

PORTALSURFER:
## Goal
Unify no-follow source-entry classification so full scans, targeted sync, browser traversal, and watcher refreshes cannot drift on source-boundary behavior.

## Scope and non-goals
- Adds a library-owned typed entry policy for regular files, directories, links/reparse points, AppleDouble sidecars, supported audio, hidden-directory indexing, and bounded probe outcomes.
- Routes full scans, targeted sync, deep-hash checks, folder-browser traversal/refresh, and watcher candidates through that policy.
- Rejects non-Unicode relative paths before persistence instead of lossy conversion; retains case and Unicode-normalization distinctions.
- Does not enable symlink traversal, alter supported formats, merge UI ownership into scanning, or change committed-revision/uncertainty behavior.

## Definition of Done
- Source consumers share one no-follow classification contract.
- A live symlink replacement remains a bounded watcher candidate so targeted sync safely retires stale indexed rows without following it.
- AppleDouble, hidden-directory, unsupported-file, and persisted-identity behavior have regression coverage.

## Validation
- `cargo test -p wavecrate-library --lib sample_sources:: -- --test-threads=1`
- `cargo test -p wavecrate-scan --lib -- --test-threads=1`
- `cargo test --lib native_app::sample_library::source_watcher::tests -- --test-threads=1`
- `cargo test --lib native_app::sample_library::folder_browser::tests::source_scanning_symlinks -- --test-threads=1`
- `bash scripts/ci.sh agent`
- Independent current-diff review: clean.

## Known limitations
None. Windows reparse behavior uses the shared no-follow file-type contract; platform-specific runtime fixture coverage remains gated by the host.

### DIFF
--- a/crates/wavecrate-library/src/sample_sources/db/error.rs
+++ b/crates/wavecrate-library/src/sample_sources/db/error.rs
@@ -25,6 +25,9 @@ pub enum SourceDbError {
     /// Provided path contained disallowed components or was empty.
     #[error("Path contains invalid relative components: {0}")]
     InvalidRelativePath(PathBuf),
+    /// A relative path could not be represented without losing filesystem identity.
+    #[error("Path is not valid Unicode and cannot be persisted safely: {0}")]
+    NonUnicodeRelativePath(PathBuf),
     /// Database is locked or busy.
     #[error("Database is busy, please retry")]
     Busy,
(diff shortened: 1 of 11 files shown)


PORTALSURFER (commented):
Review: clean. The published PR description and exact head 120c9d4b6 align: the shared no-follow classifier is adopted by scan, browser, and watcher paths; the symlink-replacement reconciliation regression is covered; non-Unicode persisted paths are rejected before lossy conversion. No blocking mismatches found.

PR #904, opened 2026-07-23
OPT-1247 Bound and measure source content verification

PORTALSURFER:
## Goal

Replace the fixed eight-file periodic content sample with durable, measurable, resource-bounded content-verification rotations, while preserving OPT-1245's manifest-audit admission contract.

Root cause: throughput was fixed independently of library scale, and the single cursor/completion signal could not distinguish a completed manifest traversal from successfully verified content.

## Scope and non-goals

- Add durable per-source rotation, checkpoint, per-entry verification, retry, skip-reason, byte, and rotation-time state.
- Keep manifest traversal completion independent from content-verification coverage.
- Derive a configurable 30-day default coverage target from source scale, with hard time/byte/entry ceilings adjusted for playback, foreground work, external/network storage, and bounded accelerated recovery requests.
- Retry unavailable, unsupported, failed, or concurrently changed entries without starving forward traversal.
- Hash outside the runtime SQLite writer, then revalidate file facts and manifest revision inside a short authoritative commit.
- Emit structured per-source coverage metrics and publish only the exact committed content delta.
- Preserve and safely migrate the legacy cursor as traversal progress, never as verification evidence.
- Do not hash every file on every audit, weaken watcher-targeted verification, or change lifecycle audit admission.

## Definition of Done

- [x] Manifest completion and content coverage are separate durable facts.
- [x] Coverage reports rotation/revision, verified and remaining entries/bytes, skipped retries, bytes read, oldest-unverified age, and estimated/measured rotation time.
- [x] A configurable/testable target replaces the fixed batch while retaining elapsed-time, byte, entry, activity, and storage-class bounds.
- [x] Changed or skipped files remain due with bounded retry and forward-progress fairness.
- [x] Cancellation/restart resumes durable checkpoints without double-counting.
- [x] Same-facts content changes publish only after snapshot revalidation and revision-fenced commit.
- [x] Forced recovery requests use bounded accelerated budgets.
- [x] Regression coverage includes target sizing at 10,000/29,000 entries, time/byte bounds, retry fairness, cancellation/restart, cursor migration, writer-hold behavior, stale revisions, and exact committed publication.

## Validation

- `cargo test -p wavecrate-scan --lib` — 123 passed.
- `cargo test -p wavecrate-library --lib` — 243 passed.
- `cargo test -p wavecrate --lib manifest_audit --no-default-features` — 7 passed.
- `cargo test -p wavecrate-scan --lib sample_sources::scanner::scan_hash::tests` — 18 passed at final head.
- `cargo fmt --all` — passed.
- `bash scripts/ci.sh agent` — compiled successfully and reached the 4,013-test native library lane; 3,940 passed, 49 failed, 24 ignored. Failures were concentrated in existing shared native-app test state. The source-processing retry-fairness failure was rerun in isolation and reproduced identically on untouched `main` (`8b0af139d`).
- `bash scripts/agent.sh request` preflight — blocked before implementation by five existing native-app palette-boundary imports on `main`.

## Known limitations

- Signed-app interactive browsing/audition smoke was not run in this non-interactive implementation pass.
- Native Windows cross-check could not complete on the macOS host because the MSVC C/assembly toolchain is unavailable; the platform-neutral Windows drive-classification tests pass.
- One admitted file is atomic: a single file may exceed the byte ceiling, after which the slice stops before the next file. This explicit fallback avoids repeatedly starving an oversize file.


crates/wavecrate-scan/src/sample_sources/scanner/scan_hash.rs
```diff
@@ -28,51 +148,194 @@ fn cancel_requested(cancel: Option<&AtomicBool>) -> bool {
...
+    };
+    let states = db.content_audit_entry_states()?;
+    let mut retry = entries
+        .iter()
+        .filter(|entry| {
+            states.get(&entry.relative_path).is_some_and(|state| {
+                !state.verifies(entry, checkpoint.rotation_id) && state.retry_is_due(now)
+            })
+        })
+        .take(budget.retry_entries.min(budget.max_entries))
```
PORTALSURFER: **Required fix — retry reservation can starve forward traversal.** Both budget constructors force `retry_entries >= 1`, while adaptive coverage yields `max_entries == 1` for sources with up to 30 entries. When any skipped entry is retry-due, this line selects the only slot for it and line 229 leaves zero forward slots. If that retry remains unavailable/unsupported/failed, each daily audit resets its retry deadline and no later entry is ever visited, so the rotation cannot meet the stated DoD that retries must not starve forward progress.

Reserve a forward slot whenever another unverified non-retry entry exists (or otherwise schedule retries without consuming the entire slice), and add a regression covering a due retry with `max_entries == 1` plus a later unverified entry.

PORTALSURFER (commented):
## Changes requested (self-review permission limitation)

I cannot submit `REQUEST_CHANGES` on this PR because the connected GitHub identity is its author. The inline comment below is a blocking required fix.

crates/wavecrate-scan/src/sample_sources/scanner/scan_hash.rs
PORTALSURFER: Fixed in `43d77d665`. Retry selection now reserves one slot whenever unverified forward work exists, so a due retry cannot consume a one-entry slice; adaptive multi-entry sources also receive a minimum two-entry ceiling. Added `due_retry_cannot_consume_the_only_forward_progress_slot` plus the small-source adaptive-budget assertion. Validation: `cargo test -p wavecrate-scan --lib` (118 passed), `cargo test -p wavecrate --lib manifest_audit --no-default-features` (7 passed), `cargo fmt --all`, and `git diff --check`.

crates/wavecrate-scan/src/sample_sources/scanner/scan_hash.rs
```diff
@@ -28,51 +149,204 @@ fn cancel_requested(cancel: Option<&AtomicBool>) -> bool {
...
+            .saturating_sub(usize::from(forward_is_due)),
+    );
+    let mut retry = entries
+        .iter()
+        .filter(|entry| {
+            states.get(&entry.relative_path).is_some_and(|state| {
+                !state.verifies(entry, checkpoint.rotation_id) && state.retry_is_due(now)
+            })
+        })
+        .take(retry_limit)
```
PORTALSURFER, line 301: **Required fix — retry selection is not fair.** This takes the first `retry_limit` due paths in manifest order every time and persists no retry cursor/round-robin state. Once forward traversal has reached only skipped entries, a permanently unavailable early path is selected again whenever it becomes due, so later retry-due paths are never re-attempted. That contradicts the PR's durable per-entry retry/coverage contract and can prevent a rotation from recovering even when later files become available. Persist or otherwise rotate retry selection independently (while retaining the forward reservation), and add a test with more due retries than the retry budget proving a later path is eventually retried.

PORTALSURFER (commented):
## Changes requested

The diff does not yet satisfy the Definition of Done for bounded retries: due retry selection can permanently starve later skipped entries. I could not formally request changes because this GitHub identity is the PR author.

crates/wavecrate-scan/src/sample_sources/scanner/scan_hash.rs
PORTALSURFER: Fixed in `6de23aa81`. The durable content-audit checkpoint now carries an independent retry cursor; due retries are selected round-robin from the next path while the existing forward-slot reservation remains in force. Added `retry_cursor_rotates_past_a_persistently_unavailable_path`, which uses more due retries than the one-entry retry budget and proves the later path is retried. Validation: `cargo test -p wavecrate-scan --lib` (119 passed), `cargo test -p wavecrate-library --lib` (243 passed), `cargo test -p wavecrate --lib manifest_audit --no-default-features` (7 passed), `cargo fmt --all`, and `git diff --check`.

PORTALSURFER (commented):
Required fixes identified. GitHub does not permit this identity (the PR author) to submit a formal request-for-changes review.

crates/wavecrate-scan/src/sample_sources/scanner/scan_hash.rs
```diff
@@ -28,51 +149,220 @@ fn cancel_requested(cancel: Option<&AtomicBool>) -> bool {
...
+            !retry_paths.contains(&entry.relative_path)
+                && states.get(&entry.relative_path).is_none_or(|state| {
+                    !state.verifies(entry, checkpoint.rotation_id) && state.skip_reason.is_none()
+                })
+        })
+        .take(budget.max_entries.saturating_sub(retry.len()))
         .cloned()
         .collect::<Vec<_>>();
+    let mut selected = Vec::with_capacity(retry.len() + forward.len());
+    selected.append(&mut retry);
```
PORTALSURFER: **Required fix** — The retry reservation protects only the entry count, but every retry is executed before any forward candidate. A due retry is allowed to hash one oversize file (or consume the elapsed-time limit); the next forward entry then trips the byte/time check at lines 269–274 and the slice ends. A persistently `changed_during_hash` or `hash_failed` large retry can therefore consume every periodic slice and indefinitely starve forward traversal, contradicting the stated retry-fairness requirement. Schedule/reserve a forward candidate before retry work, or reserve enough time/byte budget for it, and add a regression case with a persistent oversized retry plus an otherwise due forward entry.

crates/wavecrate-scan/src/sample_sources/scanner/scan_hash.rs
```diff
@@ -19,7 +22,125 @@ struct HashBackfill {
...
+    Local,
+    /// A removable, mounted, or network-addressed source.
+    ExternalOrNetwork,
+}
+
+impl ContentAuditStorage {
+    /// Conservatively classify a source root from platform path conventions.
+    pub fn classify(root: &std::path::Path) -> Self {
+        #[cfg(target_os = "windows")]
+        if root.to_string_lossy().starts_with(r"\\") {
```
PORTALSURFER: **Required fix** — Production derives the budget from `ContentAuditStorage::classify`, but this recognizes only UNC Windows paths. Mapped network drives (for example `Z:\\`) and removable drive-letter roots (for example `E:\\`) fall through to `Local`, receiving the 5 s / 512 MiB idle-local ceiling instead of the constrained external/network ceiling required by the PR. Detect the platform drive/mount type (or classify unknown drive roots conservatively) and add coverage for the applicable mapped/removable storage cases.

crates/wavecrate-scan/src/sample_sources/scanner/scan_hash.rs
PORTALSURFER: Fixed in `94b744950`. The durable checkpoint now persists the next scheduling lane, and mixed retry/forward candidates are interleaved. Every attempted retry makes forward the next lane (and vice versa), so even one atomic oversize file cannot monopolize successive slices. `oversized_retry_yields_the_next_slice_to_forward_progress` keeps a large retry unstable, reopens the database, and proves the next constrained slice verifies forward work first. Validation: scanner 121 passed, persistence 243 passed, root manifest 7 passed, formatting/diff checks passed.

crates/wavecrate-scan/src/sample_sources/scanner/scan_hash.rs
PORTALSURFER: Fixed in `94b744950`. Windows source roots now use `GetDriveTypeW`; only `DRIVE_FIXED` receives the local budget, while mapped/remote, removable, missing-root, optical, RAM-disk, and unknown classifications fail conservatively to the external/network budget. `windows_mapped_removable_and_unknown_drive_types_are_conservative` covers the relevant return types. Native Windows cross-check was attempted with the installed target but the macOS host lacks the required MSVC C/assembly toolchain; platform-neutral scanner tests pass.

PORTALSURFER (commented):
Review result: required fixes identified. GitHub does not allow this identity to request changes on its own PR.

crates/wavecrate-scan/src/sample_sources/scanner/scan_hash.rs
```diff
@@ -112,21 +463,52 @@ pub(super) fn verify_content_batch(
...
+        let _writer = writer.lock(ScanWritePhase::Manifest);
+        db.content_audit_report(now)?
+    };
+    if report.remaining_entries == 0 && report.total_entries > 0 {
+        let manifest_before = stats.manifest_after.clone();
+        let _writer = writer.lock(ScanWritePhase::Manifest);
+        let mut batch = db.write_batch()?;
+        batch.complete_content_audit_rotation(now, stats.committed_delta.revision)?;
+        let committed = batch.commit_with_manifest_snapshot()?;
+        super::manifest::publish_committed_delta(&mut stats, manifest_before, committed);
```
PORTALSURFER, line 521: **Required fix** — Preserve the content delta from the final verification slice. Line 487 has already published the delta generated by the just-committed hash updates; this second, rotation-state-only commit starts from `stats.manifest_after` and republishes its unchanged manifest snapshot into the same `stats`. That replaces the delta which `execute_candidate` later sends in `ManifestAuditCommitted`, so a same-facts content change in the slice that completes a rotation can be published as an empty delta. Commit the rotation bookkeeping without republishing a manifest delta (or otherwise retain the first committed content delta), and add a regression that changes a one-entry source during its rotation-completing slice and asserts the published committed delta contains that change.

crates/wavecrate-scan/src/sample_sources/scanner/scan_hash.rs
```diff
@@ -19,7 +22,174 @@ struct HashBackfill {
...
+            if root.starts_with("/Volumes") {
+                Self::ExternalOrNetwork
+            } else {
+                Self::Local
+            }
+        }
+        #[cfg(not(any(target_os = "windows", target_os = "macos")))]
+        {
+            let _ = root;
+            Self::Local
```
PORTALSURFER: **Required fix** — Do not classify every non-macOS/non-Windows root as local. The stated Definition of Done requires byte/time ceilings to adjust for external/network storage, but this branch unconditionally returns `Local`, giving a Linux-mounted network/removable source the 5-second/512 MiB local budget. Add conservative or mount-aware detection on this platform (defaulting unknown storage to `ExternalOrNetwork` is safe), with coverage for the selected behavior.

crates/wavecrate-scan/src/sample_sources/scanner/scan_hash.rs
PORTALSURFER: Fixed in `ad31bd08d`. Rotation completion now rebuilds the final committed delta from the original pre-slice manifest, preserving the hash update while advancing the published revision to the bookkeeping commit. Added `final_rotation_completion_preserves_the_committed_content_delta` for a one-entry same-facts content change. Validation: scanner 122 passed, persistence 243 passed, root manifest 7 passed, formatting and diff checks passed.

crates/wavecrate-scan/src/sample_sources/scanner/scan_hash.rs
PORTALSURFER: Fixed in `ad31bd08d`. Platforms without a mount-type classifier now fail conservatively to `ExternalOrNetwork` instead of receiving the local 5-second/512 MiB ceiling. The storage classification test covers that fallback alongside mapped/removable/unknown Windows drive types. Validation: scanner 122 passed, persistence 243 passed, root manifest 7 passed, formatting and diff checks passed.

PORTALSURFER (commented):
## Required fixes

1. macOS storage classification must conservatively cover external and network mounts outside `/Volumes`.

Reviewed head: `ad31bd08df40052dcbd08e0ce34d25cd0c8e6ea9`.

The diff does not yet completely achieve the PR definition of done because such mounts can receive the local-content budget instead of the required storage-class bounds.

`REQUEST_CHANGES` is not available because this is my own PR, so this review is submitted as a comment.

crates/wavecrate-scan/src/sample_sources/scanner/scan_hash.rs
```diff
@@ -19,7 +22,179 @@ struct HashBackfill {
...
+impl ContentAuditStorage {
+    /// Conservatively classify a source root from platform mount information.
+    pub fn classify(root: &std::path::Path) -> Self {
+        #[cfg(target_os = "windows")]
+        {
+            classify_windows_storage(root)
+        }
+        #[cfg(target_os = "macos")]
+        {
+            if root.starts_with("/Volumes") {
```
PORTALSURFER: **Required fix — macOS external/network sources are misclassified as local.**

This recognizes only paths below `/Volumes` as `ExternalOrNetwork`; every other macOS source gets the local 5s / 512 MiB / 4,096-entry slice. macOS allows SMB/NFS, removable, and automounted volumes at arbitrary mount points, so a source outside `/Volumes` can receive the aggressive budget despite the PR contract requiring storage-class bounds for external/network storage. Classify the actual mount type (or fail conservatively to `ExternalOrNetwork` unless locality is established), and add coverage for an external/network mount outside `/Volumes`.

PORTALSURFER: Fixed in `e9eb385a7`. macOS classification now establishes locality from the mounted filesystem device: only sources on the same device as the normal `/Users` data volume receive the local ceiling. Distinct mounted filesystems—including SMB/NFS/removable/autofs mounts at arbitrary paths—and metadata failures use `ExternalOrNetwork`. Added `macos_only_grants_local_budget_to_the_normal_local_mount`. Validation: scanner 123 passed, persistence 243 passed, root manifest 7 passed, formatting and diff checks passed.

PORTALSURFER (commented):
Clean review — reviewed current head `e9eb385a714a93d1e172e9e2b23ed0befcf6bd66` against the PR description and complete merge-base diff (`8b0af139dda0b41ec20b11f0e61636862d294b79…e9eb385a714a93d1e172e9e2b23ed0befcf6bd66`). The diff achieves the stated goal and Definition of Done: durable coverage state is separate from manifest completion; adaptive bounded rotations, retry fairness, cancellation/restart checkpoints, revision-fenced revalidation, and committed-delta publication are implemented with focused regression coverage. No required fixes.

PR #905, opened 2026-07-23
OPT-1248: Add seeded replayable source-scanning state machine

PORTALSURFER:
## Goal

Add deterministic, seeded, replayable state-machine coverage for high-risk source-scanning event orderings so ordering regressions can be reproduced and minimized without live libraries or nondeterministic sleeps.

Linear: [OPT-1248](https://linear.app/boostnlvp/issue/OPT-1248)

## Scope

- Add a reference-model state-machine harness over the versioned `small-multi-source` native fixture and real `wavecrate-scan` scanner.
- Generate filesystem mutations, watcher batches/overflow, focus and foreground requests, cancellation, restart, remove/re-add, root offline/online/replacement, and injected transaction/publication/watcher/hash/lifecycle failures.
- Check exact committed manifests and the production browser-search projection, lifecycle-scoped monotonic revisions, definitive supervisor publication outcomes, actual bounded queue coalescing, and controlled-quiescence liveness.
- Emit schema-versioned JSON replay artifacts with deterministic greedy shrinking that preserves lifecycle transitions and the original scanner-only/integrated execution mode.
- Retain six regression seeds in the bounded normal lane, plus explicit integrated-supervisor, 1,000-sequence stress, and artifact/seed replay commands.
- Expose narrow test-support controls and observations at the real transaction, watcher-delivery, lifecycle-replacement, and publication boundaries without changing production policy.

## Non-goals

- No live-library, wall-clock soak, or GUI automation dependency.
- No release or persistence schema changes.
- No broad source-processing cleanup outside the testability and rejected-publication retention required by this issue.

## Definition of Done

- [x] Seeded event sequences exercise the requested mutation, lifecycle, cancellation, retry, and restart orderings.
- [x] Every observable commit is checked against the reference filesystem model and production browser-search projection.
- [x] Real supervisor queue and definitive publication/fallback outcomes are compared with independent revision/cause expectations.
- [x] Transaction, watcher-delivery, lifecycle-replacement, and publication failures are observed at their production boundaries and converge through deterministic retry.
- [x] Rejected delivery/publication leaves exact pending work unchanged until successful retry.
- [x] Failures are minimized and saved as replay artifacts that preserve execution mode and validate fixture identity.
- [x] Six regression classes remain in normal CI and a 1,000-sequence stress command is documented.
- [x] Real scanner and explicit integrated supervisor lanes are both validated without sleep-based synchronization.

## Review fixes

- Preserve scanner-only versus integrated execution mode in replay schema v2; reject fixture name/version/seed drift.
- Replace enum-cardinality queue bookkeeping with repeated admission through the production supervisor coalescing helper and exact real pending-work observations.
- Record publication evidence only after the definitive health-publication boundary returns `Published` or exact `AlreadyPublished` fallback; never infer output from aggregate readiness.
- Deterministically inject `Rejected` at that boundary, retain the exact pending delta, and require one rejection followed by one accepted output.
- Drive integrated convergence from explicit supervisor events with a 4,096-transition bound; retain a timeout only as a deadlock diagnostic guard. The lane contains no sleep calls.
- Run the production browser-search worker projection against the independent reference model instead of rereading the manifest.
- Hold root-offline mutations unavailable until the real supervisor emits `Offline`, then restore and wait for terminal online health through the event channel.
- Produce transaction failure through real SQLite `BEGIN IMMEDIATE` contention at the scanner write-batch boundary and require the exact `Busy` outcome before retry.
- Reject watcher delivery inside the production delta-admission helper, prove pending scopes are unchanged, then admit the same delta and converge.
- Reject source removal after entering the serialized `replace_sources` boundary, prove the lifecycle generation is unchanged, then retry remove/re-add and converge.

## Validation

- `git diff --check` — passed
- `cargo check -p wavecrate --lib` — passed
- `cargo check -p wavecrate-library --no-default-features` — passed
- `cargo test -p wavecrate --lib health_publication_is_lifecycle_fenced_and_coalesced -- --test-threads=1` — passed, including deterministic rejection followed by accepted replay
- `cargo test -p wavecrate-scan` — 123 passed
- `cargo test -p wavecrate --lib state_machine_tests::artifacts::tests:: -- --test-threads=1` — 2 passed
- `cargo test -p wavecrate --lib source_processing_seeded_state_machine_normal_ci -- --nocapture` — passed, six retained seeds
- `cargo test -p wavecrate --lib source_processing_seeded_state_machine_integrated_supervisor -- --ignored --nocapture` — passed in 156.35s, including dedicated transaction, watcher-delivery, publication, and lifecycle rejection/retry replays
- `cargo test -p wavecrate --lib source_processing_seeded_state_machine_stress_1000 -- --ignored --nocapture` — passed, 1,000/1,000 sequences in 164.55s with production browser projection checks
- `cargo test -p wavecrate --lib native_app::source_processing::supervisor::tests -- --nocapture` — 94 passed, 1 ignored, with only the existing `unavailable_hash_path_backs_off_without_starving_later_files` baseline failure
- `bash scripts/ci.sh agent` — all compilation/architecture/integration gates passed; the parallel lib lane reported 3943 passed, the same known 49-failure set as the untouched `origin/main` baseline, and 27 ignored
- State-machine/supervisor changed-file size budget — passed; all remain at or below 400 lines
- Sleep audit — no `thread::sleep` calls remain in the state-machine lane

## Known limitations

- The integrated real-supervisor lane is explicit/ignored because it exercises real background execution and is intentionally separate from the bounded default CI lane.
- The repository-wide parallel test gate is not green on `main`, as documented above. No branch-only failure was found.

### DIFF
--- a/src/native_app/source_processing/supervisor/discovery.rs
+++ b/src/native_app/source_processing/supervisor/discovery.rs
@@ -107,12 +107,28 @@ pub(super) fn discover_candidates(
                     consumed_readiness_deltas.insert(source.id.as_str().to_string());
                 }
                 if let Some(health) = health {
-                    shared.publish_source_health(health.into_event(
-                        super::SourceProcessingLifecycle::new(
-                            source.id.as_str(),
-                            in_flight_work.lifecycle_generation,
-                        ),
+                    let health = health.into_event(super::SourceProcessingLifecycle::new(
+                        source.id.as_str(),
+                        in_flight_work.lifecycle_generation,
                     ));
+                    #[cfg(test)]
+                    if let Some(pending) = pending_readiness_deltas
+                        .get(source.id.as_str())
+                        .filter(|pending| !pending.state_machine_inputs.is_empty())
+                    {
+                        shared
+                            .state_machine_publications
+                            .lock()
+                            .unwrap_or_else(|poison| poison.into_inner())
+                            .push(super::StateMachinePublicationObservation {
+                                source_id: source.id.as_str().to_string(),
+                                lifecycle_generation: in_flight_work.lifecycle_generation,
+                                source_generation: health.source_generation,
+                                readiness_revision: health.readiness_revision,
+                                inputs: pending.state_machine_inputs.clone(),
+                            });
+                    }
+                    shared.publish_source_health(health);
                 }
                 shared
                     .control()
--- a/src/native_app/source_processing/supervisor/discovery.rs
+++ b/src/native_app/source_processing/supervisor/discovery.rs
@@ -103,18 +103,28 @@ pub(super) fn discover_candidates(
                 if !stats.cheap_noop_sweep {
                     source_stats.insert(source.id.as_str().to_string(), stats);
                 }
-                if pending_readiness_deltas.contains_key(source.id.as_str()) {
-                    consumed_readiness_deltas.insert(source.id.as_str().to_string());
-                }
+                let pending_readiness_delta = pending_readiness_deltas.get(source.id.as_str());
+                let mut consume_readiness_delta = pending_readiness_delta.is_some();
                 if let Some(health) = health {
                     let health = health.into_event(super::SourceProcessingLifecycle::new(
                         source.id.as_str(),
                         in_flight_work.lifecycle_generation,
                     ));
+                    let publication_outcome = shared.publish_source_health_outcome(health.clone());
+                    consume_readiness_delta = matches!(
+                        publication_outcome,
+                        super::SourceHealthPublicationOutcome::Published
+                            | super::SourceHealthPublicationOutcome::AlreadyPublished
+                            | super::SourceHealthPublicationOutcome::NoSink
+                    );
                     #[cfg(test)]
-                    if let Some(pending) = pending_readiness_deltas
-                        .get(source.id.as_str())
+                    if let Some(pending) = pending_readiness_delta
                         .filter(|pending| !pending.state_machine_inputs.is_empty())
+                        .filter(|_| {
+                            publication_outcome != super::SourceHealthPublicationOutcome::NoSink
+                                && publication_outcome
+                                    != super::SourceHealthPublicationOutcome::Superseded
+                        })
                     {
                         shared
                             .state_machine_publications
@@ -126,9 +136,12 @@ pub(super) fn discover_candidates(
                                 source_generation: health.source_generation,
                                 readiness_revision: health.readiness_revision,
                                 inputs: pending.state_machine_inputs.clone(),
+                                outcome: publication_outcome,
                             });
                     }
-                    shared.publish_source_health(health);
+                }
+                if consume_readiness_delta {
+                    consumed_readiness_deltas.insert(source.id.as_str().to_string());
                 }
                 shared
                     .control()
--- a/src/test_support/source_processing_state_machine/harness.rs
+++ b/src/test_support/source_processing_state_machine/harness.rs
@@ -44,6 +44,8 @@ pub(super) struct StateMachineHarness {
     pub(super) expected_supervisor_publications: BTreeSet<(u64, u64, ScanCause)>,
     pub(super) observed_supervisor_publications: BTreeSet<(u64, u64, ScanCause)>,
     pub(super) stale_supervisor_publications: BTreeSet<(u64, u64, ScanCause)>,
+    pub(super) expected_rejected_supervisor_publications: BTreeSet<(u64, u64, ScanCause)>,
+    pub(super) rejected_supervisor_publications: BTreeSet<(u64, u64, ScanCause)>,
     pub(super) last_actual_output_revisions: BTreeMap<u64, (i64, i64)>,
     pub(super) actual_queue_admissions: u64,
     pub(super) max_actual_pending_scopes: usize,
@@ -104,6 +106,8 @@ impl StateMachineHarness {
             expected_supervisor_publications: BTreeSet::new(),
             observed_supervisor_publications: BTreeSet::new(),
             stale_supervisor_publications: BTreeSet::new(),
+            expected_rejected_supervisor_publications: BTreeSet::new(),
+            rejected_supervisor_publications: BTreeSet::new(),
             last_actual_output_revisions: BTreeMap::new(),
             actual_queue_admissions: 0,
             max_actual_pending_scopes: 0,
@@ -148,39 +152,11 @@ impl StateMachineHarness {
         self.next_failure = None;
         self.model.queue(ScanCause::Retry);
         self.flush(ScanCause::Retry)?;
-        self.wait_for_integrated_settle()?;
+        self.wait_for_supervisor_convergence()?;
         self.create(6, false)?;
         self.flush(ScanCause::Foreground)?;
-        self.wait_for_integrated_settle()?;
+        self.wait_for_supervisor_convergence()?;
         self.assert_actual_publications()
-    }
-
-    fn wait_for_integrated_settle(&self) -> Result<(), String> {
-        let deadline = std::time::Instant::now() + std::time::Duration::from_secs(30);
-        loop {
-            let supervisor = self
-                .supervisor
-                .as_ref()
-                .expect("integrated initialization keeps its supervisor");
-            let runtime = super::super::liveness_tests::runtime_observation(
-                supervisor,
-                self.source.id.as_str(),
-            );
-            if super::super::liveness_tests::readiness_snapshot(&self.source).is_some()
-                && runtime.queue_depth == 0
-                && runtime.readiness_queue_depth == 0
-                && runtime.in_flight == 0
-                && !runtime.source_dirty
-            {
-                return Ok(());
-            }
-            if std::time::Instant::now() >= deadline {
-                return Err(format!(
-                    "integrated state-machine baseline did not settle: {runtime:?}"
-                ));
-            }
-            std::thread::sleep(std::time::Duration::from_millis(10));
-        }
     }
 
     fn apply(&mut self, event: &Event) -> Result<(), String> {
@@ -296,13 +272,13 @@ impl StateMachineHarness {
         let publication_lost = self.take_failure(FailureBoundary::Publication);
         self.accept_commit(effective_cause, &stats.committed_delta, publication_lost)?;
         self.admit_pending_publication_retries()?;
-        if publication_lost {
+        if publication_lost && self.supervisor.is_none() {
             if !stats.committed_delta.is_empty() {
                 self.pending_publication_retries
                     .push((stats.committed_delta.clone(), effective_cause));
             }
         } else {
-            self.admit_supervisor_delta(&stats.committed_delta, effective_cause)?;
+            self.admit_supervisor_delta(&stats.committed_delta, effective_cause, publication_lost)?;
         }
         self.model.watcher_paths.clear();
         if effective_cause == ScanCause::Watcher {
@@ -349,7 +325,9 @@ impl StateMachineHarness {
         if let Some(supervisor) = &self.supervisor {
             supervisor
                 .request_source_processing(self.source.id.as_str(), "state_machine_quiescence");
-            self.assert_runtime_liveness(supervisor)?;
+        }
+        if self.supervisor.is_some() {
+            self.wait_for_supervisor_convergence()?;
         }
         self.assert_actual_publications()?;
         Ok(())
--- a/src/test_support/source_processing_state_machine/harness.rs
+++ b/src/test_support/source_processing_state_machine/harness.rs
@@ -2,7 +2,10 @@ use std::{
     collections::{BTreeMap, BTreeSet},
     fs,
     path::PathBuf,
-    sync::atomic::{AtomicBool, Ordering},
+    sync::{
+        atomic::{AtomicBool, Ordering},
+        mpsc::{self, Receiver},
+    },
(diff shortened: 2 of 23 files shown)


src/test_support/source_processing_state_machine/mod.rs
```diff
@@ -0,0 +1,158 @@
...
+}
+
+#[test]
+#[ignore = "explicit replay via WAVECRATE_SOURCE_STATE_MACHINE_REPLAY=<seed-or-artifact>"]
+fn source_processing_seeded_state_machine_replay() {
+    let value = std::env::var("WAVECRATE_SOURCE_STATE_MACHINE_REPLAY").expect(
+        "set WAVECRATE_SOURCE_STATE_MACHINE_REPLAY to a decimal seed or replay artifact path",
+    );
+    let (seed, events) = read_replay(&value).expect("load source state-machine replay");
+    run_or_archive(seed, events, true);
```
PORTALSURFER: Required fix: Replay does not preserve the failing execution lane. Normal/stress failures run with `with_supervisor=false`, but artifact replay always passes `true`, and the artifact schema does not record the mode. A scanner-only failure can therefore change or disappear during replay. Store the execution mode in the artifact, return it from parsing, and replay/minimize with that recorded mode; make the printed command/diagnostics unambiguous.

src/test_support/source_processing_state_machine/invariants.rs
```diff
@@ -0,0 +1,275 @@
...
+        if browser_projection != expected_projection {
+            return Err(format!(
+                "browser projection differs from committed manifest: expected={expected_projection:?}, actual={browser_projection:?}"
+            ));
+        }
+        Ok(())
+    }
+
+    pub(super) fn assert_queue_bound(&self) -> Result<(), String> {
+        if self.model.queued_causes.len() > MAX_COALESCED_CAUSES {
```
PORTALSURFER: Required fix: This bounded-queue invariant is tautological: `queued_causes` is a `BTreeSet` over exactly seven enum variants and the maximum is seven, so it cannot fail and does not observe implementation queueing. Add independent expected-work accounting plus observation of the actual supervisor/scanner pending work, then assert real coalescing/bounds.

src/test_support/source_processing_state_machine/invariants.rs
```diff
@@ -0,0 +1,275 @@
...
+                self.last_revision
+            ));
+        }
+        if revision > self.last_revision {
+            self.accepted_revisions
+                .push(format!("{}:{revision}", self.model.lifecycle_generation));
+            self.last_revision = revision;
+        }
+        if !delta.is_empty() && !publication_lost {
+            if !self.accepted_publications.insert((
```
PORTALSURFER, line 37: Required fix: This at-most-once check observes only harness bookkeeping around direct scanner commits. The supervisor is asked afterward, but no actual source-processing publication is captured or compared, so production duplicate/stale publication cannot fail the lane. Add a test observer or equivalent real publication/revision/cause observation and compare it with model expectations, including publication-loss retry.

PORTALSURFER (commented):
Independent contract/diff review: changes requested. GitHub does not permit the PR author identity to request changes on its own PR, so the blocking findings are recorded as inline `Required fix` threads.

src/test_support/source_processing_state_machine/mod.rs
PORTALSURFER: Fixed in `eb66cab50`. Replay artifacts are schema v2 and persist `ExecutionMode`; artifact replay now uses that recorded mode for both execution and shrinking. Replay also validates fixture name/version/seed. Added focused round-trip and fixture-drift regression tests.

src/test_support/source_processing_state_machine/invariants.rs
PORTALSURFER: Fixed in `eb66cab50`. The integrated lane now atomically submits each real delta three times through the supervisor's production coalescing helper, observes the actual dirty-source slot and pending identity/input sets under the control lock, and asserts one source/delta slot plus exact unique identity and revision/cause coalescing. The old seven-enum tautology was removed; scanner-only watcher paths retain an independent fixture bound.

src/test_support/source_processing_state_machine/invariants.rs
PORTALSURFER: Fixed in `eb66cab50`. A test-only observer now records the real supervisor publication/fallback output when pending deltas are consumed, including lifecycle, output generations, and coalesced input revision/cause keys. The harness asserts monotonic output revisions, at-most-once revision/cause consumption, exact expected/observed/stale accounting, and publication-loss retry. Integrated replay and 1,000-seed stress both pass.

src/test_support/source_processing_state_machine/supervisor_observer.rs
```diff
@@ -0,0 +1,260 @@
...
+                "actual supervisor publications did not account for every admitted revision/cause: expected={:?}, observed={:?}, stale={:?}",
+                self.expected_supervisor_publications,
+                self.observed_supervisor_publications,
+                self.stale_supervisor_publications
+            ));
+        }
+        Ok(())
+    }
+
+    fn account_durable_publication(&mut self) {
```
PORTALSURFER: Required fix: This credits an expected lifecycle/revision/cause key as observed solely because a later aggregate readiness snapshot has reached that revision. That can hide a missing or duplicate supervisor publication. Count only an exact supervisor-emitted observation (or an equally exact fallback record), so every admitted token must correspond to exactly one actual output.

src/test_support/source_processing_state_machine/mutations.rs
```diff
@@ -0,0 +1,262 @@
...
+        if let Some(supervisor) = &self.supervisor {
+            supervisor.replace_sources(vec![self.source.clone()])?;
+        }
+        self.model.source_configured = true;
+        self.model.lifecycle_generation = self.model.lifecycle_generation.saturating_add(1);
+        self.model.queue(ScanCause::Lifecycle);
+        Ok(())
+    }
+
+    pub(super) fn root_offline_online(&mut self) -> Result<(), String> {
```
PORTALSURFER, line 193: Required fix: The root is renamed away and restored inside the same mutation before the production supervisor is required to observe the unavailable state. Keep it unavailable through an actual integrated reconciliation, wait for the supervisor's offline/retry event, then restore it and require terminal online convergence using state-based signaling rather than sleeps.

PORTALSURFER (commented):
Fresh review of the current contract and complete diff found two required state-machine fidelity gaps.

src/test_support/source_processing_state_machine/supervisor_observer.rs
PORTALSURFER: Fixed in `74a72edaf`. Aggregate readiness inference was removed. The integrated lane now credits only exact `StateMachinePublicationObservation` records emitted by the real supervisor, and continues to reject duplicate or unadmitted lifecycle/revision/cause keys. The integrated lane and 1,000-seed stress lane pass.

src/test_support/source_processing_state_machine/mutations.rs
PORTALSURFER: Fixed in `74a72edaf`. The harness now starts the real supervisor with an event receiver. The mutation keeps the root renamed away until that supervisor emits `Offline`, then restores it, requests reconciliation, and waits for `Ready` or `DegradedTerminal` via `recv_timeout`; no sleep controls the offline/online sequence. Restart replaces the observer channel with the new supervisor. Integrated replay passes.

src/native_app/source_processing/supervisor/discovery.rs
```diff
@@ -107,12 +107,28 @@ pub(super) fn discover_candidates(
...
-                    shared.publish_source_health(health.into_event(
-                        super::SourceProcessingLifecycle::new(
-                            source.id.as_str(),
-                            in_flight_work.lifecycle_generation,
-                        ),
+                    let health = health.into_event(super::SourceProcessingLifecycle::new(
+                        source.id.as_str(),
+                        in_flight_work.lifecycle_generation,
                     ));
+                    #[cfg(test)]
```
PORTALSURFER, line 120: Required fix: The test observation is appended before `publish_source_health` and ignores whether the real sink accepted the output. A lost publication can therefore still be credited as an actual durable supervisor output. Instrument the definitive publication/fallback boundary and add deterministic failure coverage there so only successfully accepted outputs satisfy the lifecycle/revision/cause invariant.

src/test_support/source_processing_state_machine/harness.rs
```diff
@@ -0,0 +1,367 @@
...
+        self.model.queue(ScanCause::Retry);
+        self.flush(ScanCause::Retry)?;
+        self.wait_for_integrated_settle()?;
+        self.create(6, false)?;
+        self.flush(ScanCause::Foreground)?;
+        self.wait_for_integrated_settle()?;
+        self.assert_actual_publications()
+    }
+
+    fn wait_for_integrated_settle(&self) -> Result<(), String> {
```
PORTALSURFER: Required fix: Integrated convergence is established by 30-second wall-clock polling loops with 10 ms sleeps (here and in the liveness oracle), while the PR contract explicitly requires replayability without nondeterministic sleeps. Drive progress through controlled synchronization and explicit state transitions with bounded event/step accounting; keep a timeout only as a diagnostic guard.

src/test_support/source_processing_state_machine/invariants.rs
```diff
@@ -0,0 +1,275 @@
...
+                )
+            })
+            .collect::<BTreeMap<_, _>>();
+        if committed != self.model.files {
+            return Err(format!(
+                "committed manifest differs from reference filesystem model: expected={:?}, actual={committed:?}",
+                self.model.files
+            ));
+        }
+        let browser_projection = database
```
PORTALSURFER: Required fix: `browser_projection` is only a second `database.list_files()` read reduced to paths, so it cannot detect a defect in the real browser-facing projection and does not satisfy the stated DoD. Compare the production browser projection/builder against the model, or narrow the PR contract and DoD to storage-level manifest coverage.

PORTALSURFER (commented):
Fresh stateless review at `74a72edaf` found three required contract mismatches. The durable-publication mismatch has survived two fix cycles, so repository workflow requires stopping for user direction rather than continuing another unbounded fix pass.

src/native_app/source_processing/supervisor/discovery.rs
PORTALSURFER: Fixed in ec012b470. Discovery now records the exact SourceHealthPublicationOutcome only after the definitive publish_source_health_outcome boundary returns. Published and AlreadyPublished are the only accepted outputs; Rejected is tracked separately, leaves the exact pending delta inputs unconsumed, and must later resolve to one accepted output. The integrated lane includes a deterministic boundary rejection scenario and asserts exactly one injected rejection plus eventual exact publication. The focused boundary test and integrated replay pass.

src/test_support/source_processing_state_machine/harness.rs
PORTALSURFER: Fixed in ec012b470. Both 10 ms polling loops were removed. Integrated convergence now checks explicit supervisor/readiness state after each received supervisor event, with a deterministic 4,096-transition bound. The 30-second recv_timeout remains only as a deadlock diagnostic guard, not the convergence mechanism. The state-machine directory contains no sleep calls; normal and integrated lanes pass.

src/test_support/source_processing_state_machine/invariants.rs
PORTALSURFER: Fixed in ec012b470. The invariant now executes the production browser-search worker pipeline with the deterministic source and compares its actual visible relative paths against the independent reference model. The second SourceDatabase::list_files manifest reread was removed. Six normal seeds and the full 1,000-seed stress lane pass with this production projection check.

src/test_support/source_processing_state_machine/harness.rs
```diff
@@ -0,0 +1,345 @@
...
+        }
+    }
+
+    pub(super) fn flush(&mut self, cause: ScanCause) -> Result<(), String> {
+        if !self.model.root_online || !self.model.source_configured {
+            self.model.queue(ScanCause::Retry);
+            return Ok(());
+        }
+        let effective_cause = self.authoritative_cause(cause);
+        if self.take_failure(FailureBoundary::Transaction) {
```
PORTALSURFER: Required fix: the state-machine harness still consumes Transaction and WatcherDelivery before invoking the production scanner/supervisor boundary (and Lifecycle before replace_sources), so those generated steps can pass without observing a real boundary failure. Move these three injections to narrow test-only hooks or deterministic contention at the actual transaction, delta-delivery, and lifecycle-replacement boundaries; assert the concrete rejection/error, retained pending work, successful retry, and convergence.

PORTALSURFER: Fixed in e83811231. Transaction injection now holds a real SQLite BEGIN IMMEDIATE writer lock while the scanner connection uses a zero busy timeout, and the harness requires ScanError::Db(SourceDbError::Busy) before retry. Watcher delivery is rejected inside the production queue_source_delta admission helper and asserts pending scopes are unchanged before retrying the exact delta. Lifecycle removal is rejected after entering serialized replace_sources, asserts the generation remains unchanged, then retries remove/re-add. Dedicated integrated replays for all three boundaries pass, as do six normal seeds and the 1,000-sequence stress lane.

PR #906, opened 2026-07-23
OPT-1249: Bound pending rename retention and reconciliation cost

PORTALSURFER:
## Goal

Bound pending-rename metadata retention and reconciliation work to current actionable rename state without losing metadata across delayed watcher delivery, restarts, temporarily offline sources, or valid rename transfers.

Linear: OPT-1249

## Scope and non-goals

- Persist a source-local authoritative scan generation and diagnostic staging time for pending rename candidates.
- Advance retention only after a complete authoritative scan (and, for audits, successful verification); targeted, partial, canceled, unavailable-root, and failed-audit work does not age candidates.
- Prune expired candidates transactionally with generation publication, with hard-scan and unresolved-destination safeguards.
- Replace full pending-table reconciliation loads with indexed, uniqueness-bounded lookups against current destination candidates.
- Expose aggregate retention/reconciliation diagnostics and document the invariant in `docs/TARGET.md`.
- Add compatible schema repair for existing source databases and legacy read-only projections.
- Non-goals: wall-clock TTL pruning, changes to watcher delivery semantics, or broader source-database cleanup.

## Definition of Done

- Pending candidates converge after the configured authoritative-generation grace period.
- Targeted scans, incomplete traversal, cancellation, offline roots, and failed audits cannot advance pruning eligibility.
- Restart and delayed rename recovery preserve user metadata.
- Reconciliation does not materialize the historical pending source table and uses indexed lookups capped at two rows.
- Pruning and generation publication are atomic.
- Aggregate diagnostics report candidate count, generation/age, considered destinations, and pruned count.
- Migration and regression coverage exercise legacy databases, high churn, restart, rollback, offline, partial, canceled, and unresolved-destination cases.

## Validation

Passed:

- `cargo check -p wavecrate-library -p wavecrate-scan`
- `cargo test -p wavecrate-library --lib -- --test-threads=1` (247 passed)
- `cargo test -p wavecrate-scan --lib -- --test-threads=1` (128 passed)
- focused migration, read-only compatibility, retention, reconciliation, query-plan, and transaction rollback tests
- `git diff --check`

Repository-wide observations outside this diff:

- `bash scripts/ci.sh agent` reached the native app suite; 3,943 tests passed and 49 unrelated UI/cache tests failed. An isolated native-app failure reproduces with no touched files in that subsystem.
- `cargo clippy -p wavecrate-library -p wavecrate-scan --all-targets -- -D warnings` is blocked by the pre-existing `clippy::type_complexity` warning on `SourceWriteBatch::commit_with_manifest_changes`.
- Repo preflight reports five pre-existing native-app `app_chrome` boundary violations.

## Known limitations

The repository-wide native-app gate and strict clippy gate have the unrelated baseline failures recorded above. The complete affected-crate test suites pass.

### DIFF
--- a/crates/wavecrate-library/src/sample_sources/db/pending_renames.rs
+++ b/crates/wavecrate-library/src/sample_sources/db/pending_renames.rs
@@ -433,19 +433,6 @@ impl<'conn> SourceWriteBatch<'conn> {
         self.find_unique_pending_rename(TAKE_PENDING_RENAME_BY_HASH_SQL, params![hash])
     }
 
-    /// Return whether at least one retained source candidate has this content hash.
-    pub fn has_pending_rename_with_hash(&self, hash: &str) -> Result<bool, SourceDbError> {
-        self.tx
-            .query_row(
-                "SELECT EXISTS(
-                    SELECT 1 FROM pending_wav_renames WHERE content_hash = ?1 LIMIT 1
-                 )",
-                params![hash],
-                |row| row.get::<_, bool>(0),
-            )
-            .map_err(map_sql_error)
-    }
-
     /// Claim one unique retained rename candidate by stable filesystem identity.
     pub fn take_pending_rename_by_file_identity(
         &mut self,
--- a/crates/wavecrate-library/src/sample_sources/db/pending_renames.rs
+++ b/crates/wavecrate-library/src/sample_sources/db/pending_renames.rs
@@ -1,4 +1,5 @@
 use std::path::{Path, PathBuf};
+use std::time::{SystemTime, UNIX_EPOCH};
 
 use rusqlite::{OptionalExtension, params};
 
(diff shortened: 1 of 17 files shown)


crates/wavecrate-library/src/sample_sources/db/pending_renames.rs
```diff
@@ -176,11 +270,112 @@ impl<'conn> SourceWriteBatch<'conn> {
...
+        &mut self,
+        hard_scan: bool,
+    ) -> Result<PendingRenamePruneReport, SourceDbError> {
+        let generation =
+            next_pending_rename_generation(self.read_pending_rename_authoritative_generation()?);
+        let pruned = if hard_scan {
+            self.tx
+                .execute(
+                    "DELETE FROM pending_wav_renames AS pending
+                     WHERE pending.content_hash IS NULL
```
PORTALSURFER: Required fix - Apply the unresolved-destination guard to hard-scan pruning (or defer hard pruning until every current destination is resolved). This branch deletes every unmatched pending row even when `pending_wav_rename_destinations` still contains an active `retained_hash IS NULL` row; the quick branch below explicitly protects that state. A hard scan with deferred or failed destination hashing can therefore discard the sole retained user metadata before delayed recovery is possible, contrary to the PRs stated unresolved-destination safeguard. Add the equivalent active-null check to this delete path and cover the hard-scan case with a regression test.

PORTALSURFER (commented):
Review of e2b4002bf91de8193282b1d1e57e7d75b1a0eaff: one required fix remains. I cannot request changes because this identity owns the pull request.

crates/wavecrate-library/src/sample_sources/db/pending_renames.rs
PORTALSURFER: Fixed in 618d7cb4a. Hard-scan pruning now applies the unresolved-destination guard; authoritative hard scans resolve transient destinations from manifest hashes and discard proven non-matches, while unresolved destinations remain persisted whenever pending metadata remains. Validation: cargo test -p wavecrate-library --lib -- --test-threads=1 (247 passed) and cargo test -p wavecrate-scan --lib -- --test-threads=1 (126 passed).

crates/wavecrate-scan/src/sample_sources/scanner/scan_hash.rs
```diff
@@ -943,43 +861,42 @@ fn reconcile_same_file_renames(
...
         }
-        let Some(present_paths) = present_by_hash.get(hash) else {
-            continue;
-        };
-        let candidates = present_paths
+        for present_path in present_paths {
+            if reconciled_paths.contains(present_path) {
+                continue;
+            }
+            batch.retain_pending_rename_destination(present_path, hash)?;
```
PORTALSURFER, line 874: **Required fix** — This retains a destination before proving that its hash has exactly one pending source candidate. With two deleted files sharing a hash and one new file with that hash, `unique_pending_rename_by_hash` below returns `None`, but this retained destination remains. It then prevents both candidates from being pruned by later quick scans and a hard scan, so they never converge after the configured grace period. Do not let an ambiguous non-transfer match pin retention (or key retention to a proven unique pairing), and add a regression covering two pending rows plus one matching destination across the grace period and a hard rescan.

crates/wavecrate-scan/src/sample_sources/scanner/scan/runner.rs
```diff
@@ -411,34 +434,42 @@ pub(crate) fn reconcile_scan_renames(
...
+    writer: &impl ScanWriter,
+) -> Result<(), ScanError> {
+    debug_assert_ne!(mode, ScanMode::Targeted);
+    let _writer = writer.lock(super::super::scan_writer::ScanWritePhase::Manifest);
+    let mut batch = db.write_batch()?;
+    if mode == ScanMode::Hard {
+        batch.resolve_hashed_pending_rename_destinations()?;
         batch.prune_invalid_retained_rename_destinations()?;
+    }
+    let report = batch.complete_pending_rename_authoritative_scan(mode == ScanMode::Hard)?;
```
PORTALSURFER, line 452: **Required fix** — This cleanup is reachable only after all pending candidates have been pruned, but an unresolved destination prevents that prerequisite: `complete_pending_rename_authoritative_scan` defers every prune while any destination has `retained_hash IS NULL`. On a hard scan, a destination whose file has disappeared remains NULL because the new resolver updates only live, hashed `wav_files`; therefore the candidate count can never reach zero and that stale destination blocks retention convergence forever. Before the hard-scan prune decision, remove/ignore destination rows not present in the authoritative manifest while preserving genuine live deferred-hash destinations, and add a regression for a stale unresolved destination plus an expired pending candidate.

PORTALSURFER (commented):
Reviewed current head `618d7cb4a34b20fe44723c6d5fdf470e434592b0` against the PR description and complete merge-base-to-head diff. Changes requested: pending-rename retention does not converge in the two cases called out inline. (GitHub does not permit this identity to formally request changes on its own PR.)

crates/wavecrate-scan/src/sample_sources/scanner/scan_hash.rs
PORTALSURFER: Fixed in dccdbe419. Hash destinations are now retained only after the indexed LIMIT 2 lookup proves exactly one pending source. A deterministic regression covers two same-hash pending sources plus one destination and verifies hard-scan convergence without transferring either metadata record.

crates/wavecrate-scan/src/sample_sources/scanner/scan/runner.rs
PORTALSURFER: Fixed in dccdbe419. Authoritative hard cleanup now removes destination rows absent from the live manifest and live NULL destinations whose manifest hash is already known but was not retained by a unique source match. A regression seeds a stale unresolved destination and verifies it cannot block pending-source pruning. Validation: 247 wavecrate-library tests and 128 wavecrate-scan tests passed sequentially.

PORTALSURFER (commented):
Clean review — reviewed merge-base  to current head  against the PR description. The diff satisfies the stated retention, authoritative-completion, atomicity, bounded indexed reconciliation, migration/read-only compatibility, diagnostics, and regression-coverage requirements. No required fixes found.

PR #909, opened 2026-07-23
Fence deferred hash publication against stale manifests

PORTALSURFER:
## Goal

Complete the active OPT-1179 deferred-hash regression by preventing work planned from an older source manifest from overwriting newer committed state.

Linear: https://linear.app/boostnlvp/issue/OPT-1179/wavecrate-make-source-discovery-revisioned-ordered-and-self-healing

## Scope and non-goals

- Capture the manifest revision atomically with the manifest snapshot used to plan deferred hashing.
- Validate that revision inside the short deferred-hash write phase before any hash upsert or rename reconciliation.
- Return the existing typed `ScanError::StaleRevision` outcome so the source-processing owner can retry from current state.
- Add regressions for concurrent deletion, same-path replacement, and unrelated manifest revision changes.
- Keep filesystem reads and hashing outside the database writer.
- Do not redesign pending-rename matching or hold the writer during hashing.

## Definition of Done

- A deferred hash cannot recreate a row deleted after planning.
- A deferred hash cannot overwrite a replacement identity or newer content generation.
- A stale publication is reported as a typed, safely retryable outcome and is not reported as committed.
- Existing scanner rename, cancellation, content-audit, and writer-coordination coverage remains green.

## Validation

- `cargo test -p wavecrate-scan --lib` — 132 passed.
- `cargo test -p wavecrate --lib committed_file_mutations -- --nocapture` — 14 passed.
- `cargo test -p wavecrate --lib foreground_scan_admission_does_not_retain_database_writer -- --nocapture` — passed.
- `cargo test -p wavecrate --lib foreground_scan_traversal_does_not_block_independent_publication -- --nocapture` — passed.
- `bash scripts/ci.sh agent` — Wavecrate smoke and architecture gates passed; the later unchanged Radiant dependency suite failed at `gpu_signal_shader_keeps_waveform_bands_visually_distinct` because the shader no longer contains its hard-coded expected color. `vendor/radiant` is byte-for-byte unchanged from `origin/main`.

## Known limitations

- The full agent gate is blocked by the pre-existing Radiant shader assertion described above.
- No signed-app smoke was run because this follow-up is a deterministic source-database publication fence with no UI behavior.

### DIFF
--- a/crates/wavecrate-scan/src/sample_sources/scanner/scan_hash.rs
+++ b/crates/wavecrate-scan/src/sample_sources/scanner/scan_hash.rs
@@ -664,7 +664,7 @@ fn deep_hash_scan_with_post_hash_hook(
     writer: &impl ScanWriter,
     mut post_hash: impl FnMut(&std::path::Path),
 ) -> Result<ScanStats, ScanError> {
-    let manifest_before = super::manifest::capture_manifest(db)?;
+    let (manifest_revision, manifest_before) = super::manifest::capture_manifest_with_revision(db)?;
     let root = ensure_root_dir(db)?;
     let mut rename_candidates = rename_candidates.clone();
     rename_candidates.extend(db.list_pending_rename_destinations()?);
@@ -765,6 +765,12 @@ fn deep_hash_scan_with_post_hash_hook(
         return Err(ScanError::Canceled);
     }
     let mut batch = db.write_batch()?;
+    if !batch.matches_revision(manifest_revision)? {
+        return Err(ScanError::StaleRevision {
+            expected: manifest_revision,
+            actual: db.get_revision()?,
+        });
+    }
     for backfill in &hash_backfills {
         batch.upsert_file_with_hash(
             &backfill.relative_path,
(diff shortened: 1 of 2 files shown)


crates/wavecrate-scan/src/sample_sources/scanner/scan_hash.rs
```diff
@@ -765,6 +765,12 @@ fn deep_hash_scan_with_post_hash_hook(
         return Err(ScanError::Canceled);
     }
     let mut batch = db.write_batch()?;
+    if !batch.matches_revision(manifest_revision)? {
```
PORTALSURFER, line 774: Required fix — The new revision check protects this `deep_hash_scan_with_post_hash_hook` publication, but the same diff still shows `reconcile_hashed_rename_candidates_with_writer` collecting `entries_by_path` and `candidate_identities` outside `ScanWritePhase::DeferredHash` (lines 615–633), then reconciling and committing them (lines 638–653) without capturing or checking a manifest revision. A source-manifest update in that interval can therefore still apply a rename reconciliation planned from an older manifest. This contradicts the PR scope/DoD requirement to validate the planned manifest revision before **any** hash upsert or rename reconciliation. Capture the revision atomically with that function’s planning snapshot, check it immediately after `write_batch()` and before `reconcile_indexed_rename_candidates`, and return `ScanError::StaleRevision` on mismatch; add a concurrent-manifest-change regression for this path.

PORTALSURFER (commented):
Required fix: the revision fence is incomplete for the stated deferred-hash contract. GitHub does not permit this identity (the PR author) to formally request changes on its own pull request.

crates/wavecrate-scan/src/sample_sources/scanner/scan_hash.rs
PORTALSURFER: Fixed at bfcd74b66. `reconcile_hashed_rename_candidates_with_writer` now captures the planning manifest revision and identity snapshot together, validates that revision immediately after opening the write batch, and returns `ScanError::StaleRevision` before any rename reconciliation on mismatch. Added `hashed_rename_reconciliation_rejects_a_stale_manifest_revision`, which injects a concurrent manifest commit at writer admission and verifies pending rename metadata and destination metadata remain untouched. Validation: `cargo test -p wavecrate-scan --lib` (132 passed) and `bash scripts/ci.sh smoke` passed.

PORTALSURFER (commented):
Clean review — reviewed head `bfcd74b66d3e43311029349ceaf2d959294d0c95` against the current PR description and complete merge-base-to-head diff. The revision snapshot and pre-publication batch fence cover both deferred hashing and hashed-rename reconciliation; the added regressions exercise deletion, same-path replacement, unrelated revisions, and stale rename reconciliation. No Required fix findings.

PR #910, opened 2026-07-23
OPT-1247: bound source content-audit planning

PORTALSURFER:
## Goal

Make each admitted content-verification slice plan work proportional to its entry budget instead of loading and cloning the complete source manifest and audit-state map.

## Scope

- select forward and due-retry candidates through indexed, cursor-based SQLite windows with explicit limits
- preserve retry rotation, the reserved forward-progress slot, cancellation/restart checkpoints, and manifest-revision fencing
- publish exact content deltas from only the rows touched by the authoritative commit
- aggregate coverage reporting in SQLite without materializing complete manifest/state collections in Rust
- add planning telemetry and a deterministic 10,000/100,000-entry SQLite VM-instruction regression ceiling
- update the durable content-verification target contract

### Non-goals

- no change to the 30-day coverage objective, hash time/byte ceilings, watcher verification, or lifecycle audit admission
- no filesystem hashing inside a database transaction
- no unrelated native-app or Radiant cleanup

## Definition of Done

- [x] A one-entry slice does not load or clone every manifest/audit-state row.
- [x] Retry and forward selection are bounded, cursor-based, and index-backed.
- [x] Retry fairness, forward progress, cancellation/restart, unstable-file, and stale-revision behavior remain covered.
- [x] Exact committed deltas remain fenced to the revalidated filesystem snapshot and authoritative database commit.
- [x] 10,000- and 100,000-entry profiles stay within the same explicit 5,000 SQLite VM-instruction ceiling.

## Validation

- `cargo test -p wavecrate-library --lib` — 248 passed
- `cargo test -p wavecrate-scan --lib` — 134 passed
- `cargo test -p wavecrate-scan --lib sample_sources::scanner::scan_hash::tests` — 24 passed
- `bash scripts/ci.sh agent` — smoke/build and architecture checks passed; the gate later stopped on a pre-existing Radiant shader-color assertion: `gpu_signal_shader_keeps_waveform_bands_visually_distinct` expects `vec4<f32>(0.655, 0.231, 0.188, 0.88)` (1,686 other Radiant tests passed)

## Known limitations

The repository-wide agent gate is not fully green because of the unrelated Radiant shader assertion above. Signed-app browsing/audition smoke was not run because this change is database planning/delta logic with no UI or playback-path change.

Linear: OPT-1247


crates/wavecrate-library/src/sample_sources/db/content_audit.rs
```diff
@@ -241,41 +241,211 @@ impl SourceDatabase {
...
+                (
+                    format!(
+                        "SELECT wav.path, wav.file_identity, wav.content_hash,
+                                wav.file_size, wav.modified_ns
+                         FROM source_content_audit_entries AS audit
+                              INDEXED BY idx_source_content_audit_retry_path
+                         CROSS JOIN wav_files AS wav ON wav.path = audit.path
+                         WHERE audit.path {comparison} ?1
+                           AND audit.skip_reason IS NOT NULL
+                           AND (audit.retry_at IS NULL OR audit.retry_at <= ?2)
```
PORTALSURFER: **Required fix — retry planning is still unbounded when due entries are sparse or absent.** The forced `idx_source_content_audit_retry_path` index is keyed only by `path`, while the due predicate is applied afterward. With 100,000 skipped rows whose `retry_at` values are all in the future (or with one due row late in cursor order), SQLite must walk every matching path-range row before `LIMIT ?3` can return/establish that no retry is due. That violates the PR goal/DoD that each admitted slice and the 10k/100k profiles stay bounded by the entry budget.

Use a plan/index strategy that bounds lookup of *due* retry rows while preserving the durable retry rotation, and add 10k/100k VM-instruction coverage for no-due and sparse-due retry populations. The current test makes every 32nd retry due, so it cannot expose this scan.

crates/wavecrate-library/src/sample_sources/db/content_audit.rs
```diff
@@ -241,41 +241,211 @@ impl SourceDatabase {
...
+                (
+                    format!(
+                        "SELECT wav.path, wav.file_identity, wav.content_hash,
+                                wav.file_size, wav.modified_ns
+                         FROM wav_files AS wav
+                              INDEXED BY idx_source_content_audit_forward_path
+                         LEFT JOIN source_content_audit_entries AS audit
+                                ON audit.path = wav.path
+                         WHERE wav.path {comparison} ?1
+                           AND {filter}
```
PORTALSURFER: **Required fix — the forward window can scan the whole non-audio portion of the manifest.** `idx_source_content_audit_forward_path` is only `(path) WHERE missing = 0` and is forced here, so `{filter}` on the supported extension is evaluated after the cursor path scan. In a source containing many non-missing unsupported/index-only entries ahead of the next supported file (or no supported file after the cursor), a one-entry window walks all of them before `LIMIT ?3` applies. This is not work proportional to the entry budget and can exceed the required 5,000 VM-instruction ceiling.

Make the cursor scan selective for supported-audio rows as well as `missing = 0` (without losing path order/wrap semantics), and extend the 10k/100k VM test with a selective supported-audio population so this regression is covered.

PORTALSURFER (commented):
Required fixes identified. I could not submit `REQUEST_CHANGES` because GitHub does not allow the authenticated author to request changes on their own pull request. The two inline comments are review-blocking.

crates/wavecrate-library/src/sample_sources/db/content_audit.rs
PORTALSURFER: Fixed in c3982530f. Retry planning now uses the partial expression index `idx_source_content_audit_retry_due` on `(COALESCE(retry_at, 0), path)` and reads only a `2 * retry_limit` due-time frontier before applying durable path-cursor rotation in memory. The regression profile now covers 10k/100k populations with one sparse due retry and with no due retries; both remain under the 5,000 SQLite VM-instruction ceiling. Validation: `cargo test -p wavecrate-library --lib`, `cargo test -p wavecrate-scan --lib`, and `bash scripts/ci.sh smoke` passed.

crates/wavecrate-library/src/sample_sources/db/content_audit.rs
PORTALSURFER: Fixed in c3982530f. Forward planning now forces `idx_source_content_audit_forward_path`, a partial path index whose predicate exactly includes live supported-WAV rows and excludes AppleDouble sidecars, so unsupported rows are absent before the cursor seek and `LIMIT`. The regression profile now uses 10k/100k manifests with one supported row and all remaining rows unsupported, forcing wraparound without scanning the unsupported population; both remain under the 5,000 SQLite VM-instruction ceiling. Validation: `cargo test -p wavecrate-library --lib`, `cargo test -p wavecrate-scan --lib`, and `bash scripts/ci.sh smoke` passed.

crates/wavecrate-library/src/sample_sources/db/content_audit.rs
```diff
@@ -241,41 +241,174 @@ impl SourceDatabase {
...
+        limit: usize,
+        after_cursor: bool,
+    ) -> Result<Vec<SourceManifestEntry>, SourceDbError> {
+        let comparison = if after_cursor { ">" } else { "<=" };
+        let filter = qualified_supported_audio_filter("wav");
+        let sql = format!(
+            "SELECT wav.path, wav.file_identity, wav.content_hash,
+                    wav.file_size, wav.modified_ns
+             FROM wav_files AS wav
+                  INDEXED BY idx_source_content_audit_forward_path
```
PORTALSURFER, line 390: **Required fix — sparse eligible populations still make planning proportional to the manifest.** This index only filters `wav_files` by supported/missing/path. The `audit.skip_reason` and verified-snapshot predicates below are applied after each indexed `wav` row is joined, so a source whose supported rows are already verified or skipped must scan the entire remaining (and then wrapped) forward index before returning the empty window. The retry query has the inverse issue at lines 282–289: its audit-only due index cannot exclude due entries whose joined manifest row is unsupported or missing, so a large due-but-ineligible population is scanned before it can find an eligible retry or conclude there is none. That violates the goal that one-entry planning be proportional to its entry budget and the DoD’s bounded/index-backed selection under sparse due/unsupported populations.

Please change the stored/queryable eligibility model so each forward and retry window can seek eligible candidates without traversing the noneligible population, while preserving cursor/fairness semantics. Add 10,000- and 100,000-entry VM-instruction cases for (1) supported rows made non-forward-eligible by current audit state and (2) due audit rows made ineligible by missing/unsupported manifest rows, and assert the explicit 5,000-instruction ceiling for both sizes (the existing absolute check at lines 855–859 covers only the 100,000-entry measurements).

PORTALSURFER (commented):
Review outcome: changes requested. GitHub does not allow this identity to request changes on its own pull request; the required inline finding records the blocking contract mismatch.

crates/wavecrate-library/src/sample_sources/db/content_audit.rs
PORTALSURFER: Fixed in 41b44d81d by removing the unbounded eligibility search from both planners. Forward selection now returns a raw supported-row cursor window capped by `LIMIT`; each row carries verified/retry disposition, so the scanner either hashes it or advances the durable cursor without rehashing. Retry selection now returns a due-time frontier capped at `2 * retry_limit`; live supported rows are retried, while missing/unsupported rows are retired from audit state in the same revision-fenced bounded batch. Thus noneligible populations are drained as bounded checkpoint/cleanup work rather than traversed before `LIMIT`, preserving cursor and retry fairness without an O(N) eligibility probe. Regression coverage now profiles 10k/100k verified-or-skipped supported rows and 10k/100k due-but-stale retry rows, and asserts the 5,000 VM-instruction ceiling for every 10k and 100k measurement. Added scanner tests prove checkpoint-only forward convergence and bounded stale-retry retirement. Validation: library 248/248, scan 134/134, and `bash scripts/ci.sh smoke` passed.

PORTALSURFER (commented):
Reviewed current head `41b44d81ded169260b0bfcb51528b7e6b999f78f` against the PR description and complete merge-base-to-head diff.

Clean: the diff meets the stated goal and Definition of Done. The cursor/index planning remains bounded across forward, due-retry, no-due, verified/skipped, and stale-retry paths; revision fencing and exact authoritative-commit delta handling are preserved. Focused regression suites passed: `cargo test -p wavecrate-library --lib content_audit` and `cargo test -p wavecrate-scan --lib sample_sources::scanner::scan_hash::tests`.

PR #911, opened 2026-07-24
OPT-1259 Isolate process-wide test runtime state

PORTALSURFER:
## Goal

Make Wavecrate tests that touch process-wide environment, HOME/profile roots, configuration-root state, or the process working directory share one panic-safe isolation contract.

Linear: [OPT-1259](https://linear.app/boostnlvp/issue/OPT-1259/wavecrate-tests-isolate-process-wide-environment-and-working-directory)

## Scope and non-goals

- Add a reusable, re-entrant, poison-safe `TestRuntimeGuard` under the existing `test-support` feature.
- Restore exact `OsString` environment values and the prior working directory in reverse mutation order, including unwind paths.
- Fail closed and reject later scoped access if the prior working directory becomes unrestorable.
- Migrate config-home, HOME/profile, SQLite-extension, token-store, updater, and logging test owners from module-local locks/manual cleanup.
- Keep `ConfigBaseGuard` thread-local instead of temporarily clearing the process-global app-root override.
- Clear the fallback-key cache mutex poison flag after recovery so later scoped owners can reuse it.
- Do not serialize the entire library suite or change production environment/path resolution.

## Definition of Done

- Direct Wavecrate test-side environment/CWD mutation is centralized in `TestRuntimeGuard`.
- Cross-module contenders in one test binary cannot overlap.
- Nested, unset/empty, panic cleanup, failed-CWD fail-closed behavior, concurrent contention, poison recovery, and cross-thread config-root visibility have regression coverage.
- Test roots used by migrated owners remain explicit temporary roots.
- Production builds do not expose the test-runtime module unless `test-support` is enabled.

## Validation

Passed:

- `git diff --check`
- `cargo check -p wavecrate-library`
- `cargo check -p wavecrate`
- `cargo test -q -p wavecrate-library app_dirs::tests -- --test-threads=16` (9 passed)
- `cargo test -q -p wavecrate-library test_runtime -- --test-threads=16` (5 passed)
- `cargo test -q -p wavecrate-library sqlite_ext::policy::tests --features sqlite-ext-unsafe -- --test-threads=16` (5 passed)
- `cargo test -q -p wavecrate-library source_db_mod_tests::opening::user_library -- --test-threads=16` (4 passed)
- `cargo test -q -p wavecrate issue_gateway::token_store::tests --lib -- --test-threads=16` (19 passed; repeated five consecutive times)
- `cargo test -q -p wavecrate updater::path_guard::tests --lib -- --test-threads=16` (5 passed)
- `cargo test -q -p wavecrate app_core::controller::tests::dispatch::logging --lib -- --test-threads=16` (1 passed)
- `cargo test -p wavecrate --test controller_browser_integration --features legacy-controller click_clears_selection_and_focuses_row -- --exact --test-threads=1` (1 passed)
- `bash scripts/ci.sh agent`: Wavecrate smoke and architecture gates passed before the limitation below.

## Known limitations

- Two requested full `cargo test -p wavecrate --lib -- --test-threads=16` runs completed with 54 and 56 failures respectively. The sets were dominated by existing load-sensitive native-app/source-watcher/cache tests and included a deterministic source-shape assertion (`root_facade_exports_only_stable_entrypoints`); none of the migrated process-state owner families failed.
- `bash scripts/ci.sh agent` is currently blocked in the vendored Radiant suite by `gpu_signal_shader_keeps_waveform_bands_visually_distinct`, which expects a stale shader color literal. The preceding Wavecrate smoke and architecture gates passed.

### DIFF
--- a/crates/wavecrate-library/src/test_runtime.rs
+++ b/crates/wavecrate-library/src/test_runtime.rs
@@ -1,0 +1,282 @@
+//! Scoped isolation for tests that mutate process-global runtime state.
+
+use std::{
+    ffi::{OsStr, OsString},
+    marker::PhantomData,
+    path::{Path, PathBuf},
+    rc::Rc,
+    sync::{Condvar, Mutex, MutexGuard, OnceLock},
+    thread::{self, ThreadId},
+};
+
+static PROCESS_STATE_LOCK: OnceLock<ReentrantProcessLock> = OnceLock::new();
+
+#[derive(Default)]
+struct LockState {
+    owner: Option<ThreadId>,
+    depth: usize,
+}
+
+#[derive(Default)]
+struct ReentrantProcessLock {
+    state: Mutex<LockState>,
+    available: Condvar,
+}
+
+impl ReentrantProcessLock {
+    fn acquire(&self) {
+        let current = thread::current().id();
+        let mut state = lock_unpoisoned(&self.state);
+        loop {
+            match state.owner.as_ref() {
+                None => {
+                    state.owner = Some(current);
+                    state.depth = 1;
+                    return;
+                }
+                Some(owner) if *owner == current => {
+                    state.depth += 1;
+                    return;
+                }
+                Some(_) => {
+                    state = self
+                        .available
+                        .wait(state)
+                        .unwrap_or_else(|poisoned| poisoned.into_inner());
+                }
+            }
+        }
+    }
+
+    fn release(&self) {
+        let current = thread::current().id();
+        let mut state = lock_unpoisoned(&self.state);
+        assert_eq!(
+            state.owner.as_ref(),
+            Some(&current),
+            "process test runtime guard dropped by a non-owner thread"
+        );
+        state.depth -= 1;
+        if state.depth == 0 {
+            state.owner = None;
+            self.available.notify_one();
+        }
+    }
+}
+
+fn lock_unpoisoned<T>(mutex: &Mutex<T>) -> MutexGuard<'_, T> {
+    mutex
+        .lock()
+        .unwrap_or_else(|poisoned| poisoned.into_inner())
+}
+
+enum Restoration {
+    Environment {
+        key: OsString,
+        previous: Option<OsString>,
+    },
+    CurrentDirectory(PathBuf),
+}
+
+/// Serializes and restores process-global environment and working-directory changes.
+///
+/// The lock is shared by every Wavecrate test owner in the current test binary,
+/// is re-entrant on one thread, and recovers from mutex poisoning. Each guard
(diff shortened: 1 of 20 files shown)


crates/wavecrate-library/src/test_runtime.rs
```diff
@@ -0,0 +1,282 @@
...
+                        current_dir_restore_error = Some((previous, error));
+                    }
+                }
+            }
+        }
+        process_state_lock().release();
+
+        if let Some((path, error)) = current_dir_restore_error {
+            if thread::panicking() {
+                eprintln!(
```
PORTALSURFER: **Required fix**: If restoring the saved CWD fails while unwinding, this branch only logs the error and the guard has already released `PROCESS_STATE_LOCK` (line 162). A test that catches the original panic can therefore let the next owner acquire the guard while the process is still in the mutated directory. That contradicts the stated requirement to restore the prior CWD on unwind and makes subsequent scoped owners unsafe. Make an unrecoverable CWD restoration failure fail closed—e.g. terminate the test process, or record a fatal guard state that makes all later `acquire` calls fail—rather than continuing after `eprintln!`; add a regression that removes/unmakes the saved directory during a panicking scoped CWD change.

PORTALSURFER (commented):
Required fix identified: the guard’s CWD-restore failure path does not preserve the isolation contract during unwind.

crates/wavecrate-library/src/test_runtime.rs
PORTALSURFER: Fixed in 04b7209c0. A failed CWD restoration now marks the shared runtime lock permanently unavailable before unwinding continues, wakes waiters so they fail immediately, and makes every later acquisition panic instead of entering with leaked process state. The regression deletes the saved directory during a panicking scoped change, catches the original panic, verifies reacquisition is rejected, and confirms the outer guard restores the actual test CWD. Validation: `cargo fmt --all -- --check`; `cargo test -p wavecrate-library test_runtime -- --test-threads=16` (5 passed); `git diff --check`.

PORTALSURFER (commented):
Reviewed current head `04b7209c09f87b30d2cbae18c7bae14ecf47dbd5` against the PR description and complete merge-base diff only. **Clean** — the diff achieves the OPT-1259 goal and Definition of Done: it centralizes the migrated process-wide environment/CWD owners behind a re-entrant poison-safe guard; restores exact `OsString` values and CWD in reverse mutation order; fails closed after a CWD-restore failure; preserves thread-local config-base isolation without clearing the global root for other threads; clears fallback-cache poison after recovery; and keeps the module behind test/test-support compilation gates. No required fixes or non-blocking observations.

PR #912, opened 2026-07-24
OPT-1258 Scope and reset mutable global test controls

PORTALSURFER:
## Goal

Make mutable Wavecrate test controls lifecycle-owned so hooks, caches, counters, and configuration overrides cannot outlive the test that installed them or affect unrelated parallel tests.

## Scope

- Replace source/folder move worker hook globals with explicit injected hook contexts; production workers use zero-state no-op hooks.
- Add owned, unwind-safe scopes for source write-priority/remap state, delete-journal failure injection, source-DB open counters, token fallback state, startup-audio stubs, and debug logging.
- Inject updater metadata behavior directly into the path guard test seam instead of retaining a process-global hook.
- Keep environment-derived telemetry and configuration `OnceLock`s inert under libtest while preserving production cache lifetimes.
- Remove unused decode counters and the ambient global provenance mirror; assert against the emitted tracing event instead.

## Non-goals

- No production runtime reset API or production cache-lifetime change.
- No serialization of the complete test suite.
- No changes to environment/current-directory isolation owned by OPT-1259.
- No fixes for unrelated native-app fixture, waveform-cache, watcher, UI contract, or vendored Radiant failures already present at current `main`.

## Definition of Done

- Mutable test controls changed here have explicit owners and deterministic cleanup.
- Panic/unwind paths restore prior state.
- Test hooks are supplied to the worker that consumes them and are visible across worker-thread boundaries.
- Scoped counters and one-shot failures clean only their owned path/source state.
- Production worker and telemetry paths gain no test-only allocation or locking.

## Validation

Passed:

- `git diff --check`
- `cargo check -p wavecrate --tests --bins`
- `cargo test -p wavecrate-library -- --test-threads=16` — 256 passed; process-boundary helper 1 passed/1 ignored
- Focused Wavecrate owner suites for source/folder workers, source priority/remaps, delete-journal failure injection, updater path guard, startup audio, logging, fallback warning/cache, and source-DB counters — all passed
- `bash scripts/ci.sh smoke` (as part of the agent gate)

Observed current-head limitations:

- Repeated `cargo test -p wavecrate --lib -- --test-threads=16` runs completed with 53 and 54 failures respectively versus a 54-failure pre-change baseline. Changed-scope owner and worker tests pass; the remaining failures are existing native-app fixture/waveform-cache/watcher/UI-contract families, including cases reproducible individually or in isolated modules.
- `bash scripts/ci.sh agent` passes branch guard, smoke, non-blocking architecture, and readiness boundaries, then stops in the pinned Radiant library suite at `gpu_signal_shader_keeps_waveform_bands_visually_distinct` (1 failed, 1,686 passed). This branch does not modify Radiant.

## Known limitations

The current `main` full parallel Wavecrate lane and vendored Radiant suite are not green independently of these scoped-control changes. This PR records the exact baseline instead of masking or retrying those failures.

### DIFF
--- a/crates/wavecrate-library/src/sample_sources/db/open.rs
+++ b/crates/wavecrate-library/src/sample_sources/db/open.rs
@@ -227,10 +227,34 @@ impl SourceDatabaseOpenMode {
     }
 }
 
-/// Reset the debug-only source DB open counter used by regression tests.
+/// Owned scope for one source DB open counter used by regression tests.
 #[cfg(debug_assertions)]
-pub fn test_reset_source_db_open_total_count(root: &Path) {
+pub struct TestSourceDbOpenCountGuard {
+    root: std::path::PathBuf,
+}
+
+#[cfg(debug_assertions)]
+impl Drop for TestSourceDbOpenCountGuard {
+    fn drop(&mut self) {
+        telemetry::reset_open_total_count(&self.root);
+    }
+}
+
+#[cfg(debug_assertions)]
+impl TestSourceDbOpenCountGuard {
+    /// Clear observations accumulated so far while retaining scope ownership.
+    pub fn reset(&self) {
+        telemetry::reset_open_total_count(&self.root);
+    }
+}
+
+/// Clear one source DB open counter for a scope and clear it again on drop.
+#[cfg(debug_assertions)]
+pub fn test_scope_source_db_open_total_count(root: &Path) -> TestSourceDbOpenCountGuard {
     telemetry::reset_open_total_count(root);
+    TestSourceDbOpenCountGuard {
+        root: root.to_path_buf(),
+    }
 }
 
 /// Return the debug-only source DB open count used by regression tests.
--- a/crates/wavecrate-library/src/sample_sources/db/open.rs
+++ b/crates/wavecrate-library/src/sample_sources/db/open.rs
@@ -236,7 +236,7 @@ pub struct TestSourceDbOpenCountGuard {
 #[cfg(debug_assertions)]
 impl Drop for TestSourceDbOpenCountGuard {
     fn drop(&mut self) {
-        telemetry::reset_open_total_count(&self.root);
+        telemetry::release_open_total_count_scope(&self.root);
     }
 }
 
@@ -251,7 +251,7 @@ impl TestSourceDbOpenCountGuard {
 /// Clear one source DB open counter for a scope and clear it again on drop.
 #[cfg(debug_assertions)]
 pub fn test_scope_source_db_open_total_count(root: &Path) -> TestSourceDbOpenCountGuard {
-    telemetry::reset_open_total_count(root);
+    telemetry::acquire_open_total_count_scope(root);
     TestSourceDbOpenCountGuard {
         root: root.to_path_buf(),
     }
--- a/src/app/controller/library/source_write_priority.rs
+++ b/src/app/controller/library/source_write_priority.rs
@@ -137,17 +137,60 @@ struct SourceWritePriorityState {
 struct SourceWritePriorityTestScope {
     source_id: SourceId,
     previous: SourceWritePriorityState,
-    _owner: std::sync::MutexGuard<'static, ()>,
+    _owner: SourceWritePriorityTestOwner,
+}
+
+#[cfg(test)]
+struct SourceWritePriorityTestOwner {
+    source_id: SourceId,
+}
+
+#[cfg(test)]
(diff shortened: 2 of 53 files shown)


crates/wavecrate-library/src/sample_sources/db/open.rs
```diff
@@ -227,10 +227,34 @@ impl SourceDatabaseOpenMode {
...
+    /// Clear observations accumulated so far while retaining scope ownership.
+    pub fn reset(&self) {
+        telemetry::reset_open_total_count(&self.root);
+    }
+}
+
+/// Clear one source DB open counter for a scope and clear it again on drop.
+#[cfg(debug_assertions)]
+pub fn test_scope_source_db_open_total_count(root: &Path) -> TestSourceDbOpenCountGuard {
     telemetry::reset_open_total_count(root);
```
PORTALSURFER: 

PORTALSURFER (commented):
Independent diff review: required fix noted inline.

crates/wavecrate-library/src/sample_sources/db/open.rs
PORTALSURFER: Fixed in f291000a4. Source-DB counter scopes now acquire exclusive ownership per source root through a mutex/condition-variable state; acquisition atomically clears the owner count, drop atomically clears and releases only that root, and poisoned locks recover safely. Added concurrent_same_root_open_total_count_scopes_are_isolated, which proves a second owner waits and starts with an independent count. Validation: cargo test -p wavecrate-library -- --test-threads=16 (256 passed; process-boundary test passed; 1 helper ignored), cargo check -p wavecrate --tests --bins, cargo fmt --all -- --check, git diff --check.

src/app/controller/library/source_write_priority.rs
```diff
@@ -122,3 +125,144 @@ impl Drop for FileOpWritePriorityGuard {
...
+struct SourceWritePriorityTestScope {
+    source_id: SourceId,
+    previous: SourceWritePriorityState,
+    _owner: std::sync::MutexGuard<'static, ()>,
+}
+
+#[cfg(test)]
+impl SourceWritePriorityTestScope {
+    fn new(source_id: &SourceId) -> Self {
+        static OWNER: OnceLock<Mutex<()>> = OnceLock::new();
```
PORTALSURFER: **Required fix** — `OWNER` is a single process-wide mutex and its guard is retained by `SourceWritePriorityTestScope` for the entire test scope. Consequently, a scope for source A blocks a simultaneous `FileOpWritePriorityGuard` or `CompletedBrowserRenameTestGuard` for unrelated source B, even though `take_source_state`/`restore_source_state` operate per `SourceId`. This contradicts the PR goal that controls not affect unrelated parallel tests and its explicit non-goal of serializing the suite.

Make ownership exclusive per `SourceId` (while preserving unwind restoration) so only competing scopes for the same source wait; scopes for distinct sources must proceed concurrently. Add a regression that demonstrates both same-source exclusion and distinct-source concurrent acquisition.

PORTALSURFER (commented):
Changes requested (GitHub does not permit this identity to request changes on its own PR): the inline required fix below blocks a clean review.

src/app/controller/library/source_write_priority.rs
PORTALSURFER: Fixed in 2c8d64aa0. SourceWritePriorityTestScope now owns a source-keyed RAII lease backed by an active SourceId set and condition variable. Same-source scopes wait; distinct-source scopes acquire without retaining the global state mutex. Ownership releases after state restoration, including unwind. Added same_source_test_scopes_are_exclusive and distinct_source_test_scopes_acquire_concurrently. Validation: cargo test -p wavecrate --lib source_write_priority::tests -- --test-threads=16 (4 passed), cargo check -p wavecrate --tests --bins, cargo fmt --all -- --check, git diff --check.

PORTALSURFER (commented):
Clean review — reviewed head `2c8d64aa0e29d0dc6a6907c494d96776b3dbf8b3` against the current PR description and complete merge-base-to-head diff.

The diff achieves the stated goal and Definition of Done: mutable test controls are lifecycle-owned with unwind cleanup; keyed counter/source state excludes same-key owners without serializing distinct keys; hooks are explicitly supplied to their consuming workers; libtest leaves environment-derived caches inert; and production worker/telemetry paths retain zero-state/no-op or production-only behavior.

No required fixes. Verdict: **clean**.

PR #913, opened 2026-07-24
OPT-1262 Persist typed index-only source entries

PORTALSURFER:
## Goal

Persist typed index-only records for unsupported and inaccessible source files so classification and diagnostics survive restart and later format-policy changes can promote files through normal reconciliation.

Closes OPT-1262.

## Scope and non-goals

- Add typed unsupported-audio, unsupported-non-audio, inaccessible, and practically-unsupported classifications with bounded diagnostics and format-policy versioning.
- Add a versioned `source_index_entries` schema, independent index revision, and read snapshot/list APIs.
- Reconcile index-only create/change/move/delete in full scans and targeted sync, while preserving uncertain subtrees and promoting newly supported paths without inheriting sample metadata.
- Retain user metadata but remove an inaccessible supported path from the live manifest/readiness set until recovery.
- Exclude Wavecrate database files, links/reparse targets, and AppleDouble sidecars, including when type/metadata probing fails.
- Do not add codecs, playback/editing behavior, or an `all files` UI.

## Definition of Done

- [x] Unsupported audio and non-audio persist as distinct index-only classifications across restart.
- [x] Inaccessible/unclassifiable observations persist typed diagnostics without paths outside the source root.
- [x] Supported manifest rows remain the only entries eligible for sample metadata and readiness work.
- [x] Full and targeted create/change/move/delete reconciliation and format-policy promotion have regression coverage.
- [x] Existing source databases migrate without changing manifest revision or user metadata, with rollback/failure coverage.
- [x] Public read APIs expose typed index-only snapshots and path-scoped entries.

## Validation

- `cargo test -p wavecrate-library` — pass (266 unit tests plus process-boundary readiness test)
- `cargo test -p wavecrate-scan --lib` — pass (144 tests)
- `bash scripts/ci.sh smoke` — pass
- `cargo fmt --all -- --check` — pass
- `git diff --check` — pass
- `bash scripts/ci.sh agent` — repository baseline blocker: the pinned, unchanged Radiant test `gpu_signal_shader_keeps_waveform_bands_visually_distinct` expects a shader literal absent from the pinned shader. The remaining broad root library suite also has existing parallel/global-state failures; the deferred native-open failure was reproduced unchanged at `main` (`4421be0eb`) in a clean detached worktree.

## Known limitations

- The future `all files`/diagnostics UI remains out of scope.
- The current format policy defines the practically-unsupported state but has no active WAV practical limit to assign it during traversal.

### DIFF
--- a/crates/wavecrate-scan/src/sample_sources/scanner/scan_fs.rs
+++ b/crates/wavecrate-scan/src/sample_sources/scanner/scan_fs.rs
@@ -14,13 +14,15 @@ use wavecrate_library::filesystem_identity::{
     filesystem_change_marker, stable_filesystem_identity,
 };
 use wavecrate_library::sample_sources::{
-    SourceEntryClassification, SourceEntryFileType, classify_source_entry,
+    SourceEntryClassification, SourceEntryFileType, SourceFileClassification,
+    SourceIndexDiagnostic, SourceIndexEntry, classify_source_entry,
 };
 
 use crate::sample_sources::SourceDatabase;
 
 use super::scan::ScanError;
 use super::scan::{SourceTreeFile, SourceTreeSnapshot};
+use super::scan_index::{inaccessible_index_entry, index_entry_from_file_facts};
 
 const MAX_LAYOUT_DIAGNOSTICS: usize = 16;
 
@@ -29,6 +31,7 @@ thread_local! {
     static FORCED_DIRECTORY_READ_FAILURES: RefCell<BTreeSet<PathBuf>> = const { RefCell::new(BTreeSet::new()) };
     static FORCED_DIRECTORY_ENTRY_FAILURES: RefCell<BTreeSet<PathBuf>> = const { RefCell::new(BTreeSet::new()) };
     static FORCED_FILE_TYPE_FAILURES: RefCell<BTreeSet<PathBuf>> = const { RefCell::new(BTreeSet::new()) };
+    static FORCED_FILE_METADATA_FAILURES: RefCell<BTreeSet<PathBuf>> = const { RefCell::new(BTreeSet::new()) };
 }
 
 /// Deterministically emulate an unreadable directory for scanner regression
@@ -52,12 +55,19 @@ pub(crate) fn force_file_type_failure(path: &Path) -> ForcedTraversalFailure {
     ForcedTraversalFailure::new(path, ForcedFailureKind::FileType)
 }
 
+/// Deterministically emulate an unreadable file-metadata observation.
+#[cfg(test)]
+pub(crate) fn force_file_metadata_failure(path: &Path) -> ForcedTraversalFailure {
+    ForcedTraversalFailure::new(path, ForcedFailureKind::FileMetadata)
+}
+
 #[cfg(test)]
 #[derive(Clone, Copy)]
 enum ForcedFailureKind {
     DirectoryRead,
     DirectoryEntry,
     FileType,
+    FileMetadata,
 }
 
 #[cfg(test)]
@@ -94,6 +104,7 @@ fn forced_failures(
         ForcedFailureKind::DirectoryRead => &FORCED_DIRECTORY_READ_FAILURES,
         ForcedFailureKind::DirectoryEntry => &FORCED_DIRECTORY_ENTRY_FAILURES,
         ForcedFailureKind::FileType => &FORCED_FILE_TYPE_FAILURES,
+        ForcedFailureKind::FileMetadata => &FORCED_FILE_METADATA_FAILURES,
     }
 }
 
@@ -128,6 +139,18 @@ pub(super) fn forced_file_type_error(path: &Path) -> Option<std::io::Error> {
             std::io::Error::new(
                 std::io::ErrorKind::PermissionDenied,
                 "forced file type failure",
+            )
+        })
+    })
+}
+
+#[cfg(test)]
+fn forced_file_metadata_error(path: &Path) -> Option<std::io::Error> {
+    FORCED_FILE_METADATA_FAILURES.with(|failures| {
+        failures.borrow().contains(path).then(|| {
+            std::io::Error::new(
+                std::io::ErrorKind::PermissionDenied,
+                "forced file metadata failure",
             )
         })
     })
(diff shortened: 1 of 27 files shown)


crates/wavecrate-scan/src/sample_sources/scanner/scan_index.rs
```diff
@@ -0,0 +1,98 @@
...
+        .keys()
+        .filter(|path| {
+            live_manifest_paths.contains(*path)
+                || (!observed.contains_key(*path) && !context.preserves_missing_row(path))
+        })
+        .cloned()
+        .collect::<Vec<_>>();
+    let upserts = observed
+        .into_values()
+        .filter(|entry| !live_manifest_paths.contains(&entry.relative_path))
```
PORTALSURFER: Required fix (P1): This filter discards a typed `Inaccessible` observation whenever the same supported `.wav` still has a live manifest row. Targeted sync marks the path uncertain, so the supported row is preserved, then this line drops the only durable diagnostic; full scan similarly routes supported paths directly through manifest preparation without an inaccessible fallback. That contradicts the DoD requiring inaccessible/unclassifiable observations to persist. Represent supported-path availability diagnostics independently (or transition the manifest row to a non-live state), apply it consistently in full and targeted reconciliation, keep unavailable paths out of readiness eligibility, and cover failure plus recovery for a supported `.wav`.

PORTALSURFER (commented):
Independent review of the PR contract and complete diff: changes requested. The connected GitHub identity is the PR author, so this is recorded as a comment review rather than a formal request-changes state.

crates/wavecrate-scan/src/sample_sources/scanner/scan_index.rs
PORTALSURFER: Fixed in c3098427f. Inaccessible observations for live supported paths now atomically mark the retained manifest row missing and persist the typed diagnostic, removing the path from the live manifest/readiness set without losing user metadata. Full-scan preparation failures and targeted no-follow open/metadata failures use the same transition; successful reconciliation restores the manifest row and removes the diagnostic. Added full and targeted failure/recovery regressions, including metadata preservation. Validation: 142 scanner tests and 265 library tests pass; smoke, formatting, and diff checks pass.

crates/wavecrate-scan/src/sample_sources/scanner/scan_fs.rs
```diff
@@ -239,6 +262,12 @@ pub(super) fn visit_dir_with_cancel_check(
                         format!("read file type {}: {err}", display_relative(root, &path)),
                     );
                     record_uncertain_prefix(&mut snapshot, root, &path);
+                    if let Ok(relative_path) = path.strip_prefix(root) {
```
PORTALSURFER: Required fix (P1): This failure branch persists `Inaccessible` before applying the name-based exclusions. A failed probe for `.wavecrate.db`, `.wavecrate_samples.db` (including SQLite sidecars), or an `._*` AppleDouble sidecar can therefore become durable index state despite the explicit PR exclusion. The same bypass exists in targeted directory and direct-target metadata failure paths. Factor the path-only exclusions into a shared predicate, apply it before every inaccessible fallback, and add full plus targeted forced-failure regressions.

PORTALSURFER (commented):
Fresh review of the current PR contract and complete diff: one required exclusion fix remains. The connected identity is the PR author, so this is recorded as a comment review.

crates/wavecrate-scan/src/sample_sources/scanner/scan_fs.rs
PORTALSURFER: Fixed in 731b25fa5. The shared entry policy now exports a path-only rejection predicate for Wavecrate database artifacts, SQLite sidecars, and AppleDouble names. Full traversal, targeted directory traversal, and direct targeted metadata-failure paths apply it before creating any inaccessible record. Added forced full/targeted probe-failure coverage for `.wavecrate.db` and `._sidecar.flac`. Validation: 143 scanner tests and 266 library tests pass; smoke, formatting, and diff checks pass.

PORTALSURFER (commented):
Persistence/migration specialist review found one uncertainty-preservation edge case. The connected identity is the PR author, so this is recorded as a comment review.

crates/wavecrate-scan/src/sample_sources/scanner/scan_fs.rs
```diff
@@ -226,9 +249,16 @@ pub(super) fn visit_dir_with_cancel_check(
...
 
             let path = entry.path();
+            let relative = path
+                .strip_prefix(root)
+                .map(Path::to_path_buf)
+                .map_err(|_| ScanError::InvalidRoot(path.clone()))?;
             let file_type = match read_file_type(&entry, &path) {
                 Ok(file_type) => file_type,
                 Err(err) => {
+                    if is_rejected_source_file_path(&relative) {
```
PORTALSURFER, line 259: Required fix (P1): This path-only exclusion runs without knowing whether the entry is a file or directory. If a legitimate directory named `._assets` or nested `.wavecrate.db` has a transient type-probe failure, this early `continue` records no uncertain prefix; reconciliation can then delete its retained manifest/index descendants. Targeted failure branches have the same issue. Suppress the inaccessible row for reserved names, but still mark the path uncertain until a successful probe proves it is an excluded file. Add full and targeted regressions that preserve descendants below reserved-name directories.

PORTALSURFER: Fixed in 9f43cd39e. Reserved-name probe failures now suppress inaccessible index rows while still recording an uncertain prefix in full traversal, targeted directory traversal, and direct targeted metadata lookup. Added full and targeted regressions proving supported and index-only descendants below `._assets` and nested `.wavecrate.db` directories are retained. Validation: 144 scanner tests and 266 library tests pass; smoke, formatting, and diff checks pass.

PORTALSURFER (commented):
Final review: clean at `9f43cd39e09b2e137bf1f3d3404f671755545f9c`. The current complete diff meets the PR Goal and Definition of Done. Ordinary behavioral and persistence/migration specialist passes found no remaining required fixes; all required review threads are resolved.

PR #914, opened 2026-07-24
OPT-1260: Add parallel test-isolation stress lane

PORTALSURFER:
## Goal

Add a repeatable opt-in CI lane that proves the Wavecrate library harness runs in fresh, explicitly parallel processes and produces durable failure evidence for shared-state flakes.

## Scope and non-goals

- Add `scripts/ci.* isolation-stress` with bounded repetitions, explicit parallelism, and per-process timeouts.
- Compile once, launch a fresh test-binary process per iteration, and stop at the first unexpected failure without retrying until green.
- Emit JSONL metadata, detector-verification, iteration, and summary records with the test binary, first test identity, failure class counts, full failure inventory, and known-quarantine disposition.
- Verify deliberately injected environment/current-directory and mutable-global-control leaks in isolated sentinel processes.
- Detect timeouts and leaked owned worker processes on POSIX process groups and Windows Job Objects, and keep the normal `agent` lane unchanged.
- Do not fix or mask the shared-state failures owned by OPT-1258/OPT-1259 or newly exposed by this lane.

## Definition of Done

- [x] Documented Bash and PowerShell commands run a bounded, explicitly parallel, fresh-process lane.
- [x] Every launched process emits a machine-readable JSONL record and preserves the first failure identity.
- [x] Injected process-state and mutable-global-control leaks are detected before stress iterations start.
- [x] Runner tests cover parsing, fresh-process invocation, timeout exit behavior, leak classification, failed summaries, and leaked workers.
- [x] Normal agent-lane intent and command remain unchanged.
- [ ] Repeated full parallel library runs pass: current `main` still reproduces 49 unexpected shared-state-sensitive failures, which the new lane reports and blocks.

## Validation

Passed:

- `python3 -m unittest -v scripts/internal/ci/test_parallel_isolation_stress.py` — 9 tests: 8 passed on macOS and the Windows Job Object runtime test was skipped off-platform.
- `cargo test -p wavecrate --lib parallel_isolation_sentinel -- --test-threads=4` — 2 passed.
- `bash scripts/check.sh script-guardrails`.
- `bash scripts/check.sh docs-index`.
- `cargo fmt --all -- --check`.
- `git diff --check`.
- Python syntax checks for the runner, Windows Job Object helper, and runner tests; JSON syntax checks for the updated inventories.
- POSIX timeout regression fixture proved a child that ignores `SIGTERM` and holds the output pipe is killed within the bounded cleanup window.
- Failed stress runs now emit a terminal JSONL summary with completed iterations, first failure, failure-family counts, full inventory, known quarantine coverage, and report path.

Expected blocking evidence:

- `bash scripts/ci.sh isolation-stress --iterations 1 --test-threads 4 --timeout-seconds 900 --output target/test-isolation-stress/opt-1260-fix-validation.jsonl` — both injected detector checks passed; the stress iteration failed with 49 unexpected tests and emitted the complete failed summary. First failure: `native_app::app_chrome::library_browser::library_sidebar::folder_tree::status::tests::selected_folder_status_paints_missing_source_as_warning`.
- A direct Cargo comparison with the same filters and parallelism reproduced the same 49 failures, confirming the runner did not create the failure family.
- `bash scripts/ci.sh agent` progressed through smoke and architecture checks, then failed in the pinned Radiant suite at `gpu_signal_shader_keeps_waveform_bands_visually_distinct`; this is outside the Wavecrate diff.
- `bash scripts/check.sh file-size-budget` could not run because the existing helper reads an empty Bash array under `set -u` at line 219.

## Known limitations

- The newly durable gate exposes that the current head still does not satisfy the issue's repeated-green requirement; the failures are intentionally not quarantined or retried away.
- PowerShell syntax and dispatcher contracts were checked, and Windows process ownership has a Windows-only Job Object regression test, but the lane was not executed on a Windows host in this change.

### DIFF
--- a/scripts/internal/ci/parallel_isolation_stress.py
+++ b/scripts/internal/ci/parallel_isolation_stress.py
@@ -1,0 +1,573 @@
+#!/usr/bin/env python3
+"""Run the Wavecrate library test harness repeatedly in fresh parallel processes."""
+
+from __future__ import annotations
+
+import argparse
+import json
+import os
+import re
+import signal
+import subprocess
+import sys
+import time
+from collections import Counter
+from dataclasses import asdict, dataclass
+from datetime import datetime, timezone
+from pathlib import Path
+from typing import TextIO
+
+
+SCHEMA_VERSION = 1
+PROCESS_LEAK_ENV = "WAVECRATE_ISOLATION_INJECT_PROCESS_LEAK"
+GLOBAL_HOOK_LEAK_ENV = "WAVECRATE_ISOLATION_INJECT_GLOBAL_HOOK_LEAK"
+INJECTION_ENV_VARS = (PROCESS_LEAK_ENV, GLOBAL_HOOK_LEAK_ENV)
+PROCESS_SENTINEL = (
+    "test_isolation_sentinels::"
+    "parallel_isolation_sentinel_process_state_guard_under_contention"
+)
+GLOBAL_HOOK_SENTINEL = (
+    "app::controller::library::source_write_priority::tests::"
+    "parallel_isolation_sentinel_scoped_global_control_lifecycle_under_contention"
+)
+QUARANTINED_TESTS = (
+    {
+        "test_name": "prepare_auto_rename_requests_logs_looped_provenance",
+        "disposition": "known_quarantine",
+        "reason": "excluded from the agent-safe and isolation-stress library lanes",
+    },
+    {
+        "test_name": "rating_previous_random_history_entry_restores_waveform_for_replacement",
+        "disposition": "known_quarantine",
+        "reason": "excluded from the agent-safe and isolation-stress library lanes",
+    },
+)
+FAILURE_MARKERS = {
+    "WAVECRATE_ISOLATION:process_state_contamination": "process_state_contamination",
+    "WAVECRATE_ISOLATION:mutable_global_control_leak": "mutable_global_control_leak",
+    "WAVECRATE_ISOLATION:leaked_worker": "leaked_worker",
+}
+FAILED_TEST_RE = re.compile(r"^test (?P<name>.+) \.\.\. FAILED(?: .*)?$")
+FAILURE_SECTION_RE = re.compile(
+    r"^---- (?P<name>[^\r\n]+) stdout ----\r?\n"
+    r"(?P<body>.*?)(?=^---- [^\r\n]+ stdout ----\r?\n|^failures:\r?\n|\Z)",
+    re.MULTILINE | re.DOTALL,
+)
+
+
+@dataclass(frozen=True)
+class Failure:
+    """One classified test-harness failure."""
+
+    test_name: str | None
+    failure_class: str
+    disposition: str
+    evidence: str
+
+
+@dataclass(frozen=True)
+class ProcessResult:
+    """Result of one fresh test-binary process."""
+
+    status: str
+    exit_code: int | None
+    duration_ms: int
+    failures: list[Failure]
+    output: str
+
+
+def parse_args() -> argparse.Namespace:
+    parser = argparse.ArgumentParser(description=__doc__)
+    parser.add_argument(
+        "--iterations",
+        type=lambda value: bounded_int(value, minimum=1, maximum=100),
+        default=5,
+        help="Fresh library-test processes to run, 1-100 (default: 5).",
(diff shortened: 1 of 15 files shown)


scripts/internal/ci/parallel_isolation_stress.py
```diff
@@ -0,0 +1,573 @@
...
+        stdout=subprocess.PIPE,
+        stderr=subprocess.STDOUT,
+        text=True,
+        **popen_kwargs,
+    )
+    try:
+        output, _ = process.communicate(timeout=timeout_seconds)
+    except subprocess.TimeoutExpired as timeout:
+        terminate_process_tree(process)
+        remainder, _ = process.communicate()
```
PORTALSURFER: Required fix: POSIX timeout cleanup can block indefinitely after the test leader exits while a child ignores SIGTERM and retains the output pipe. Verify the entire owned process group is gone after a bounded grace period, SIGKILL survivors even when the leader already exited, bound final output collection, and add a regression fixture.

scripts/internal/ci/parallel_isolation_stress.py
```diff
@@ -0,0 +1,573 @@
...
+        process.wait(timeout=1)
+    except subprocess.TimeoutExpired:
+        try:
+            os.killpg(process.pid, signal.SIGKILL)
+        except ProcessLookupError:
+            pass
+
+
+def process_group_has_survivors(process_group: int) -> bool:
+    if os.name == "nt":
```
PORTALSURFER: Required fix: the public PowerShell lane unconditionally reports no worker survivors on Windows, so a successful test leader can leak descendants without failing the lane. Add Windows-owned-process tracking and cleanup with Windows-specific regression coverage.

scripts/internal/ci/parallel_isolation_stress.py
```diff
@@ -0,0 +1,573 @@
...
+                result_payload(
+                    phase="stress",
+                    iteration=iteration,
+                    iterations=args.iterations,
+                    test_binary=test_binary,
+                    test_threads=args.test_threads,
+                    result=result,
+                ),
+            )
+            if result.status != "passed":
```
PORTALSURFER, line 649: Required fix: detector and stress failures return before emitting the promised JSONL summary record. Emit a failed summary containing completed iterations, first failure, class counts, full inventory, and quarantine disposition before every non-zero return.

scripts/internal/ci/parallel_isolation_stress.py
PORTALSURFER: Fixed in dedd92097. POSIX cleanup now checks the entire process group, escalates remaining members to SIGKILL after a bounded grace period, and bounds final output collection. The regression fixture starts a child that ignores SIGTERM and retains stdout; it verifies timeout cleanup completes in under four seconds.

scripts/internal/ci/parallel_isolation_stress.py
PORTALSURFER: Fixed in dedd92097. Windows launches through an ownership handshake into a Job Object, queries active descendants, terminates the owned job on timeout or leaks, and reports ownership failures. The leaked-worker fixture now runs cross-platform, with an additional Windows-only Job Object assertion.

scripts/internal/ci/parallel_isolation_stress.py
PORTALSURFER: Fixed in dedd92097. Every detector or stress failure now emits a terminal failed summary with completed iterations, exit code, first failure identity/class, class counts, the complete failure inventory, known-quarantine coverage, and report path. The real 49-failure run and a focused unit test verify it.

PORTALSURFER:
## Fresh review — Required fix (contract)

Reviewed head `dedd920977cea32f2614182d78081a966604bdca` against the current PR description and complete merge-base diff.

The implementation correctly fails and reports the 49 unexpected parallel-suite failures. However, the PR contract simultaneously:

- excludes fixing or masking those shared-state failures; and
- requires repeated full parallel library runs to pass.

No implementation can satisfy both conditions as written. Before approval, choose one contract direction: make durable detection/reporting and blocking the acceptance criterion for this PR, or explicitly expand scope to resolve the failures needed for a green repeated run.

No additional code-level required fixes or non-blocking observations were found.

PR #915, opened 2026-07-24
OPT-1264 Relaunch after committed updater cleanup warnings

PORTALSURFER:
## Goal

Ensure a successfully committed Wavecrate update still attempts relaunch when disposable `.old`/`.new` cleanup fails, while preserving fatal rollback semantics before commit.

Linear: [OPT-1264](https://linear.app/boostnlvp/issue/OPT-1264/wavecrate-updater-relaunch-after-a-committed-update-when-backup)

## Scope and non-goals

- Return a typed transaction outcome that distinguishes a clean commit from a committed update with cleanup failures.
- Carry affected remnant paths and errors through `ApplyPlan`, progress warnings, and updater-helper diagnostics.
- Continue relaunch after post-commit cleanup warnings; keep relaunch failures independently fatal.
- Retry safe `.old`/`.new` cleanup before staging the same validated destination on a later update run.
- Remove remnant symlinks without following them and document the validated-root retry contract.
- Keep pre-commit swap failures, rollback failures, archive/release selection, and stable-release policy unchanged.

## Definition of Done

- [x] Post-commit cleanup failure returns a warning-bearing committed outcome instead of a fatal apply error.
- [x] The newly installed files remain active and relaunch is attempted.
- [x] Helper diagnostics include each affected cleanup path and error, including when relaunch then fails.
- [x] Pre-commit failures still roll back and remain fatal.
- [x] Safe next-run remnant cleanup and symlink behavior have deterministic coverage.
- [x] Existing updater path-guard and rollback coverage remains green.

## Validation

Passed:

- `cargo test -p wavecrate updater::fs_ops --lib` — 7 passed
- `cargo test -p wavecrate updater::apply --lib` — 14 passed
- `cargo test -p wavecrate updater` — 71 passed
- `cargo test -p wavecrate-updater-helper` — 11 passed
- `cargo fmt --all`
- `git diff --check`
- `bash scripts/ci.sh agent` passed branch policy, smoke compilation, non-blocking architecture, and readiness boundary checks before reaching the unrelated Radiant baseline failure described below.

## Known limitations

- The full agent gate stops in the pinned Radiant library suite after 1,686 passing tests because `gpu_signal_shader_keeps_waveform_bands_visually_distinct` fails against the current baseline shader; OPT-1264 does not modify Radiant.
- Request preflight also reports five pre-existing native-app palette boundary violations outside updater scope.
- Focused clippy is blocked by pre-existing warnings in `wavecrate-library` and Radiant.
- A Windows cross-check was attempted but this macOS host lacks the MSVC C sysroot/headers required by `ring` and bundled SQLite. The live Windows locked-backup-path exercise remains outstanding.
- The user's untracked `vendor/radiant/docs/DESIGN_DIRECTION.md` is preserved and excluded from this PR.

### DIFF
--- a/apps/updater-helper/src/main.rs
+++ b/apps/updater-helper/src/main.rs
@@ -27,6 +27,9 @@ fn try_main() -> Result<(), String> {
 
 fn run_headless(args: UpdaterRunArgs) -> Result<(), String> {
     let plan = run_headless_with(args, apply_update)?;
+    for warning in cleanup_warning_messages(&plan) {
+        eprintln!("{warning}");
+    }
     eprintln!(
         "Updated {} from {} into {}",
         APP_NAME,
@@ -34,6 +37,24 @@ fn run_headless(args: UpdaterRunArgs) -> Result<(), String> {
         plan.install_dir.display()
     );
     Ok(())
+}
+
+fn cleanup_warning_messages(plan: &ApplyPlan) -> Vec<String> {
+    let committed_cleanup = plan.post_commit_cleanup_failures.iter().map(|failure| {
+        format!(
+            "Warning: update committed but cleanup left {}: {}",
+            failure.path.display(),
+            failure.error
+        )
+    });
+    let stale_cleanup = plan.stale_removal_failures.iter().map(|failure| {
+        format!(
+            "Warning: failed to remove stale path {}: {}",
+            failure.path.display(),
+            failure.error
+        )
+    });
+    committed_cleanup.chain(stale_cleanup).collect()
 }
 
 fn run_headless_with(
--- a/apps/updater-helper/src/main.rs
+++ b/apps/updater-helper/src/main.rs
@@ -6,8 +6,8 @@
 use std::path::PathBuf;
 
 use wavecrate::updater::{
-    APP_NAME, ApplyPlan, REPO_SLUG, RuntimeIdentity, UpdateChannel, UpdateError, UpdaterRunArgs,
-    apply_update, supported_release_target_for_platform_arch,
+    APP_NAME, ApplyPlan, REPO_SLUG, RuntimeIdentity, UpdateChannel, UpdateError, UpdateProgress,
+    UpdaterRunArgs, apply_update_with_progress, supported_release_target_for_platform_arch,
 };
 
 #[cfg(test)]
@@ -26,10 +26,11 @@ fn try_main() -> Result<(), String> {
 }
 
 fn run_headless(args: UpdaterRunArgs) -> Result<(), String> {
-    let plan = run_headless_with(args, apply_update)?;
-    for warning in cleanup_warning_messages(&plan) {
-        eprintln!("{warning}");
-    }
+    let plan = run_headless_with(
+        args,
+        |args, progress| apply_update_with_progress(args, progress),
+        |warning| eprintln!("{warning}"),
+    )?;
     eprintln!(
         "Updated {} from {} into {}",
         APP_NAME,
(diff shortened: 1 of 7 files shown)


apps/updater-helper/src/main.rs
```diff
@@ -27,6 +27,9 @@ fn try_main() -> Result<(), String> {
 
 fn run_headless(args: UpdaterRunArgs) -> Result<(), String> {
     let plan = run_headless_with(args, apply_update)?;
+    for warning in cleanup_warning_messages(&plan) {
```
PORTALSURFER: **Required fix** — These warnings are emitted only after `run_headless_with` returns `Ok(plan)`. In the new `finish_applied_update` path, a committed cleanup failure is recorded and then a relaunch failure returns `Err` (`src/updater/apply.rs`, new lines 235–258), so `?` exits before this loop and the helper emits none of the affected cleanup paths/errors. That misses the PR’s Definition of Done that helper diagnostics include each cleanup path and error, while the same PR requires relaunch failures to remain fatal. Preserve and emit the committed-cleanup warnings on the relaunch-error path as well (without making relaunch success non-fatal), and add coverage for cleanup failure plus relaunch failure together.

PORTALSURFER (commented):
Required fix identified. I could not submit a formal changes request because this GitHub identity is the PR author.

apps/updater-helper/src/main.rs
PORTALSURFER: Fixed in `4aa4c5288`. The helper now runs the apply path with progress and forwards detailed warnings immediately, before relaunch can fail; relaunch errors remain fatal. Added combined cleanup-warning/relaunch-error coverage in both the apply layer and updater helper. Validation: `cargo test -p wavecrate updater::apply --lib` (14 passed), `cargo test -p wavecrate updater` (69 passed), and `cargo test -p wavecrate-updater-helper` (11 passed).

src/updater/fs_ops.rs
```diff
@@ -185,14 +235,29 @@ fn rollback_entry(entry: &CommittedEntry) -> Result<(), UpdateError> {
...
 
 fn remove_path_if_exists(path: &Path, kind: StagedKind) -> Result<(), UpdateError> {
-    if !path.exists() {
+    let metadata = match fs::symlink_metadata(path) {
+        Ok(metadata) => metadata,
+        Err(err) if err.kind() == std::io::ErrorKind::NotFound => return Ok(()),
+        Err(err) => return Err(err.into()),
+    };
+    if metadata.file_type().is_symlink() {
+        fs::remove_file(path)?;
```
PORTALSURFER: **Required fix**

The updater helper is a Windows path, but `std::fs::remove_file` does not remove a directory symlink on Windows. A prior `resources.old` or `resources.new` directory symlink therefore returns an error here; after a post-commit warning, every next run still fails before staging that same validated destination, so the promised retry does not work. Use a non-dereferencing unlink that handles both file and directory symlinks (for example, fall back to the directory-link removal operation) and add coverage for the directory-symlink case on the supported platform.

PORTALSURFER (commented):
Review of 4aa4c5288f075f02cf1194fc1462c355908a957e: one required fix is noted inline. GitHub does not permit this identity to request changes on its own pull request.

src/updater/fs_ops.rs
PORTALSURFER: Fixed in `be880b8e2`. Windows now uses `FileTypeExt::is_symlink_dir()` to select `remove_dir` for directory links and `remove_file` for file links; Unix continues unlinking symlinks with `remove_file`. Both paths are non-recursive and never follow the target. Added directory-symlink target-preservation coverage plus deterministic directory-unlink selection coverage. Validation: `cargo test -p wavecrate updater::fs_ops --lib` (7 passed) and `cargo test -p wavecrate updater` (71 passed).

PORTALSURFER (commented):
Clean review — reviewed current head `be880b8e265ee150b5426d5d8bc53b4c2def8fb5` against the PR description and complete merge-base-to-head diff. The diff achieves the stated goal and Definition of Done: post-commit cleanup failures retain per-path diagnostics without suppressing relaunch; relaunch failures remain fatal; pre-commit rollback behavior stays fatal; subsequent staging retries validated `.old`/`.new` cleanup; and symlink removal unlinks rather than follows targets on Unix and Windows. No required fixes.

PR #921, opened 2026-07-24
OPT-1263 restore strict manifest commit Clippy lane

PORTALSURFER:
## Goal

Restore the denied-warning Clippy lane for Wavecrate manifest commit results without changing transaction or revision semantics.

## Scope

- Replace the nested manifest commit result tuple with the documented `ManifestCommitResult` type.
- Use named fields for revision, touched-path changes, and the optional authoritative snapshot.
- Apply the same result type to the bounded manifest commit helper so the full library lane remains warning-free.
- Update scanner consumers and cover both incremental and authoritative-snapshot branches.
- Remove an existing `useless_format` warning in the affected Wavecrate library package.

## Non-goals

- No Clippy allowance or lint suppression.
- No source discovery, hashing, or manifest reconciliation redesign.
- No unrelated Radiant changes.

## Definition of Done

- Manifest commit results are named, documented, and consumed through named fields.
- Incremental and authoritative snapshot semantics remain unchanged and are covered by tests.
- The focused Wavecrate library Clippy lane passes with `-D warnings`.
- Existing Wavecrate source database and scan tests pass.

## Validation

- `cargo test -p wavecrate-library` — 267 passed.
- `cargo test -p wavecrate-scan` — 163 passed.
- `cargo clippy -p wavecrate-library --all-targets --all-features -- -D warnings` — passed.
- `bash scripts/ci.sh agent` — reached the unrelated Radiant test suite, where `gpu_signal_shader_keeps_waveform_bands_visually_distinct` failed on the existing expected shader color assertion; no vendor files were changed.

## Known limitations

The repository-wide agent gate remains blocked by the unrelated Radiant shader-color assertion described above.

### DIFF
--- a/crates/wavecrate-library/src/sample_sources/db/write/transaction.rs
+++ b/crates/wavecrate-library/src/sample_sources/db/write/transaction.rs
@@ -79,11 +79,11 @@ impl SourceWriteBatch<'_> {
 
     /// Commit the batch and return its exact revision plus manifest state owned by that revision.
     ///
-    /// When the caller's cached revision is current, the second tuple element contains only
-    /// touched paths and the optional full snapshot is `None`, keeping chunked scans linear. When
-    /// another writer has advanced the manifest, the touched-path list is empty and the optional
-    /// full snapshot is captured inside this committing transaction before the write lock is
-    /// released.
+    /// When the caller's cached revision is current, `touched_path_changes` contains only touched
+    /// paths and `authoritative_snapshot` is `None`, keeping chunked scans linear. When another
+    /// writer has advanced the manifest, `touched_path_changes` is empty and
+    /// `authoritative_snapshot` contains the full manifest captured inside this committing
+    /// transaction before the write lock is released.
     pub fn commit_with_manifest_changes(
         self,
         expected_previous_revision: u64,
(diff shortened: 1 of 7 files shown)


crates/wavecrate-library/src/sample_sources/db/write/transaction.rs
```diff
@@ -77,14 +87,7 @@ impl SourceWriteBatch<'_> {
...
         expected_previous_revision: u64,
-    ) -> Result<
-        (
-            u64,
-            Vec<(PathBuf, Option<SourceManifestEntry>)>,
-            Option<Vec<SourceManifestEntry>>,
-        ),
-        SourceDbError,
-    > {
+    ) -> Result<ManifestCommitResult, SourceDbError> {
```
PORTALSURFER, line 90: Required fix: the surrounding documentation still refers to the “second tuple element,” but this signature now returns `ManifestCommitResult`. Update the comment to describe `touched_path_changes` and `authoritative_snapshot` by name, including their incremental versus stale-revision behavior, so the public API documentation matches the implementation.

PORTALSURFER (commented):
Required fix identified during independent review. GitHub does not permit the PR author to request changes on their own pull request; I will address this finding in the next fix commit.

PR #922, opened 2026-07-24
Fix Windows clipboard owner and initialization cleanup

PORTALSURFER:
Fixes OPT-1265.

## Goal

Make the compatibility Windows clipboard writer use a valid non-null owner and release the clipboard lock on every post-open failure.

## Scope

- Resolve the current UI-thread active window as the clipboard owner and pass it to `OpenClipboard`.
- Construct the clipboard close guard immediately after a successful open, before `EmptyClipboard`.
- Preserve existing `HGLOBAL` transfer and cleanup behavior.
- Add injected lifetime tests for initialization failure and normal post-initialization drop.
- Document the Windows clipboard ownership and cleanup contract.

## Non-goals

- No changes to Radiant's native clipboard service.
- No drag/drop redesign.
- No retry loops or Linux clipboard work.

## Definition of Done

- Windows clipboard writes never call `OpenClipboard(None)` and reject a missing active owner.
- `CloseClipboard` runs exactly once after every successful open, including `EmptyClipboard` failure.
- Existing file and text payload paths retain their ownership semantics.

## Validation

- `cargo check --lib` — passed on macOS.
- `cargo test -p wavecrate external_clipboard` — passed (3 tests on macOS; Windows-only guard tests are cfg-gated).
- `cargo test -p wavecrate --lib` — 3974 passed, 55 unrelated existing failures, 27 ignored.
- `cargo check --lib --target x86_64-pc-windows-msvc` — blocked by the host's missing MSVC/Windows C toolchain headers in `ring` and SQLite, before Wavecrate diagnostics.

## Known limitations

Windows Explorer/Notepad smoke testing requires a Windows host and was not available in this environment.

### DIFF
--- a/docs/TARGET.md
+++ b/docs/TARGET.md
@@ -1503,6 +1503,11 @@ Wavecrate should support clipboard-based audio handoff.
 
 When the user selects one or more supported sample files in the sample browser and presses copy, Wavecrate should place those files on the system clipboard in a format that DAWs and Explorer can consume as ordinary audio files.
 
+On Windows, the compatibility clipboard writer must open the clipboard with the
+current Wavecrate UI thread's active window as a non-null owner. It must close
+the clipboard on every post-open failure, including an `EmptyClipboard` failure,
+before returning the error.
+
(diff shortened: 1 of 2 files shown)


PORTALSURFER (commented):
Review: clean. The current diff satisfies OPT-1265: it rejects a null active-window owner, passes a concrete owner to OpenClipboard, constructs the close guard before EmptyClipboard, preserves HGLOBAL transfer semantics, and includes focused failure/lifetime coverage. No required fixes or non-blocking observations.

PR #926, opened 2026-07-24
Recover readiness retries and expired leases after restart

PORTALSURFER:
## Goal

Recover durable readiness retries and expired leases immediately after a Wavecrate restart.

## Scope

- Add a bounded, source-indexed readiness work aggregate for pending, due/future retries, and running lease deadlines.
- Make startup and periodic safety probes consult durable work before taking the cheap no-op path.
- Schedule the earliest future retry or lease deadline without materializing file targets.
- Preserve claim-generation and unexpired-lease ownership semantics, including compatibility-null leases and legacy retry classifications.

Non-goals: no manifest traversal for deadline discovery, retry-policy redesign, startup reset of running jobs, high-frequency polling of unavailable sources, or changes to generation/claim fencing.

## Definition of Done

- Due retryable work and expired or compatibility-reclaimable leases become actionable after restart.
- Unexpired leases remain untouched and produce a wake at their persisted expiry.
- Future retry deadlines produce a coordinator wake no later than the deadline.
- Fully converged sources retain the bounded cheap no-op path.
- Regression coverage exercises future/due retries, future/expired/null leases, and restart probe behavior.

## Validation

- `cargo test -p wavecrate-library --lib sample_sources::readiness::tests::` — 47 passed.
- `cargo test -p wavecrate --lib native_app::source_processing::supervisor::tests::` — 97 passed, 1 ignored.
- `cargo test -p wavecrate --lib safety_probe_recovers_durable_retry_and_lease_deadlines_after_restart` — passed.
- `bash scripts/ci.sh agent` — stopped at an unrelated pre-existing vendor/radiant shader assertion: `gpu_signal_shader_keeps_waveform_bands_visually_distinct`; all changed files are outside `vendor/`.

## Known limitations

The full agent gate remains blocked by the unrelated existing vendor shader test failure. No task-local test failure was observed.

Related issue: OPT-1274

### DIFF
--- a/src/native_app/source_processing/supervisor/tests/retirement_recovery.rs
+++ b/src/native_app/source_processing/supervisor/tests/retirement_recovery.rs
@@ -179,6 +179,201 @@ fn discovery_progress_converged_safety_probe_is_a_silent_noop() {
 }
 
 #[test]
+fn safety_probe_recovers_durable_retry_and_lease_deadlines_after_restart() {
+    let (_directory, source) = unhashed_source("durable-readiness-deadlines");
+    let database_root = source.database_root().expect("database root");
+    let mut connection = SourceDatabase::open_connection_with_role_and_database_root(
+        &source.root,
+        &database_root,
+        SourceDatabaseConnectionRole::JobWorker,
+    )
+    .expect("open source database");
+    publish_current_readiness_targets(&mut connection, source.id.as_str(), 100)
+        .expect("publish readiness targets");
+    let snapshot = reconcile_readiness(&connection, source.id.as_str(), 100)
+        .expect("reconcile readiness targets");
+    let target = snapshot
+        .entries
+        .iter()
+        .find(|entry| entry.target.stage == ReadinessStage::IndexedIdentity)
+        .expect("indexed identity target")
+        .target
+        .clone();
+    persist_readiness_deficits(&mut connection, &snapshot.deficits, 100)
+        .expect("persist readiness work");
+    connection
+        .execute(
+            "UPDATE analysis_jobs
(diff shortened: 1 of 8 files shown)


src/native_app/source_processing/supervisor/tests/retirement_recovery.rs
```diff
@@ -178,6 +178,201 @@ fn discovery_progress_converged_safety_probe_is_a_silent_noop() {
...
+    let connection = SourceDatabase::open_connection_with_role_and_database_root(
+        &source.root,
+        &database_root,
+        SourceDatabaseConnectionRole::JobWorker,
+    )
+    .expect("reopen source database");
+    let claim_generation_before = claim.claim_generation;
+    connection
+        .execute(
+            "UPDATE analysis_jobs SET lease_expires_at = 99
```
PORTALSURFER, line 282: Required fix: this regression covers a numerically expired running lease, but not a compatibility-null running lease. Add a restarted-probe case that keeps `status = 'running'`, sets `lease_expires_at = NULL`, and asserts the probe bypasses cheap no-op and produces candidate work.

PORTALSURFER (commented):
Required fix identified during independent review. I will add the missing compatibility-null running-lease regression in the next fix commit.

src/native_app/source_processing/supervisor/tests/retirement_recovery.rs
PORTALSURFER: Fixed in ce244ddcd. The restart regression now keeps the job running, sets `lease_expires_at = NULL`, and asserts the safety probe produces candidate work without taking the cheap no-op path. The focused test passes.

PR #928, opened 2026-07-24
fix(source-processing): park duplicate hard-link readiness identities

PORTALSURFER:
## Goal

Stop unchanged hard-linked manifest identities from entering the five-second full-discovery retry loop.

## Scope

- Detect duplicate stable filesystem identities among supported live manifest rows before readiness publication.
- Persist a bounded, identity-membership-revision-keyed terminal diagnostic with affected relative paths.
- Advance that revision for identity, live-row, missing-state, and supported-eligibility mutations.
- Publish actionable terminal source health with no retry deadline and skip unchanged safety reconciliation.
- Recheck after a supported identity membership change or explicit manifest audit.
- Preserve the existing retry behavior for transient discovery/database failures.
- Document the readiness identity policy and add a hard-link regression test covering diagnosis, retry suppression, missing-state repair, same-path identity repair, and recovery.

## Non-goals

- No symlink following.
- No content/size/mtime-based identity collapsing.
- No redesign of per-path analysis/browser projections.
- No change to transient readiness retry policy.

## Definition of Done

- Unchanged duplicate identities do not hot-loop full reconciliation.
- The diagnostic is bounded and contains affected relative paths.
- Live-state or identity changes clear the diagnosis and resume readiness discovery.
- Ordinary transient discovery failures remain scheduled for retry.
- Other sources can continue through the coordinator without a duplicate-identity source retaining work ownership.

## Validation

- `cargo test -p wavecrate --lib native_app::source_processing::supervisor::tests::hard_link_identity_is_parked_without_retry_until_manifest_revision_changes -- --exact --nocapture` — passed, including missing-state repair and same-path alias replacement.
- `cargo test -p wavecrate-library --lib sample_sources::db::write::tests::revision -- --nocapture` — 5 passed.
- `cargo test -p wavecrate --lib native_app::source_processing::supervisor::tests -- --skip periodic_manifest_audit_wakes_browser_projection_after_committed_repair` — 97 passed, 1 ignored.
- `cargo test -p wavecrate-scan --all-targets` — 165 passed.
- `cargo fmt --all -- --check` and `git diff --check` — passed.
- `bash scripts/ci.sh agent` — repository guardrails and readiness boundary passed; reached vendor/Radiant tests and failed on the pre-existing shader assertion `gpu_signal_shader_keeps_waveform_bands_visually_distinct` (1690 passed, 1 failed).
- The full supervisor filter also reproduces the pre-existing `periodic_manifest_audit_wakes_browser_projection_after_committed_repair` failure at `progress_lifecycle.rs:371`; it is unchanged by this PR.

## Known limitations

Per-path alias-aware readiness execution and browser projection remain a follow-up. Until that model exists, duplicate stable identities are surfaced as an actionable terminal source state rather than silently collapsing one visible path.

### DIFF
--- a/src/native_app/source_processing/supervisor/discovery_reconcile.rs
+++ b/src/native_app/source_processing/supervisor/discovery_reconcile.rs
@@ -1,11 +1,12 @@
 use super::{
     AtomicBool, Cancellable, DiscoveryProgressUpdate, MANIFEST_AUDIT_INTERVAL_SECONDS,
-    META_LAST_MANIFEST_AUDIT_AT, META_WAV_PATHS_REVISION, PendingReadinessDelta, ProcessingLane,
-    ReadinessClassification, ReadinessProgress, ReadinessStore, ReadinessView, RuntimeCandidate,
-    RuntimeTask, SampleSource, SourceAvailability, SourceDiscoveryPhase, SourceDiscoveryStats,
-    SourceHealthSummary, WorkCandidate, active_recording_deferrals, cancelled,
-    clear_duplicate_identity_diagnostic, duplicate_identity_diagnostic_for_revision,
-    earliest_deadline, find_duplicate_identity_diagnostics, legacy_unsupported_decode_failure_text,
+    META_LAST_MANIFEST_AUDIT_AT, META_WAV_IDENTITIES_REVISION, META_WAV_PATHS_REVISION,
+    PendingReadinessDelta, ProcessingLane, ReadinessClassification, ReadinessProgress,
+    ReadinessStore, ReadinessView, RuntimeCandidate, RuntimeTask, SampleSource, SourceAvailability,
+    SourceDiscoveryPhase, SourceDiscoveryStats, SourceHealthSummary, WorkCandidate,
+    active_recording_deferrals, cancelled, clear_duplicate_identity_diagnostic,
+    duplicate_identity_diagnostic_for_revision, earliest_deadline,
+    find_duplicate_identity_diagnostics, legacy_unsupported_decode_failure_text,
     persist_duplicate_identity_diagnostic, publish_current_readiness_delta_with_cancel,
     publish_current_readiness_targets_with_cancel_and_checkpoint, readiness_contract_version,
     retire_legacy_playback_readiness, similarity_prerequisite_blocker_stats,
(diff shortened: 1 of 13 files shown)


PORTALSURFER (commented):
Review result: **changes requested**. The inline required fix identifies a revision-recovery/liveness gap: the duplicate-identity terminal marker can remain active after an in-place repair that preserves the manifest path set. GitHub does not allow this account to submit `REQUEST_CHANGES` on its own PR, so this is recorded as a comment review.

src/native_app/source_processing/supervisor/discovery_reconcile.rs
```diff
@@ -204,6 +206,62 @@ pub(super) fn discover_source_candidates_with_connection_and_progress(
             )),
         )));
     }
+    let manifest_revision = connection
+        .query_row(
+            "SELECT COALESCE(
+                (SELECT CAST(value AS INTEGER) FROM metadata WHERE key = ?1),
+                0
```
PORTALSURFER, line 214: **Required fix — fence recovery on an identity-sensitive manifest revision.** `META_WAV_PATHS_REVISION` is documented in this diff as changing only when the ordered WAV path set changes. A duplicate can be repaired by replacing one alias with a distinct file at the same relative path; that changes `file_identity` but leaves the ordered path set unchanged. The stored diagnostic then still matches `manifest_revision`, so the `!force_manifest_audit` branch above keeps returning the terminal state indefinitely and readiness never resumes without a manual audit.

Use a revision that advances when the supported live identity membership changes (or advance this revision for `file_identity` changes), and add coverage for repairing a duplicate in place at the same path.

PORTALSURFER:
**Non-blocking observation** — [terminal_inputs.rs:100](https://github.com/PORTALSURFER/wavecrate/blob/b71718169ab82acf9a44be7c7b8dd515361fb1a1/src/native_app/source_processing/supervisor/tests/terminal_inputs.rs#L100) manually assigns both rows the same `file_identity` after creating the hard link. That keeps this focused on the reconciliation query, but it does not prove the manifest-refresh path records identical stable identities for an actual hard link. A small integration-level assertion through the normal manifest update would make the regression test cover that boundary as well.

src/native_app/source_processing/supervisor/discovery_reconcile.rs
PORTALSURFER: Fixed in commit c8d2d4c33. Duplicate parking now uses the dedicated `META_WAV_IDENTITIES_REVISION`, which advances when `file_identity` changes and when live rows are removed or renamed, while preserving the ordered-path revision semantics. The regression now replaces the alias with a distinct file at the same path and verifies readiness resumes without changing the path set.

PR #932, opened 2026-07-24
fix(scan): make targeted subtree SQL path-component exact

PORTALSURFER:
## Goal

Implement Linear issue OPT-1280 so targeted source refreshes select only the requested persisted source-relative subtree.

## Scope and non-goals

- Centralize one exact binary path-range predicate for wav-file and index-only subtree reads.
- Preserve direct-target equality, path-component boundaries, literal wildcard characters, case distinctions, and existing path indexes.
- Add database planner and targeted-sync regressions.
- No schema migration, audio-filter change, path-schema redesign, or targeted-manifest performance work.

## Definition of Done

- Percent, underscore, and exclamation-mark path components are treated literally.
- Case-distinct and similarly named sibling directories are not crossed.
- Wav and index-only subtree reads use the same predicate contract.
- Query-plan evidence confirms path-index use.
- Targeted reconciliation does not select or mutate unrelated rows.

## Validation

- cargo test -p wavecrate-library --all-targets --no-fail-fast — 270 passed.
- cargo test -p wavecrate-scan --all-targets --no-fail-fast — 178 passed.
- cargo fmt --all -- --check — passed.
- git diff --check — passed.
- bash scripts/ci.sh agent — Wavecrate checks passed; the gate then failed in the unrelated vendored Radiant test gui_runtime::native_vello::generic_runtime::tests::gpu_surface_runtime::signal_summary::gpu_signal_shader_keeps_waveform_bands_visually_distinct at vendor/radiant/src/gui_runtime/native_vello/generic_runtime/tests/gpu_surface_runtime/signal_summary.rs:69.

## Known limitations

The agent gate remains blocked by the pre-existing vendored Radiant shader-color assertion; no vendor changes are included.

### DIFF
--- a/crates/wavecrate-library/src/sample_sources/db/index_entries.rs
+++ b/crates/wavecrate-library/src/sample_sources/db/index_entries.rs
@@ -3,7 +3,10 @@ use std::path::Path;
 use rusqlite::{Connection, OptionalExtension, params};
 
 use super::schema;
-use super::util::{map_sql_error, normalize_relative_path, parse_relative_path_from_db};
+use super::util::{
+    EXACT_SUBTREE_PATH_PREDICATE, exact_subtree_path_bounds, map_sql_error,
+    normalize_relative_path, parse_relative_path_from_db,
+};
 use super::{
     META_SOURCE_INDEX_REVISION, SourceDatabase, SourceDbError, SourceIndexClassification,
     SourceIndexDiagnostic, SourceIndexEntry, SourceIndexSnapshot, SourceWriteBatch,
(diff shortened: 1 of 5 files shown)


PORTALSURFER (commented):
Review of the current PR description and complete diff: clean. The shared BINARY path equality/range predicate satisfies the stated exact-subtree contract, preserves path-component boundaries and wildcard/case distinctions, is used by both wav and index-only reads, and has focused planner plus targeted-reconciliation coverage. No blocking goal mismatch found.

PR #933, opened 2026-07-24
OPT-1281: isolate non-Unicode source paths during scan

PORTALSURFER:
## Goal

Prevent a single Unix filesystem entry with a non-UTF-8 path component from aborting Wavecrate source scan publication.

## Scope

- Persist non-Unicode source paths losslessly in the source index with an explicit path-encoding marker.
- Preserve a typed NonUnicodePath diagnostic for supported audio entries while allowing sibling entries to publish.
- Apply equivalent behavior to full scans and targeted add, modify, rename, and delete synchronization.
- Add schema migration compatibility and Unix regression coverage.

## Non-goals

- Changing ordinary UTF-8 path normalization or sample database path semantics.
- Converting arbitrary non-Unicode filesystem names into displayable Unicode.
- Broad scanner or schema cleanup unrelated to this failure mode.

## Definition of Done

- Supported and unsupported non-Unicode entries remain represented in the source index without normalization failure.
- Supported non-Unicode entries receive a typed diagnostic and do not become readable samples.
- Sibling entries continue to scan and publish.
- Targeted synchronization handles add, modify, rename, and removal of non-Unicode entries.
- Existing databases migrate and remain readable.

## Validation

- cargo test -p wavecrate-library --all-targets --no-fail-fast: 277 passed.
- cargo test -p wavecrate-scan --all-targets --no-fail-fast: 178 passed.
- Focused migration, reserved-prefix lifecycle, and manifest/index key-collision regressions: passed.
- Focused index-entry regression: 10 passed on macOS; raw-byte filesystem cases are platform-gated because Darwin rejects such filenames.
- cargo fmt --all -- --check: passed.
- git diff --check: passed.

## Known limitations

- bash scripts/ci.sh agent reaches the unrelated Radiant assertion gpu_signal_shader_keeps_waveform_bands_visually_distinct. Tracked in OPT-1288.
- Package-wide Clippy reports pre-existing warnings in untouched files; changed code and focused tests pass.

Fixes OPT-1281.

User approval is required before merge. Approval received on 2026-07-24.

### DIFF
--- a/crates/wavecrate-library/src/sample_sources/db/index_entries.rs
+++ b/crates/wavecrate-library/src/sample_sources/db/index_entries.rs
@@ -5,7 +5,7 @@ use rusqlite::{Connection, OptionalExtension, params};
 use super::schema;
 use super::util::{
     EXACT_SUBTREE_PATH_PREDICATE, exact_subtree_path_bounds, map_sql_error,
-    normalize_relative_path, parse_relative_path_from_db,
+    normalize_source_index_path, parse_source_index_path_from_db,
 };
 use super::{
     META_SOURCE_INDEX_REVISION, SourceDatabase, SourceDbError, SourceIndexClassification,
@@ -39,12 +39,10 @@ impl SourceDatabase {
             .unchecked_transaction()
             .map_err(map_sql_error)?;
         let revision = read_index_revision(&transaction)?;
+        let path_encoding = source_index_path_encoding_available(&transaction)?;
         let entries = collect_entries(
             &transaction,
-            "SELECT path, classification, file_size, modified_ns, file_identity,
-                    diagnostic, format_policy_version
-             FROM source_index_entries
-             ORDER BY path ASC",
+            &source_index_select_sql(path_encoding, false),
             [],
         )?;
         transaction.rollback().map_err(map_sql_error)?;
@@ -64,17 +62,12 @@ impl SourceDatabase {
         if !source_index_schema_available(&self.connection)? {
             return Ok(Vec::new());
         }
-        let normalized = normalize_relative_path(relative_path)?;
+        let (normalized, _path_encoding) = normalize_source_index_path(relative_path)?;
         let (lower_bound, upper_bound) = exact_subtree_path_bounds(&normalized);
+        let path_encoding = source_index_path_encoding_available(&self.connection)?;
         collect_entries(
             &self.connection,
-            &format!(
-                "SELECT path, classification, file_size, modified_ns, file_identity,
-                    diagnostic, format_policy_version
-             FROM source_index_entries
-             WHERE {EXACT_SUBTREE_PATH_PREDICATE}
-             ORDER BY path ASC"
-            ),
+            &source_index_select_sql(path_encoding, true),
             params![normalized, lower_bound, upper_bound],
         )
     }
@@ -87,7 +80,7 @@ impl SourceWriteBatch<'_> {
         entry: &SourceIndexEntry,
     ) -> Result<(), SourceDbError> {
         validate_entry(entry)?;
-        let path = normalize_relative_path(&entry.relative_path)?;
+        let (path, path_encoding) = normalize_source_index_path(&entry.relative_path)?;
         let live_manifest_row = self
             .tx
             .query_row(
@@ -105,10 +98,11 @@ impl SourceWriteBatch<'_> {
             .tx
             .execute(
                 "INSERT INTO source_index_entries (
-                    path, classification, file_size, modified_ns, file_identity,
+                    path, path_encoding, classification, file_size, modified_ns, file_identity,
                     diagnostic, format_policy_version
-                 ) VALUES (?1, ?2, ?3, ?4, ?5, ?6, ?7)
+                 ) VALUES (?1, ?2, ?3, ?4, ?5, ?6, ?7, ?8)
                  ON CONFLICT(path) DO UPDATE SET
+                    path_encoding = excluded.path_encoding,
                     classification = excluded.classification,
                     file_size = excluded.file_size,
                     modified_ns = excluded.modified_ns,
@@ -123,6 +117,7 @@ impl SourceWriteBatch<'_> {
                     OR format_policy_version IS NOT excluded.format_policy_version",
                 params![
                     path,
+                    path_encoding,
                     entry.classification.token(),
                     entry.file_size.map(saturating_i64),
                     entry.modified_ns,
@@ -138,7 +133,7 @@ impl SourceWriteBatch<'_> {
 
     /// Remove one index-only row, including during promotion to the supported manifest.
     pub fn remove_source_index_entry(&mut self, relative_path: &Path) -> Result<(), SourceDbError> {
-        let path = normalize_relative_path(relative_path)?;
+        let (path, _path_encoding) = normalize_source_index_path(relative_path)?;
         let changed = self
             .tx
             .execute("DELETE FROM source_index_entries WHERE path = ?1", [path])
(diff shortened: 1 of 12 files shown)


crates/wavecrate-library/src/sample_sources/db/util.rs
```diff
@@ -43,13 +52,140 @@ pub fn normalize_relative_path(path: &Path) -> Result<String, SourceDbError> {
...
+/// encoding bit so their SQLite key can retain the exact raw bytes without
+/// changing the representation of existing Unicode paths. Plain components
+/// stay readable and therefore retain exact-subtree prefix ordering.
+pub(super) fn normalize_source_index_path(path: &Path) -> Result<(String, i64), SourceDbError> {
+    let cleaned = sanitize_relative_path(path)?;
+    let has_reserved_component = cleaned.components().any(|component| {
+        let Component::Normal(part) = component else {
+            return false;
+        };
+        part.to_str().is_some_and(|value| {
```
PORTALSURFER, line 75: **Required fix — migration compatibility:** Legacy databases already persist ordinary UTF-8 index paths such as `~wavecrate-nu~ff.wav` with the old plain key. After this migration, this branch emits the escaped lossless key for the same logical path. On a subsequent metadata change, reconciliation sees the same decoded `PathBuf` in `existing`, so it does not schedule removal, but `upsert_source_index_entry` inserts the escaped key; on deletion, `remove_source_index_entry` only targets the escaped key and leaves the legacy row behind. Migrate/rekey affected existing rows transactionally (including both reserved prefixes), or use a lossless-key scheme that remains compatible with every legacy plain key. Add a regression that updates and then deletes a legacy reserved-prefix row without creating or retaining a duplicate.

crates/wavecrate-scan/src/sample_sources/scanner/scan/tests/index_entries.rs
```diff
@@ -58,6 +58,68 @@ fn full_scan_persists_typed_index_only_entries_across_restart() {
...
+    use std::ffi::OsString;
+    use std::os::unix::ffi::OsStringExt;
+
+    let directory = tempdir().unwrap();
+    let raw_name = OsString::from_vec(b"raw-\xFF.wav".to_vec());
+    let renamed_name = OsString::from_vec(b"renamed-\xFE.wav".to_vec());
+    let raw_path = PathBuf::from(&raw_name);
+    let renamed_path = PathBuf::from(&renamed_name);
+    std::fs::write(directory.path().join("ordinary.wav"), b"ordinary").unwrap();
+    std::fs::write(directory.path().join(&raw_name), b"raw").unwrap();
```
PORTALSURFER: **Required fix — missing required regression coverage:** This initial full scan creates the non-Unicode entry, and the test later exercises targeted modify, rename, and removal; the final add at line 115 again uses `scan_once`. The Definition of Done explicitly requires targeted synchronization to handle **add**. Create the raw filename after a prior scan and call `sync_paths` for it, then assert the lossless `NonUnicodePath` index row and that the ordinary sibling remains published.

PORTALSURFER (commented):
## Required fixes

1. Existing index-only rows whose ordinary Unicode component begins with either newly reserved encoding prefix are not migrated to the new canonical key. Once such a file changes, the scan inserts a second row; once it is deleted, reconciliation deletes only that new key and leaves the legacy row behind. This violates the migration/readability and ordinary UTF-8 compatibility requirements.

2. The Unix regression covers targeted modify, rename, and removal, but performs the final addition through a full scan. Add targeted-add coverage as required by the Definition of Done.

Reviewed only the PR description and complete merge-base (`51e49e333161be1fdccdfcb1a1bf2c92089c0f78`) to head diff. GitHub does not permit this reviewer identity to submit `REQUEST_CHANGES` on its own PR.

crates/wavecrate-library/src/sample_sources/db/util.rs
PORTALSURFER: Fixed in c8bbbe423. Source-index schema setup now transactionally rekeys legacy plain rows whose components use either reserved prefix, preserves an existing canonical row on mixed-state collisions, and prevents duplicate/stale rows across update and delete. Added migration collision coverage plus an update/delete lifecycle regression. Validation: 276 wavecrate-library tests passed.

crates/wavecrate-scan/src/sample_sources/scanner/scan/tests/index_entries.rs
PORTALSURFER: Fixed in c8bbbe423. The Unix regression now creates the raw non-Unicode filename after the initial scan and adds it through sync_paths, then verifies the lossless NonUnicodePath index row and the ordinary sibling before continuing through modify, rename, and removal. Validation: 178 wavecrate-scan tests passed; the invalid-byte case remains platform-gated on macOS.

crates/wavecrate-scan/src/sample_sources/scanner/scan_paths.rs
```diff
@@ -626,6 +631,18 @@ fn collect_current_file(
             return Ok(());
         }
     };
+    if relative_path.to_str().is_none() {
+        current_index_entries.insert(
+            relative_path.to_path_buf(),
+            non_unicode_index_entry(
```
PORTALSURFER, line 640: **Required fix** — This path bypasses the existing file classification and writes *every* non-Unicode targeted entry as `Inaccessible` with `NonUnicodePath`. A targeted add/modify/rename for a non-Unicode unsupported audio or non-audio file therefore no longer matches the full-scan behavior, and `NonUnicodePath` is applied even though the PR contract reserves that typed diagnostic for supported audio. Preserve `index_entry_from_file_facts` classifications for unsupported entries; create the `NonUnicodePath` entry only for supported audio. Add a targeted-sync regression covering at least one unsupported non-Unicode entry.

PORTALSURFER (commented):
One required fix found. GitHub does not permit this identity to request changes on its own pull request.

crates/wavecrate-scan/src/sample_sources/scanner/scan_paths.rs
PORTALSURFER: Addressed in 0e2f3cdf8. The referenced helper is only called for `indexes_audio()` entries, which by policy implies `SupportedAudio`; unsupported audio and non-audio entries already flow through `collect_current_index_file` and `index_entry_from_file_facts`. I documented that invariant at the branch and added a Unix targeted-sync regression covering raw non-Unicode `.flac` and `.txt` paths, asserting `UnsupportedAudio` / `UnsupportedNonAudio` with no diagnostic while the ordinary sibling remains published. Validation: all 178 wavecrate-scan tests pass on macOS; raw-byte cases are platform-gated because Darwin rejects those filenames.

crates/wavecrate-library/src/sample_sources/db/index_entries.rs
```diff
@@ -87,7 +80,7 @@ impl SourceWriteBatch<'_> {
         entry: &SourceIndexEntry,
     ) -> Result<(), SourceDbError> {
         validate_entry(entry)?;
-        let path = normalize_relative_path(&entry.relative_path)?;
+        let (path, path_encoding) = normalize_source_index_path(&entry.relative_path)?;
```
PORTALSURFER, line 83: **Required fix** — `path_encoding` is not part of the `source_index_entries` primary key, and the following `wav_files` lookup compares the encoded key as if it were an ordinary manifest path. On Unix, a raw `raw-\xFF.wav` is encoded as `~wavecrate-nu~7261772dff2e776176`; a sibling UTF-8 file with that literal name is a valid supported sample and has exactly that `wav_files.path`. This makes `live_manifest_row` true and returns `Unexpected`, aborting scan publication—the failure this PR is intended to isolate. Preserve the distinction at the manifest/index boundary (for example, only apply this manifest collision check to plain-encoded index paths) and add a Unix regression with both files present to prove the normal sample and the typed raw-path index entry publish together.

PORTALSURFER (commented):
## Required changes

The lossless source-index key can collide with an ordinary UTF-8 manifest path, so the change does not yet guarantee that a non-Unicode entry and its siblings publish together.

GitHub does not permit this reviewer identity to request changes on its own pull request.

crates/wavecrate-library/src/sample_sources/db/index_entries.rs
PORTALSURFER: Fixed in 03e5b677c. Manifest collision checks now apply only to plain index paths; lossless-encoded index keys remain a separate identity namespace even when their stored text equals a Unicode manifest path. Added a Unix-capable database regression that constructs a raw parent component and an equal plain manifest storage key, plus a filesystem end-to-end regression for both entries publishing together. Validation: 277 wavecrate-library tests and 178 wavecrate-scan tests pass; formatting and diff checks pass.

PORTALSURFER (commented):
Clean review — reviewed the PR description and complete merge-base-to-head diff at `03e5b677c03bad855053964350ca862829fac4c5`. No blocking goal or Definition-of-Done mismatches found.

PR #935, opened 2026-07-24
fix(scan): bound retained hash descriptors and stale-window replay

PORTALSURFER:
## Goal

Fix OPT-1283 by bounding retained source-file descriptors during content-audit and deferred deep-hash publication, while preserving resumable progress when a manifest revision becomes stale.

## Scope

- Hash and publish in independent 64-descriptor windows for content audits and deferred deep hashing.
- Continue unbounded deferred queues through a path cursor instead of retaining one process-wide descriptor set, including progress past unavailable rows.
- Replay only the current unpublished window after a stale manifest revision.
- Preserve committed deep-hash checkpoint stats when cancellation arrives after an earlier window.
- Preserve the original global time/byte/entry budget semantics across descriptor windows, including the one oversize-file allowance.
- Revalidate rename candidates with short-lived descriptors and preserve rename identity/ambiguity safeguards.
- Propagate the existing targeted-rename overflow error through the top-level source-processing boundary so the current main branch test target remains exhaustive.

Non-goals: changing no-follow validation, writer-hold duration, hashing budgets, or default concurrency.

## Definition of Done

- Retained source descriptors are bounded by a small window independent of time/byte/entry budgets.
- Unbounded deferred hashing drains across checkpointed windows, even when a row in an earlier window is unavailable.
- Stale audit and deferred-hash revisions replay only the unpublished window.
- Cancellation reports committed deep-hash progress and resumes from the durable checkpoint.
- Existing audit byte/time/entry limits remain global to the invocation.
- Regression coverage exercises descriptor bounds, stale replay, unavailable-row cursor progress, cancellation/resume, budget preservation, and rename ambiguity/identity behavior.

## Validation

- `cargo test -p wavecrate-scan --all-targets` — 188 passed at head `b9fc6c0a4`.
- `cargo test -p wavecrate --lib native_app::source_processing::supervisor::tests` — top-level app compiles; 97 passed, 1 existing repeatable failure in `periodic_manifest_audit_wakes_browser_projection_after_committed_repair`.
- `bash scripts/ci.sh agent` — Wavecrate checks/build/guardrails passed; the full Radiant suite has 1 existing unrelated shader-constant failure in `gpu_signal_shader_keeps_waveform_bands_visually_distinct`.
- Independent Terra review found three required fixes across two passes; all are included in the current head with regression coverage.

## Known limitations

The two baseline test failures above remain outside this hashing change and are recorded explicitly.

### DIFF
--- a/crates/wavecrate-scan/src/sample_sources/scanner/scan_hash.rs
+++ b/crates/wavecrate-scan/src/sample_sources/scanner/scan_hash.rs
@@ -58,6 +58,7 @@ struct ContentAuditWindow {
     attempted_entries: usize,
     attempted_bytes: u64,
     completed_rotation: bool,
+    had_content_work: bool,
 }
 
 struct DeepHashWindow {
@@ -379,6 +380,7 @@ fn verify_content_batch_with_source_root_hooks(
     let mut remaining_bytes = budget.max_bytes;
     let mut first_window = true;
     let mut stale_replayed = false;
+    let mut content_work_seen = false;
 
     loop {
         let window_budget = ContentAuditBudget {
@@ -398,6 +400,7 @@ fn verify_content_batch_with_source_root_hooks(
             window_budget,
             now,
             writer,
+            content_work_seen,
             post_hash,
             precommit,
             elapsed,
@@ -407,6 +410,7 @@ fn verify_content_batch_with_source_root_hooks(
                 stale_replayed = false;
                 remaining_entries = remaining_entries.saturating_sub(window.attempted_entries);
                 remaining_bytes = remaining_bytes.saturating_sub(window.attempted_bytes);
+                content_work_seen |= window.had_content_work;
                 combined.merge_deferred_hashes(window.stats);
                 if window.attempted_entries == 0
                     || window.completed_rotation
@@ -448,6 +452,7 @@ fn verify_content_batch_window_with_source_root_hooks(
     budget: ContentAuditBudget,
     now: i64,
     writer: &impl ScanWriter,
+    prior_content_work: bool,
     post_hash: &mut impl FnMut(&std::path::Path),
     precommit: &mut impl FnMut(&std::path::Path),
     elapsed: &mut impl FnMut() -> Duration,
@@ -534,7 +539,7 @@ fn verify_content_batch_window_with_source_root_hooks(
                 (entry, true)
             }
         };
-        if !verified.is_empty() || !skipped.is_empty() {
+        if prior_content_work || !verified.is_empty() || !skipped.is_empty() {
             if elapsed() >= budget.max_elapsed
                 || attempted_bytes.saturating_add(entry.file_size) > budget.max_bytes
             {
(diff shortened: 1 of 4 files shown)


crates/wavecrate-scan/src/sample_sources/scanner/scan_hash.rs
```diff
@@ -1172,6 +1370,9 @@ fn deep_hash_scan_with_root_hooks(
             stats.hashes_computed += 1;
         }
     }
+    if hash_backfills.len() >= descriptor_window && descriptor_window > 0 {
```
PORTALSURFER, line 1399: Required fix: exhaustion is currently inferred from `hash_backfills.len()`. A bounded query can return a full path window containing an unavailable or changed row, leaving fewer retained descriptors and causing the outer loop to stop even though later pending paths exist. Track whether the fetched query/window reached its boundary and advance the cursor after every examined window; add coverage where an early pending row fails but later rows still hash.

crates/wavecrate-scan/src/sample_sources/scanner/scan_hash.rs
```diff
@@ -1036,6 +1148,68 @@ fn deep_hash_scan_with_root_hooks(
...
+            &mut post_hash,
+            &mut precommit,
+        ) {
+            Err(ScanError::StaleRevision { .. })
+                if cursor.is_some() && replayed_cursor.as_deref() != cursor.as_deref() =>
+            {
+                replayed_cursor = cursor.clone();
+                continue;
+            }
+            result => result?,
```
PORTALSURFER, line 1197: Required fix: if a later deep-hash window is canceled after earlier windows committed, `result?` returns `Canceled` and drops the accumulated `combined` checkpoint stats. Preserve committed progress using the existing incomplete/checkpoint result shape, and add a cancellation-after-first-window regression that verifies the reported progress and resume.

PORTALSURFER (commented):
Required fixes identified by the independent review are recorded below. I am addressing both in the next fix batch.

crates/wavecrate-scan/src/sample_sources/scanner/scan_hash.rs
PORTALSURFER: Fixed in `b9fc6c0a4`: exhaustion now follows the bounded pending-query result rather than successful descriptor admissions, and the cursor advances past unavailable rows. Added `unbounded_deferred_hashes_advance_past_an_unavailable_window_entry`; the full scan suite passes 188/188.

crates/wavecrate-scan/src/sample_sources/scanner/scan_hash.rs
PORTALSURFER: Fixed in `62702324e`: cancellation after a committed deep-hash window returns accumulated `Incomplete` checkpoint stats and resumes correctly. Added `canceled_deferred_hashing_reports_committed_window_and_resumes`; the full scan suite passes 188/188.

PR #937, opened 2026-07-25
fix(scan): persist signed pre-epoch filesystem timestamps

PORTALSURFER:
## Goal

Fix scanner timestamp persistence so filesystem modification times before the Unix epoch are represented consistently as signed nanoseconds.

## Scope and non-goals

- Centralize scanner `SystemTime` conversion for manifest and source-index facts.
- Preserve exact in-range negative nanoseconds and saturate outside the `i64` range.
- Cover full scans, targeted scans, index-only entries, and descriptor revalidation through the shared facts path.
- Add boundary, ordering, and pre-epoch filetime regression coverage.
- Document the durable signed-nanosecond contract.
- No SQLite schema/unit changes, filesystem timestamp rewriting, timezone handling, or timestamp-only content identity changes.

## Definition of Done

- Epoch and pre-epoch timestamps persist as signed nanoseconds.
- Exact `i64` bounds and deterministic overflow saturation are covered.
- Full and targeted manifest/index paths converge without rejecting pre-epoch files.
- Existing ordinary timestamp and changed-file behavior remains green.

## Validation

- `cargo test -p wavecrate-scan --all-targets` — 193 passed.
- `cargo test -p wavecrate-library --all-targets` — 282 unit tests passed plus process-boundary coverage passed.
- `bash scripts/ci.sh agent` — baseline failure in unrelated Radiant visual-shader test `gpu_signal_shader_keeps_waveform_bands_visually_distinct`; Wavecrate package tests passed.

## Known limitations

Windows execution was not available locally; the filetime integration test is gated for Unix and Windows, with a deterministic unit-test contract on other platforms.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6261,6 +6261,7 @@ dependencies = [
  "blake3",
  "cap-fs-ext",
  "cap-std",
+ "filetime",
  "libc",
  "tempfile",
  "thiserror 2.0.18",
--- a/crates/wavecrate-scan/Cargo.toml
+++ b/crates/wavecrate-scan/Cargo.toml
@@ -16,6 +16,7 @@ thiserror = "2.0.17"
 tracing = "0.1.41"
 
 [dev-dependencies]
+filetime = "0.2.26"
 tempfile = "3.23.0"
 
 [target.'cfg(unix)'.dev-dependencies]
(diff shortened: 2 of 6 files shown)


PORTALSURFER (commented):
Terra/high complete-diff review: clean. The shared signed-nanosecond conversion, pre-epoch persistence/convergence coverage, i64-bound saturation, full/targeted/index-only/descriptor paths, documentation, and regression scope match OPT-1285. No required fixes or non-blocking observations.

PR #955, opened 2026-07-25
Defer browser refreshes during playmark drags

PORTALSURFER:
## Goal
Prevent playmark-drag stalls while a source folder scan updates browser contents.

## Scope
Defer scan-driven browser projection refreshes during an active waveform drag, then apply the latest refresh when the drag ends. Preserve immediate foreground file refreshes during a drag.

## Non-goals
No changes to scanning, source processing, extraction behavior, or waveform audio behavior.

## Definition of Done
- Scan-backed dirty browser refreshes do not rebuild during an active playmark drag.
- The latest browser projection refreshes after release.
- Direct foreground file refreshes remain visible during a drag.

## Validation
- `cargo test -p wavecrate active_waveform_drag_defers_scan_dirty_refresh_until_release` — passed.

## Known limitations
None.

User approval authorizes merge.

### DIFF
--- a/src/native_app/app_chrome/view_models/sample_browser.rs
+++ b/src/native_app/app_chrome/view_models/sample_browser.rs
@@ -291,10 +291,11 @@ fn empty_starmap_items() -> Arc<[StarmapItem]> {
 
 fn waveform_drag_defers_sample_browser_preparation(state: &NativeAppState) -> bool {
     state.waveform.current.active_drag_kind().is_some()
-        && !state
+        && (!state
             .library
             .folder_browser
             .visible_sample_window_needs_content_refresh()
+            || state.library.folder_browser.scan_content_refresh_pending())
 }
 
 fn starmap_audition_drag_defers_sample_browser_preparation(state: &NativeAppState) -> bool {
@@ -316,7 +317,11 @@ mod tests {
     use crate::native_app::test_support::state::{NativeAppStateFixture, WaveformInteraction};
     use crate::native_app::waveform::WaveformSelectionKind;
     use crate::native_app::{
-        app::SampleBrowserDisplayMode, sample_library::folder_browser::FolderBrowserState,
+        app::SampleBrowserDisplayMode,
+        sample_library::folder_browser::{
+            FolderBrowserState,
+            scan::{FolderScanDiscoveryBatch, FolderScanLifecycle, scan_source_with_progress},
+        },
     };
     use radiant::{
         gui::types::Point,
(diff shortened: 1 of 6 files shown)


src/native_app/app_chrome/view_models/sample_browser.rs
```diff
@@ -291,10 +291,11 @@ fn empty_starmap_items() -> Arc<[StarmapItem]> {
 
 fn waveform_drag_defers_sample_browser_preparation(state: &NativeAppState) -> bool {
     state.waveform.current.active_drag_kind().is_some()
-        && !state
+        && (!state
             .library
             .folder_browser
             .visible_sample_window_needs_content_refresh()
```
PORTALSURFER, line 297: **Required fix** — `folder_scan_active()` makes this predicate defer *every* pending browser refresh during a playmark drag whenever a scan is active. That includes a direct foreground `refresh_file_path` request, even though the PR description explicitly requires foreground file refreshes to remain immediate during a drag. Track or carry whether the pending refresh is scan-driven, and defer only that case; add coverage that a foreground refresh applies immediately while a scan is active, alongside the deferred scan-refresh/release case.

PORTALSURFER (commented):
## Required fixes

- The refresh gate must distinguish scan-driven pending work from a direct foreground file refresh. As written, the active-scan condition defers both, so the diff does not meet the Definition of Done.

Requesting changes was not permitted because this GitHub identity owns the PR. Required fix recorded inline at `b64ae3c8af68f39857a89e931a3267830980d0f7`.

src/native_app/app_chrome/view_models/sample_browser.rs
PORTALSURFER: Addressed: scan-originated browser revisions are now marked separately; generic foreground file refreshes clear that marker. The drag gate defers only dirty scan-originated work while the scan is active. Added coverage for scan deferral/release and immediate foreground refresh during an active scan.

PORTALSURFER (commented):
Required fix: the current diff does not defer every scan-driven dirty projection refresh required by the PR contract.

src/native_app/app_chrome/view_models/sample_browser.rs
```diff
@@ -291,10 +291,12 @@ fn empty_starmap_items() -> Arc<[StarmapItem]> {
...
+            || (state.library.folder_scan_active()
```
PORTALSURFER: Required fix — `scan_content_refresh_pending()` already identifies that the current dirty revision came from a scan, but this predicate also requires `folder_scan_active()`. The scan-finalization paths in this diff mark their resulting dirty revision as scan-backed; once the lifecycle flips inactive before the next view preparation, the extra check makes this return false and rebuilds that scan update during the active playmark drag. That violates the stated requirement to defer scan-driven projection refreshes and reintroduces the stall at scan completion. Defer based on the scan-pending revision alone (while preserving `bump_file_content_revision()` clearing it for foreground refreshes), and add a regression case where the scan is no longer active but its final scan refresh is still dirty.

PORTALSURFER: Addressed: the gate now defers a dirty scan-provenance revision throughout the active waveform drag without checking live scan state. The regression completes the scan before the drag and confirms the marked revision waits until release; foreground refresh coverage remains.

PORTALSURFER (commented):
Clean review — the current diff satisfies the PR description and Definition of Done. It defers scan-originated dirty browser projection revisions during an active waveform drag, refreshes the latest projection after release, and preserves immediate foreground file refreshes (including while a scan is active). Focused regression coverage exercises both required paths.

src/native_app/sample_library/folder_browser/state.rs
```diff
@@ -541,6 +544,16 @@ impl FolderBrowserState {
 
     pub(super) fn bump_file_content_revision(&mut self) {
         self.sample_list.bump_content_revision();
+        self.scan_content_refresh_revision = None;
```
PORTALSURFER: **Required fix — foreground refresh clears an already-deferred scan refresh**

`bump_file_content_revision()` unconditionally clears `scan_content_refresh_revision`, while the drag defer predicate recognizes a scan update only when that marker equals the current content revision. If scan content becomes dirty, a playmark drag starts, and a foreground file refresh then advances the content revision, the marker is lost. The next prepare during the drag rebuilds the dirty projection, exposing the scan-driven update and reintroducing the stall this PR must prevent.

Preserve the pending scan boundary across a later foreground refresh (or represent scan and foreground invalidations separately), so the foreground update is visible immediately while the already-pending scan refresh remains deferred until release. Add a regression for: scan change -> begin playmark drag -> foreground file refresh -> prepare during drag; assert the foreground update is visible and the scan refresh is still deferred. The existing foreground test only starts a scan; it does not first create a pending scan projection, so it misses this ordering.

PORTALSURFER (commented):
Required fix recorded inline. The prior clean summary is superseded: the current diff does not preserve a deferred scan refresh when a later foreground refresh occurs during the same playmark drag.

src/native_app/sample_library/folder_browser/state.rs
PORTALSURFER: Addressed: a pending scan marker is now preserved and retied across a later foreground content revision, and clears only when visible-window preparation completes. The new mixed-order regression proves scan dirty → drag → foreground refresh keeps the foreground row visible while deferring scan work until release.

src/native_app/app_chrome/view_models/sample_browser.rs
```diff
@@ -291,10 +291,11 @@ fn empty_starmap_items() -> Arc<[StarmapItem]> {
 
 fn waveform_drag_defers_sample_browser_preparation(state: &NativeAppState) -> bool {
```
PORTALSURFER, line 292: Required fix — This condition defers `prepare_sample_browser_view` for every dirty browser update whenever the scan marker is set. A subsequent foreground refresh shares the same `content_revision` and `bump_file_content_revision` deliberately preserves the marker (`state.rs:545-551`), so it still takes this early-return path. That violates the DoD guarantee that direct foreground file refreshes remain visibly immediate during a drag: a foreground update requiring visible-window preparation is deferred with the scan.

Split the pending scan projection from the foreground-refresh projection, or otherwise permit preparation of the foreground delta while retaining the scan-backed projection for release. Extend the mixed-order regression so the completed scan introduces a distinct visible change and the foreground update requires visible-window reconciliation; assert the foreground result is visible while the scan result remains absent until release, then both appear after release.

PORTALSURFER (commented):
Changes requested: the current mixed scan/foreground ordering still does not meet the Definition of Done. (GitHub does not permit this identity to submit a blocking review on its own pull request.)

src/native_app/app_chrome/view_models/sample_browser.rs
PORTALSURFER: Addressed: scan-added file IDs are now deferred separately from the foreground projection. During a drag, foreground reconciliation can render its row while scan-added rows remain excluded until release. The mixed-order regression proves scan row absent plus foreground row visible during drag, then both visible after release.

PR #967, opened 2026-07-27
Align Wavecrate with Radiant 095ddf59

PORTALSURFER:
## Goal

Align Wavecrate with Radiant `095ddf597fbac5a1055550c64b30d27ce62f7c38` and restore a reproducible locked release build.

## Scope and non-goals

- Pin the canonical sibling Radiant contract and release fixtures to `095ddf59`.
- Migrate the GUI worker subscription to `Subscription::worker_payload` with identity mapping.
- Migrate Wavecrate test command execution from removed Radiant inline helpers to the current runtime task-host API.
- Preserve existing GUI-worker payload behavior.
- No Radiant source changes, GUI message redesign, external-drag work, or unrelated cleanup.

## Definition of Done

- Canonical dependency metadata and fixtures contain the exact Radiant SHA.
- The isolated sibling pair reports a clean revision match.
- Wavecrate compiles across the locked workspace and all targets.
- Workspace test targets and standalone Radiant test targets build.
- Release validation and an optimized release build succeed.
- Current complete diff receives an independent review.

## Validation

Validated in an isolated paired checkout with Radiant detached cleanly at the pinned SHA:

- `bash scripts/radiant.sh locate` — pass (`RADIANT_REVISION_MATCH=yes`)
- `git diff --check` — pass
- `rustfmt --edition 2024 --check src/native_app/tests.rs` — pass
- `cargo check --workspace --all-targets --locked` — pass
- `cargo test --locked --test radiant_dependency_contract` — pass (1/1)
- `cargo test --locked --test release_orchestration` — pass (17/17)
- volume focused tests — pass (4/4)
- similarity focused tests — pass (12/12)
- `cargo test --workspace --locked --no-run` — pass
- `bash scripts/internal/release/run_release_validation.sh` — pass, including standalone Radiant targets, release contracts, and scanner tests (193/193)
- `cargo build --release --locked` — pass

## Known limitations

- `cargo fmt --all -- --check` still reports two pre-existing formatting differences outside this diff: `src/native_app/shell/message_dispatch/navigation.rs` and `tests/release_contract.rs`.
- `bash scripts/ci.sh agent` stops at its branch guard in the isolated detached worktree; its constituent locked compile and release gates above were run directly.
- Source-refresh focused tests are 5/7 on this head versus 4/7 on the untouched old-pin baseline. The same two failures reproduce on the baseline.
- File-operation focused tests are 25/52 on this head versus 16/52 on the untouched old-pin baseline. The remaining failures are recorded as existing harness/fixture behavior and are outside this Radiant alignment scope.

### DIFF
--- a/src/native_app/tests.rs
+++ b/src/native_app/tests.rs
@@ -74,6 +74,7 @@ type NativeRuntimeForTests = SurfaceRuntime<
 
 struct CaptureBridge {
     messages: Vec<super::test_support::state::GuiMessage>,
+    target_name: Option<&'static str>,
 }
 
 impl RuntimeBridge<super::test_support::state::GuiMessage> for CaptureBridge {
@@ -105,11 +106,14 @@ impl RuntimeBridge<super::test_support::state::GuiMessage> for CaptureBridge {
 impl RuntimeTaskHost<super::test_support::state::GuiMessage> for CaptureBridge {
     fn spawn_worker_task(
         &mut self,
-        _name: &'static str,
+        name: &'static str,
         _priority: radiant::prelude::TaskPriority,
         _is_cancelled: Option<Box<dyn Fn() -> bool + Send + Sync + 'static>>,
         work: Box<dyn FnOnce() + Send + 'static>,
     ) -> bool {
+        if self.target_name.is_some_and(|target| target != name) {
+            return false;
+        }
         work();
         true
     }
@@ -336,7 +340,12 @@ fn run_worker_message_for_tests(
     if command.business_task_priority(target_name).is_none() {
         return None;
     }
-    run_command_and_capture_messages_for_tests(command).pop()
+    let messages = run_command_and_capture_messages_for_target_tests(command, target_name);
+    match messages.len() {
+        0 => None,
+        1 => messages.into_iter().next(),
+        count => panic!("worker task {target_name} emitted {count} messages; expected exactly one"),
+    }
 }
 
 fn run_command_and_capture_messages_for_tests(
@@ -345,6 +354,7 @@ fn run_command_and_capture_messages_for_tests(
     let mut runtime = radiant::runtime::SurfaceRuntime::new(
         CaptureBridge {
             messages: Vec::new(),
+            target_name: None,
         },
         Vector2::new(900.0, 620.0),
     );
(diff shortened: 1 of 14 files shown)


src/native_app/tests.rs
```diff
@@ -243,46 +286,100 @@ fn run_command_for_tests(
...
+}
+
+fn run_worker_message_for_tests(
+    command: Command<super::test_support::state::GuiMessage>,
+    target_name: &'static str,
+) -> Option<super::test_support::state::GuiMessage> {
+    if command.business_task_priority(target_name).is_none() {
+        return None;
+    }
+    run_command_and_capture_messages_for_tests(command).pop()
```
PORTALSURFER: **Required fix:** Return the completion belonging to `target_name`, not the last completion from the whole command. This helper only proves that a task with that name exists, then executes every task and uses `.pop()` on all captured messages. A batch containing a later message-producing task will therefore make the migrated caller apply/assert against the wrong completion; the replaced helpers recursed to the matching named `Perform`. Isolate and execute the matching task, or retain task-to-message association while capturing, and add a batched-command regression that proves the named completion is returned.

src/native_app/tests/waveform_playback/active_folder_cache.rs
```diff
@@ -1900,18 +1900,7 @@ fn normal_sample_load_persists_bright_cache_indicator_before_restart() {
...
-            Some("latest")
-        }
-        Command::Batch(commands) => commands
-            .iter()
-            .find_map(active_folder_cache_warm_stream_kind),
-        _ => None,
-    }
+    command
+        .business_task_priority("gui-active-folder-cache-warm")
+        .map(|_| "ordered")
```
PORTALSURFER: **Required fix:** Preserve the stream-mode assertion instead of labelling every matching task `ordered`. The old helper returned `ordered` only for `PerformStream` and `latest` for `PerformStreamLatest`; this replacement returns `ordered` for any task with this name, including a coalescing/latest or one-shot implementation. Adapt the test to observe ordered delivery through the current task-host API (or expose equivalent mode metadata), so the Radiant migration still detects a change in cache-warm delivery semantics.

PORTALSURFER (commented):
Independent review of the PR description and complete merge-base-to-head diff. Required fixes are inline. GitHub does not permit this identity to request changes on its own PR.

src/native_app/tests.rs
PORTALSURFER: Fixed in 657f90fbf. The test runtime host now rejects non-target tasks before their closures execute, the one-shot helper captures only target_name and fails on multiple emissions, and a batched target/decoy regression proves the requested completion is returned while the decoy never runs. Focused regression, affected helper lanes, locked all-target check, workspace no-run build, release validation, and release build pass.

src/native_app/tests/waveform_playback/active_folder_cache.rs
PORTALSURFER: Fixed in 657f90fbf. The cache-warm test now executes the command through the runtime host and asserts at least four concrete progress events, the first CheckingCache/LoadingCache pair, nondecreasing processed order, and final completion last. This behaviorally distinguishes ordered delivery from coalesced/latest or one-shot delivery. Focused regression plus the full locked/release validation gates pass.

PORTALSURFER:
**Independent review — clean**

Reviewed head: `657f90fbf9dd4ac460a536d3361de9c6f2a93454` (merge base `f7e7cdc4aaf97912cdcad44d2af8289cf7b82c42`).

Against the current PR description and complete merge-base-to-head diff, the change achieves the stated goal and diff-verifiable Definition of Done: it pins the requested Radiant SHA in canonical dependency metadata and fixtures, preserves GUI-worker payload behavior through identity mapping with `Subscription::worker_payload`, and updates the test command harness to the current runtime task-host API with ordered-stream coverage.

No required fixes. No non-blocking observations.

PR #981, opened 2026-07-30
Bound Playmark extraction reconciliation to committed projections

PORTALSURFER:
## Goal

Make normal Playmark extraction publish its known artifact through one bounded committed-mutation path: authoritative path/identity, browser row/projection, metadata, and readiness, without source-wide manifest/browser/tag materialization or an unnecessary second watcher-triggered full reconciliation.

Root cause addressed: extraction metadata writers used generic upsert behavior that dirtied source path/identity revisions. The targeted scanner then observed a revision gap, while the nominal path also materialized full source snapshots. Large outputs could not be synchronously hashed/acknowledged, allowing a later duplicate watcher sync to widen again.

## Scope and non-goals

- Add typed existing-row metadata-only updates with explicit missing-row deferral; keep rating, lock, playback-tag, collection, and playback-history persistence revision-neutral.
- Build exact affected-path browser projection deltas at the committed source revision, including authoritative empty/duplicate watcher-sync revisions.
- Sequence normal Playmark metadata publication after the created source row is authoritative, preserve optimistic overlays, and avoid forced whole-source tag hydration on the healthy path.
- Fence projection application, watcher acknowledgement, readiness wake, lifecycle generations, operation IDs, identity/content generations, protected-source routing, and conservative recovery.
- Replace committed-mutation full manifest reads with bounded affected-path snapshots.
- Non-goals: Finder folder refresh, extraction DSP, Radiant changes, audio-thread work, UI redesign, or broad profiling/rearchitecture.

## Definition of Done

- Healthy normal extraction uses exact affected-path manifest and browser queries only.
- The exact browser projection revision is applied before watcher acknowledgement and readiness publication.
- Empty/duplicate watcher syncs are no-ops only when the browser already covers the exact authoritative revision.
- Metadata-only writers do not advance path/identity revisions and report missing rows explicitly.
- True revision gaps, incomplete hydration, lifecycle/root changes, unavailable roots, duplicate identities, watcher uncertainty/overflow, and schema/integrity failures retain conservative full recovery.
- Focused regression coverage covers metadata revisions/missing rows, exact browser projection, large/small sync behavior, empty watcher syncs, extraction ordering, lifecycle fencing, and bounded materialization.

## Validation

- `cargo check --locked -p wavecrate --all-targets` — passed.
- `cargo test --locked -p wavecrate-library --all-targets` — passed, 296 tests.
- `cargo test --locked -p wavecrate committed_file_mutations --lib -- --test-threads=1` — passed, 16 tests.
- `cargo test --locked -p wavecrate filesystem_sync --lib -- --test-threads=1` — passed, 7 tests.
- Focused Playmark extraction tests — passed individually, including normal/protected routing, cache eviction, metadata/rating visibility, and harvest persistence.
- `rustfmt --edition 2024 --check` on all changed Rust files — passed.

The combined `cargo test ... playmark_extraction --lib -- --test-threads=1` reached the affected assertions but did not exit after the final fixture test; the affected tests pass individually. This is a known fixture/background-thread shutdown caveat.

Repository-wide formatting has unrelated pre-existing drift outside this change; changed-file formatting is clean.

## Known limitations

- The delegated worktree required a temporary untracked sibling `radiant` symlink for local Cargo validation; it was removed and no Radiant files are part of this PR.
- Live-app validation with real user sources and DAW/audible acceptance remains outside this automated validation.

## Review and merge

This PR is ready for independent Terra review. Explicit user approval is required before merge. Do not merge without that approval.

### DIFF
--- a/crates/wavecrate-library/src/sample_sources/db/write/batch.rs
+++ b/crates/wavecrate-library/src/sample_sources/db/write/batch.rs
@@ -6,7 +6,9 @@ use std::{
 use rusqlite::{OptionalExtension, params};
 
 use super::super::util::map_sql_error;
-use super::super::{Rating, SampleSoundType, SourceDbError, SourceWriteBatch};
+use super::super::{
+    ExistingFileMetadataUpdate, Rating, SampleSoundType, SourceDbError, SourceWriteBatch,
+};
 use super::command::{SourceContentHashWrite, SourceFileWrite, SourceTagWrite};
 use super::mutation::{
     update_flag_statement, update_path_i64_statement, update_path_null_statement,
@@ -31,6 +33,31 @@ const CLEAR_LAST_CURATED_AT_SQL: &str =
     "UPDATE wav_files SET last_curated_at = NULL WHERE path = ?1";
 
 impl<'conn> SourceWriteBatch<'conn> {
+    /// Update filesystem facts for an existing live row without inserting or dirtying path or
+    /// identity revisions.
+    pub fn update_existing_file_metadata(
+        &mut self,
+        relative_path: &Path,
+        file_size: u64,
+        modified_ns: i64,
+    ) -> Result<ExistingFileMetadataUpdate, SourceDbError> {
+        let updated = self.tx.execute(
+            "UPDATE wav_files
+                 SET file_size = ?1, modified_ns = ?2, missing = 0
+                 WHERE path = ?3 AND missing = 0",
+            rusqlite::params![
+                file_size as i64,
+                modified_ns,
+                relative_path.to_string_lossy()
+            ],
+        )?;
+        Ok(if updated == 0 {
+            ExistingFileMetadataUpdate::Missing
+        } else {
+            ExistingFileMetadataUpdate::Updated
+        })
+    }
+
     pub(super) fn apply_file_write(
         &mut self,
         write: SourceFileWrite<'_>,
--- a/crates/wavecrate-library/src/sample_sources/db/write/batch.rs
+++ b/crates/wavecrate-library/src/sample_sources/db/write/batch.rs
@@ -33,6 +33,32 @@ const CLEAR_LAST_CURATED_AT_SQL: &str =
     "UPDATE wav_files SET last_curated_at = NULL WHERE path = ?1";
 
 impl<'conn> SourceWriteBatch<'conn> {
+    /// Check that an existing live row is present without changing scanner-owned facts.
+    ///
+    /// Auxiliary metadata writers use this guard in their own transaction so a stale file
+    /// observation cannot overwrite the manifest's size, mtime, or missing state while an
+    /// external scanner update is racing with the write.
+    pub fn ensure_existing_live_file(
+        &self,
+        relative_path: &Path,
+    ) -> Result<ExistingFileMetadataUpdate, SourceDbError> {
+        let exists = self
+            .tx
+            .query_row(
+                "SELECT 1 FROM wav_files WHERE path = ?1 AND missing = 0 LIMIT 1",
+                [relative_path.to_string_lossy().as_ref()],
+                |_| Ok(()),
+            )
+            .optional()
+            .map_err(map_sql_error)?
+            .is_some();
+        Ok(if exists {
+            ExistingFileMetadataUpdate::Updated
+        } else {
+            ExistingFileMetadataUpdate::Missing
+        })
+    }
+
     /// Update filesystem facts for an existing live row without inserting or dirtying path or
     /// identity revisions.
     pub fn update_existing_file_metadata(
--- a/src/native_app/sample_library/folder_scan_actions/filesystem_refresh.rs
+++ b/src/native_app/sample_library/folder_scan_actions/filesystem_refresh.rs
@@ -174,7 +174,18 @@ impl NativeAppState {
                 } else {
                     false
                 };
-                if incomplete_error.is_none() && !browser_delta_applied {
+                let projection_accepted = if incomplete_error.is_none() && browser_delta_applied {
+                    success
+                        .projection_handoff_ticket
+                        .as_ref()
+                        .is_some_and(|ticket| ticket.accept())
+                } else {
+                    if let Some(ticket) = success.projection_handoff_ticket.as_ref() {
+                        ticket.reject("projection_handoff_projection_rejected");
+                    }
+                    false
+                };
+                if incomplete_error.is_none() && !projection_accepted {
                     incomplete_error = Some(String::from(
                         "committed filesystem sync did not apply an exact browser projection",
                     ));
@@ -195,15 +206,7 @@ impl NativeAppState {
                     renames_reconciled,
                     "Committed filesystem source delta"
                 );
-                if !result.cancelled && incomplete_error.is_none() {
-                    self.background.source_processing.request_source_delta(
-                        &source_id,
-                        result.lifecycle_generation,
-                        &delta,
-                        "filesystem_sync_committed_delta",
-                    );
-                }
-                if !result.cancelled && !delta.is_empty() && incomplete_error.is_none() {
+                if !result.cancelled && !delta.is_empty() && projection_accepted {
                     self.ui.status.sample = format!("Synced {changed_count} filesystem change(s)");
                     self.queue_source_prep(
                         source_id.clone(),
@@ -221,7 +224,7 @@ impl NativeAppState {
                         );
                 }
                 if !result.cancelled
-                    && incomplete_error.is_none()
+                    && projection_accepted
                     && let (Some(watcher), Some(event_id)) = (
                         self.library.source_watcher.as_ref(),
                         journal_checkpoint_event_id,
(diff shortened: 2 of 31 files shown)


crates/wavecrate-library/src/sample_sources/db/write/batch.rs
```diff
@@ -31,6 +33,31 @@ const CLEAR_LAST_CURATED_AT_SQL: &str =
...
+    /// identity revisions.
+    pub fn update_existing_file_metadata(
+        &mut self,
+        relative_path: &Path,
+        file_size: u64,
+        modified_ns: i64,
+    ) -> Result<ExistingFileMetadataUpdate, SourceDbError> {
+        let updated = self.tx.execute(
+            "UPDATE wav_files
+                 SET file_size = ?1, modified_ns = ?2, missing = 0
```
PORTALSURFER, line 72: Required fix — this helper overwrites the manifest’s `file_size` and `modified_ns` while `commit_auxiliary_state()` deliberately leaves the manifest revision unchanged. If a file is changed externally and a queued rating/metadata/playback write runs before its watcher sync, this makes the stored row match the new on-disk facts without clearing/re-hashing content. The scanner’s no-op path then sees matching facts and reuses the prior `content_hash` (`scan_walk.rs` 458–489), so no committed delta/readiness work is published for the real content change. Preserve the existing-row/missing-row check without mutating scanner-owned file facts, or route a detected fact mismatch through the normal manifest mutation/reconciliation path. Add a regression that changes a tracked file, performs this metadata write before the targeted scan, and verifies the scan still produces a content delta.

src/native_app/sample_library/selected_file_actions.rs
```diff
@@ -504,6 +504,49 @@ impl NativeAppState {
     ) {
         match completion.result {
             Ok(path) => {
+                if drag_position.is_none() {
```
PORTALSURFER, line 508: Required fix — the new post-commit ordering applies only when `drag_position` is absent; the drag branch below still calls `assign_extracted_file_metadata()` before `queue_partially_committed_file_mutation()`. Those newly revision-neutral rating and tag workers now return “deferred until source row exists” if they run first, and this branch neither retries them after reconciliation nor carries a post-commit hook. A user who drags an extracted Playmark output can therefore get the optimistic tags/rating for the session, but the durable rating/playback tag is lost after restart (and the status reports incomplete metadata) whenever the metadata worker wins the race. Preserve the immediate drag UI handoff, but defer/retry the durable metadata assignment until the created row is authoritative, and add a drag-extraction regression that forces the metadata worker to run before source reconciliation.

src/native_app/sample_library/folder_scan_actions/filesystem_refresh_worker.rs
```diff
@@ -197,38 +194,44 @@ fn build_browser_projection_delta(
...
         .chain(
             delta
                 .moved
                 .iter()
-                .map(|entry| entry.new_relative_path.as_path()),
+                .map(|entry| entry.new_relative_path.clone()),
         )
+        .collect::<Vec<_>>();
+    let snapshot = db
+        .browser_metadata_for_paths_at_revision(delta.revision, &projection_paths)
```
PORTALSURFER, line 231: Required fix — `browser_metadata_for_paths_at_revision` can fail on a revision race, which makes this worker return `browser_projection_delta: None` without setting `incomplete_error`. The successful result is then handed to `release_after_handoff(CommittedDelta)` before GUI delivery, and `finish_source_filesystem_sync` advances the watcher journal checkpoint and publishes the delta even when `apply_committed_projection_delta` is false. Thus a concurrent manifest revision/projection gap acknowledges the external event and starts readiness before the browser covers its authoritative revision, contrary to the Definition of Done. Treat an unavailable or unapplied exact projection as a full-recovery handoff, and withhold the watcher checkpoint and readiness publication until the projection is applied; add a regression test for that race/failure path.

src/native_app/sample_library/folder_scan_actions/filesystem_refresh_worker.rs
```diff
@@ -165,13 +165,25 @@ fn sync_source_database_paths_once(
                     }
                 }
             };
-            let browser_projection_delta = if browser_delta_eligible
-                && incomplete_error.is_none()
-                && completed.committed_delta.revision > 0
-            {
+            let browser_projection_delta = if browser_delta_eligible && incomplete_error.is_none() {
```
PORTALSURFER, line 168: Required fix: the false branch leaves `browser_projection_delta = None` with no `incomplete_error`. `finish_source_filesystem_sync` then treats that as healthy: it publishes `committed_delta`, queues readiness, and advances the watcher journal checkpoint before its later projection-gap recovery. A mixed/non-audio watcher batch therefore violates the stated rule that an unavailable/rejected exact projection must prevent delta/readiness/checkpoint publication. Mark this case incomplete (and likewise ensure projection-application rejection is fenced) so it goes directly to conservative full recovery.

PORTALSURFER (commented):
Required fix (review identity cannot request changes on its own pull request): unavailable/rejected projection must fence committed-delta publication, readiness, and watcher checkpointing.

src/native_app/sample_library/folder_scan_actions/filesystem_refresh.rs
```diff
@@ -145,16 +161,29 @@ impl NativeAppState {
...
-                            .apply_committed_projection_delta(&source_id, projection)
-                    })
-                    .unwrap_or(false);
+                            .apply_committed_projection_delta(&source_id, projection),
+                        None => false,
+                    }
+                } else {
+                    false
+                };
+                if incomplete_error.is_none() && !browser_delta_applied {
```
PORTALSURFER: Required fix: this UI-side rejection is too late to enforce the stated completion boundary. `queue_source_filesystem_sync` releases `ExternalScanHandoff::CommittedDelta` before `SourceFilesystemSyncFinished` is delivered; `register_handoff` immediately queues it into `pending_readiness_deltas` and wakes the supervisor (its admission test explicitly verifies this is owned before GUI delivery). Thus a later `apply_committed_projection_delta` rejection can leave readiness consuming/publishing the committed delta before this branch switches to full recovery. Keep the task completion/permit fenced until the UI accepts the exact projection; only then publish the delta and advance the journal. On no projection or rejection, hand off only full recovery. Add an end-to-end ordering regression that includes the worker handoff followed by a rejected projection.

PORTALSURFER (commented):
Required fix (review identity cannot submit REQUEST_CHANGES on its own PR): completion must not publish a targeted source delta before the browser projection acceptance boundary. See inline comment.

src/native_app/source_processing/supervisor/admission.rs
```diff
@@ -15,6 +15,163 @@ pub(in crate::native_app) enum ExternalScanHandoff {
...
+        control.pending_projection_fences.remove(&self.source_id);
+        let accepted = self.delta.is_empty()
+            || matches!(
+                control.queue_source_delta(
+                    &self.source_id,
+                    self.lifecycle_generation,
+                    &self.delta,
+                    "projection_handoff_accepted",
+                ),
+                SourceDeltaQueueResult::Queued | SourceDeltaQueueResult::Ignored
```
PORTALSURFER: Required fix: do not accept SourceDeltaQueueResult::Ignored here. It is returned for a stale or duplicate manifest revision (and a stale pending-delta merge), so a replayed or duplicate ticket clears the projection fence and proceeds to watcher acknowledgement and readiness without the required full recovery. Treat only Queued as success (besides the explicit healthy empty-delta checkpoint), and send Ignored through the same cancel and dirty full-reconciliation path as Fallback or Rejected; add a duplicate or stale-ticket regression test.

PORTALSURFER (commented):
Required fix: duplicate or stale projection-handoff acceptance is currently treated as a healthy targeted handoff. The DoD requires conservative full recovery for duplicate/stale resolution; see the inline comment. I could not submit REQUEST_CHANGES because this GitHub identity owns the PR.

src/native_app/source_processing/supervisor/admission.rs
```diff
@@ -15,6 +15,163 @@ pub(in crate::native_app) enum ExternalScanHandoff {
...
+        if *state != ProjectionTicketState::Pending {
+            return false;
+        }
+        *state = resolution;
+        true
+    }
+
+    fn request_full_reconciliation(&self, reason: &'static str) {
+        let mut control = self.shared.control();
+        control.pending_projection_fences.remove(&self.source_id);
```
PORTALSURFER: Required fix: this removes the fence by source ID without checking the ticket generation and revision. A stale delayed ticket (or a second resolution of a clone) can arrive after a newer ticket has fenced the same source; it erases that newer fence, allowing same-source discovery/readiness to run before the newer GUI projection is accepted. Remove only a matching fence, and only cancel or mark dirty for the matching lifecycle. Add a replacement or duplicate-resolution regression that proves a later fence remains installed.

PORTALSURFER (commented):
Required fix: stale or duplicate tickets can remove a newer source fence. See the inline comment.

PR #1065, opened 2026-09-22
OPT-1802: document and stress bounded log retention

PORTALSURFER:
## Work in this draft
- Document the ten-file and 10 MiB prospective segment limits, intact oversized events, legacy-file behavior, profile location, support-bundle selection, and degraded-retention diagnostics in module and user/support docs.
- Add a no-sleep temporary-profile stress test that sends 18 events through the nonblocking worker, causes 17 rotations, and checks ten retained ordered names, every retained segment's contents, size/overshoot, and newest-log selection.
- Clarify helper comments and support-bundle wording. Resolve the `bug-bundle` output directory before archiving so its documented relative default works.

## Verification completed
- Focused stress test at `e32dd87c3` and `cargo test -p wavecrate --locked --lib logging::files::tests` passed; the stress test now verifies every retained segment's contents.
- `cargo check -p wavecrate --locked --all-targets`, `bash scripts/ci.sh smoke`, Bash syntax, targeted rustfmt, and `git diff --check` passed at the implementation head. The later support-helper change passed `bash -n` and `git diff --check`.
- Disposable-profile `bug-bundle` runs with both relative and absolute output directories produced ZIP archives containing `meta/info.txt` at `b2be21e7f`.
- A debug native launch under a disposable automated-test profile initialized logging, created the expected segment, opened the Radiant runtime and audio output, and showed no startup or logging-worker panic before it was stopped. The prepared window was not revealed because application activation timed out, so representative navigation/playback was not exercised.

## Remaining acceptance gates
- `cargo test -p wavecrate --locked --lib` showed widespread unrelated failures and a long-running test; the run was stopped. One failing slice-export test was rerun alone on this branch and on a clean `origin/main` checkout with the same pinned Radiant dependency. Both failed identically with `Source manifest changed while the scan checkpoint was being finalized (expected revision 2, found 4)`. This baseline failure is outside this logging change, but the broad suite is not green.
- Obtain visible native debug/verbose navigation/playback soak evidence before treating OPT-1802 and parent OPT-1444 as accepted. Windows PowerShell helpers remain unexecuted on this macOS host.


PORTALSURFER:
Support-helper review found that `bug-bundle` could not write its archive when `--out-dir` was relative (including the documented default): the archive command changes to a temporary directory before using that relative path. Commit b2be21e7f resolves the output directory before archiving. I verified disposable-profile bundles with both relative and absolute output directories, checked each ZIP contains `meta/info.txt`, and passed `bash -n` and `git diff --check`. Native debug/verbose soak and Windows PowerShell helper evidence remain open.

